PR #7, opened 2017-04-24
Hotfix/name propagation

JonathanCSmith:
Propagates the dataset naming config value into the get_cube_file_name, get_cube_full_path, read_cube, write_cube and make_mag1_cubes_from_z_stack functions.

Fix required to allow automated processing / deployment of datasets and prevents name collision.

### DIFF
--- a/wkcuber/cube_io.py
+++ b/wkcuber/cube_io.py
@@ -11,20 +11,20 @@ def get_cube_folder(target_path, mag, x, y, z):
                      'z{:04d}'.format(z))
 
 
-def get_cube_file_name(mag, x, y, z):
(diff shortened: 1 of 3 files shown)


hotzenklotz:
@JonathanCSmith Thanks for your PR.

@normanrz Was there any specific reason why we didn't include config's dataset name in the resulting cubes?

normanrz:
@JonathanCSmith Thank you very much for your contribution. I have been on vacation, so I couldn't respond earlier.

@hotzenklotz I suppose webKnossos is indifferent about the filename prefix, so I didn't implement this.

PR #11, opened 2018-04-19
extract functions

valentin-pinkau:
Extract compress, downsample_mags, and write_webknossos_metadata into functions.
Enable other projects to use these functions directly.


wkcuber/metadata.py
```diff
@@ -122,11 +134,16 @@ def create_parser():
         help="Scale of the dataset (e.g. 11.2,11.2,25)",
         default="1,1,1")
 
+    parser.add_argument(
```
normanrz: I would invert the logic here: Explicitly flag to search for max id, because it is such a resource-intense activity

PR #16, opened 2018-06-05
actually enforce data type

valentin-pinkau:
This PR makes sure that written data has the same type as defined in header.

### DIFF
--- a/wkcuber/cubing.py
+++ b/wkcuber/cubing.py
@@ -64,9 +64,9 @@ def create_parser():
 
     return parser
 
-def read_image_file(file_name):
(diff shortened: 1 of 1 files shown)


wkcuber/cubing.py
```diff
@@ -64,9 +64,9 @@ def create_parser():
 
     return parser
 
-def read_image_file(file_name):
+def read_image_file(file_name, dtype):
```
georgwiese: Where is this function used? Did you check that it's never called without the `dtype` parameter?

valentin-pinkau: This function is called from `cubing_job` only

PR #22, opened 2018-07-12
Add max_id parameter to metadata

valentin-pinkau:
This PR adds the `max_id` flag.


wkcuber/metadata.py
```diff
@@ -139,18 +143,22 @@ def create_parser():
         default="1,1,1",
     )
 
-    parser.add_argument(
+    group = parser.add_mutually_exclusive_group()
```
normanrz: nice!

PR #24, opened 2018-08-23
add read_metadata_for_layer function

valentin-pinkau:
This PR adds a function to parse the datasource-properties.json file.
Given a layer name the function returns dtype, bounding_box , origin, and the corresponding layer field of the properties file.

### DIFF
--- a/wkcuber/__init__.py
+++ b/wkcuber/__init__.py
@@ -1,4 +1,4 @@
 from .cubing import cubing
 from .downsampling import downsample_mag
 from .compress import compress_mag
-from .metadata import write_webknossos_metadata, read_metadata_for_layer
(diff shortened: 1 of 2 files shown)


wkcuber/__init__.py
```diff
@@ -1,4 +1,4 @@
 from .cubing import cubing
 from .downsampling import downsample_mag
 from .compress import compress_mag
-from .metadata import write_webknossos_metadata
+from .metadata import write_webknossos_metadata, read_metadata_for_layer
```
normanrz: Why are these references required? Couldn't you access the function like this `wkcuber.metadata.read_metadata_for_layer`?

PR #38, opened 2018-11-21
Implement downsampling for multi-channel wkw data

philippotto:
This PR implements downsampling for multi-channel wkw files.
The cubing itself also had an error for batch sizes != 1 (which would be the most likely scenario) which I fixed.

### DIFF
--- a/wkcuber/downsampling.py
+++ b/wkcuber/downsampling.py
@@ -172,6 +172,7 @@ def downsample_cube_job(
     logging.info("Downsampling of {}".format(target_cube_xyz))
 
     try:
+        time_start("Downsampling of {}".format(target_cube_xyz))
         header_block_type = (
             wkw.Header.BLOCK_TYPE_LZ4HC if compress else wkw.Header.BLOCK_TYPE_RAW
         )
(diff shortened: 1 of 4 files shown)


wkcuber/metadata.py
```diff
@@ -86,7 +86,13 @@ def detect_dtype(dataset_path, layer, mag=1):
     layer_path = path.join(dataset_path, layer, str(mag))
     if path.exists(layer_path):
         with wkw.Dataset.open(layer_path) as dataset:
-            return str(np.dtype(dataset.header.voxel_type))
+            voxel_type = dataset.header.voxel_type
+            num_channels = dataset.header.num_channels
+            voxel_size = np.dtype(voxel_type)
+            if voxel_size == np.uint8 and num_channels > 1:
+                return "uint" + str(8 * num_channels)
```
philippotto: @normanrz The code for writing the element class into the meta data is quite ugly. Do you have a better idea here? WK expects "uint24" for multi-channel uint8 data. However, numpy doesn't even know uint24 as a data type. Maybe we should change WK's interpretation instead (the meta data could include voxel type and channel count independent of each other? However, backwards compatibility would break then.).

wkcuber/downsampling.py
```diff
@@ -176,66 +176,71 @@ def downsample_cube_job(
...
+                            buffer_end = buffer_offset + tile_length
+
+                            file_buffer[
+                                channel_index,
+                                buffer_offset[0] : buffer_end[0],
+                                buffer_offset[1] : buffer_end[1],
+                                buffer_offset[2] : buffer_end[2],
+                            ] = data_cube
+
+                time_start("Downsampling of {}".format(target_cube_xyz))
```
normanrz: This `time_start` seems misplaced (prob from other PR)

wkcuber/metadata.py
normanrz: I wouldn't want to change the wk interpretation. This metadata is redundant anyways, because the information is also stored in the wkw header.

PR #43, opened 2018-11-28
Don't use f string, making wkcuber python 3.5 compatible

georgwiese:
Using an f-string made this lib incompatible with python 3.5. We still use python 3.5 `segmentation-tools`, so it would be nice to restore python 3.5 compatibility.


philippotto:
The PR certainly works for now, but we should think about how we can avoid this in the future. Probably, the python version for cuber should be officially set to 3.5 (so that nobody uses 3.6 features). When the segtools are ready, both repositories should be upgraded to 3.6. Maybe @normanrz has a thought on that?

PR #55, opened 2019-01-14
Mag

### LINKED ISSUE #54: Consistently support anisotropic mags
The cuber now supports creating anisotropic mags, but it fails to handle them appropriately in many places, e.g. compressing them, generating the `datasource-properties.json`, and other things.

For example, searching for `str(mag)` in the codebase should point you to a lot of places where it doesn't properly convert an anisotropic mag to its string representation.

valentin-pinkau:
This PR introduces a helper class for magnifications in order to support different representations more conveniently.

Mag can be created from strings, an int or a list of ints. Sanity checks like the power of two are checked as well.

closes #54


wkcuber/metadata.py
```diff
@@ -95,14 +97,14 @@ def detect_dtype(dataset_path, layer, mag=1):
                 return str(np.dtype(voxel_type))
 
 
-def detect_cubeLength(dataset_path, layer, mag=1):
+def detect_cubeLength(dataset_path, layer, mag: Mag):
```
normanrz: `mag: Mag = Mag(1)`?

wkcuber/downsampling.py
```diff
@@ -423,18 +394,17 @@ def downsample_mag(
...
+    max_mag: Mag,
     dtype,
     interpolation_mode,
     cube_edge_len,
     jobs,
     compress,
 ):
-    assert log2(from_mag) % 1 == 0, "'from_mag' needs to be power of 2."
-    target_mag = from_mag * 2
+    target_mag = from_mag.scaled_by(2)
```
georgwiese: We might want to implement something like "Go from mag 1 to mag 16-16-4", and the code would figure out that the mags are `2-2-1`, `4-4-1`, `8-8-2`, `16-16-4`.

wkcuber/mag.py
```diff
@@ -0,0 +1,62 @@
+import re
+from math import log2
+from functools import total_ordering
+
+
+@total_ordering
+class Mag:
```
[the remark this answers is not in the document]
valentin-pinkau: how about `divide(coord)`?

wkcuber/downsampling.py
valentin-pinkau: Yeah, I thought about this too. I don't know how much this feature is needed, but would be cleaner anyway.

wkcuber/mag.py
```diff
@@ -0,0 +1,63 @@
...
+    def __init__(self, mag):
+        self.mag = None
+
+        if isinstance(mag, int):
+            self.mag = [mag] * 3
+        elif isinstance(mag, list):
```
georgwiese: Pls also handle tuples.

PR #67, opened 2019-04-03
Integrate clustertools

philippotto:
Most notable, the executor won't catch failed futures automatically, as the old `ParallelExecutor` did. This is according to the common concurrent.futures.executor interface. I added an explicit `wait_and_ensure_success` call to handle errors.

Also, the locking logic was removed in favor of ensuring the existence of a dataset before distributing the jobs.

I tried to not change the interface of most of the functions (e.g., `cubing(...)`)  so that external libraries don't break. For that reason, the code around `args` is a bit verbose.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,3 +6,4 @@ pytest
 wkw>=0.0.6
 requests
 black
+git+git://github.com/scalableminds/cluster_tools@master#egg=cluster_tools
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,4 +6,4 @@ pytest
 wkw>=0.0.6
 requests
 black
-git+git://github.com/scalableminds/cluster_tools@master#egg=cluster_tools
+git+git://github.com/scalableminds/cluster_tools@v1.0#egg=cluster_tools
--- a/wkcuber/compress.py
+++ b/wkcuber/compress.py
@@ -10,10 +10,11 @@ from .mag import Mag
 
 from .utils import (
     add_verbose_flag,
-    add_jobs_flag,
     open_wkw,
     WkwDatasetInfo,
-    ParallelExecutor,
+    add_distribution_flags,
+    get_executor_for_args,
+    wait_and_ensure_success,
 )
 from .metadata import detect_resolutions
 from typing import List
@@ -44,8 +45,8 @@ def create_parser():
         "--mag", "-m", nargs="*", help="Magnification level", default=None
     )
 
-    add_jobs_flag(parser)
     add_verbose_flag(parser)
+    add_distribution_flags(parser)
 
     return parser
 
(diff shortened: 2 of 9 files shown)


requirements.txt
```diff
@@ -6,3 +6,4 @@ pytest
 wkw>=0.0.6
 requests
 black
+git+git://github.com/scalableminds/cluster_tools@master#egg=cluster_tools
```
normanrz: Should we pin to a release or tag?

wkcuber/utils.py
```diff
@@ -74,48 +77,66 @@ def get_regular_chunks(min_z, max_z, chunk_size):
...
+    executor = None
 
-class ParallelExecutor:
-    def __init__(self, jobs):
-        self.lock = Lock()
-        if python_version() >= "3.7.0":
-            self.exec = ProcessPoolExecutor(
-                jobs, initializer=pool_init, initargs=(self.lock,)
+    if args.distribution_strategy == "multiprocessing":
+        if args.jobs is None:
```
normanrz: Is this mandatory or could we use a sensible default (e.g. num of cpus)?

wkcuber/compress.py
```diff
@@ -80,13 +81,21 @@ def compress_mag(source_path, layer_name, target_path, mag: Mag, jobs):
...
+            futures = []
+            for file in source_wkw.list_files():
+                rel_file = path.relpath(file, source_wkw.root)
+                futures.append(
+                    executor.submit(
+                        compress_file_job, file, path.join(target_mag_path, rel_file)
+                    )
+                )
+
+            wait_and_ensure_success(futures)
```
normanrz: Would a `executor.map(...)` catch potential errors?

philippotto: Yes, it would. However, the map interface seemed a bit cumbersome to me since the "preparation code" for one submit is > 10 lines long. We could make a for-loop which pushes the task descriptors into a list and then map over that. However, I think it makes sense to establish a good precedent case on how to properly use `submit` for the concurrent.futures.executor interface. Also, it makes the success-checking explicit. For `map`, the success checking is a quite unobvious side-effect, in my opinion.

wkcuber/utils.py
philippotto: Good point, the default is already num of cpus, so there is no sense in checking for None here. I removed the code.

requirements.txt
philippotto: Good point, done.

wkcuber/compress.py
normanrz: Fine with me.

wkcuber/utils.py
```diff
@@ -74,48 +77,61 @@ def get_regular_chunks(min_z, max_z, chunk_size):
...
-    def __init__(self, jobs):
-        self.lock = Lock()
-        if python_version() >= "3.7.0":
-            self.exec = ProcessPoolExecutor(
-                jobs, initializer=pool_init, initargs=(self.lock,)
+    if args.distribution_strategy == "multiprocessing":
+        executor = cluster_tools.get_executor("multiprocessing", args.jobs)
+        logging.info("Using pool of {} workers.".format(args.jobs))
+    elif args.distribution_strategy == "slurm":
+        if args.job_resources is None:
```
normanrz: Similar question here: Should this be required?

Issue #68, opened 2019-04-10
Add reusable BufferedWriter


georgwiese:
Similar to what is implemented in alignment tools

MartinBuessemeyer:
AlignmentTools: BufferedSliceWriter, move to wkcuber

after the PR is merged. go to segmentation tools and remove from alignment tools and import it.

philippotto:
> after the PR is merged. go to segmentation tools and remove from alignment tools and import it.

@MartinBuessemeyer the second part is actionable now, I think :)

philippotto:
@normanrz The BufferedSliceWriter wasn't removed from the segmentation tools, yet. Also, I think the implementations have diverged in the mean time. Not sure whether we want to pick this up again or whether we want to unify it at a later point.

normanrz:
I think it would be good to unify both implementations and only use the one in this repo.

PR #72, opened 2019-04-22
Add BufferedSliceWriter from segmentation tools

MartinBuessemeyer:
This PR will add the BufferedSliceWriter from `segmentation-tools` to the `utils.py`. Therefore it will complete the first part of #68.

Is `utils.py` the right file or is it better to create a new file?

TODO: Testing


wkcuber/utils.py
```diff
@@ -152,3 +152,84 @@ def time_stop(identifier):
...
+    def write_slice(self, z: int, data: np.ndarray):
+        """Takes in a slice in [y, x] shape, writes to WKW file."""
+
+        if len(self.buffer) == 0:
+            self.current_z = z
+            self.buffer_start_z = z
+
+        assert (
+            z == self.current_z
+        ), "({}) Slices have to be written sequentially!".format(getpid())
```
normanrz: I don't think we need to print the pid in the logs

MartinBuessemeyer:
How should I do the testing? Or can I skip them since this is just a copy of the class from the `segmentation tools`?

wkcuber/utils.py
MartinBuessemeyer: Done :grin:

normanrz:
I think a test would be useful. Maybe something like:
- Write 13 slices -> nothing should be written to disk yet
- Write 19 more slices -> 32 slices on disk
- Write 3 more slices -> still only 32 slices on disk
- Close the writer -> 35 slices on disk

MartinBuessemeyer:
Has someone an idea of why the test fails? Or to be precise: Why is there data on the disk?

normanrz:
Have tried reproducing that issue locally or on one our servers?

normanrz:
wkw will always return data, if you do `dataset.read(..)`. if there is no actual data, you'll get a bunch of zeros

MartinBuessemeyer:
> wkw will always return data, if you do `dataset.read(..)`. if there is no actual data, you'll get a bunch of zeros

Thanks a lot.

Fixed the test accordingly.

Additionally, I made the `Magnification` of the `BufferedSliceWriter` to an optional argument because I think that we might want to write in another Magnification in the future.
Should be ready for the review now :)

hotzenklotz:
@philippotto Bump. Review required.

tests/test_utils.py
```diff
@@ -27,3 +30,49 @@ def test_get_regular_chunks_max_inclusive():
...
+                pass
+
+        for i in range(13, 32):
+            writer.write_slice(i, test_img)
+        with wkw.Dataset.open(dataset_path, wkw.Header(dtype)) as data:
+            read_data = data.read(origin, (24, 24, 32))
+            assert np.squeeze(read_data).shape == (24, 24, 32), "The read data should have the shape: (24, 24, 32) " \
+                                                                "but has a shape of: {}"\
+                                                                .format(np.squeeze(read_data).shape)
+            assert read_data.size == read_data[read_data.nonzero()].size, "The read data contains zeros while the " \
```
philippotto: Nice assertion!

tests/test_utils.py
```diff
@@ -27,3 +30,49 @@ def test_get_regular_chunks_max_inclusive():
...
+            writer.write_slice(i, test_img)
+
+    with wkw.Dataset.open(dataset_path, wkw.Header(dtype)) as data:
+        read_data = data.read(origin, (24, 24, 35))
+        assert np.squeeze(read_data).shape == (24, 24, 35), "The read data should have the shape: (24, 24, 35) " \
+                                                                "but has a shape of: {}"\
+                                                                .format(np.squeeze(read_data).shape)
+        assert read_data.size == read_data[read_data.nonzero()].size, "The read data contains zeros while the " \
+                                                                          "written image has no zeros"
+
```
philippotto: Wouldn't it be useful to have an assertion which checks that `test_img` and `read_data` hold the same data?

wkcuber/utils.py
```diff
@@ -152,3 +153,89 @@ def time_stop(identifier):
...
+        self.buffer_start_z = None
+
+    def write_slice(self, z: int, data: np.ndarray):
+        """Takes in a slice in [y, x] shape, writes to WKW file."""
+
+        if len(self.buffer) == 0:
+            self.current_z = z
+            self.buffer_start_z = z
+
+        assert z == self.current_z, "Slices have to be written sequentially!"
```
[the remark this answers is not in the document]
MartinBuessemeyer: OK. I reverted the change :)

tests/test_utils.py
MartinBuessemeyer: Just added it :D

MartinBuessemeyer:
Should I merge?

philippotto:
Sure!

PR #74, opened 2019-04-29
Fix get_executor invocation

philippotto:
The argument max_workers arg has to be named and not positional.


philippotto:
@normanrz the circleci build has finished, but I cannot merge the pr. do you know what's going on? I found this page here, but I'm not sure what the actual problem is, since it worked before: https://circleci.com/docs/2.0/workflows-waiting-status/

PR #75, opened 2019-04-30
Also detect prediction layer & add refresh_metadata function

philippotto:
Useful for the segmentation-tools. In the long run, we should rewrite the code a bit so that the caller is more flexible when generating datasource-properties.json files.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Created with [Python3](https://www.python.org/).
 * `wkcuber.convert_knossos`: Convert KNOSSOS cubes to WKW cubes
 * `wkcuber.downsampling`: Create downsampled magnifications (with `median`, `mode` and linear interpolation modes)
 * `wkcuber.compress`: Compress WKW cubes for efficient file storage (especially useful for segmentation data)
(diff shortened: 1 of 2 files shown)


philippotto:
Additionally, I added a `refresh_metadata` function which can be used to re-create the datasource-properties.json without having to provide properties, such as `scale`, again. At the moment, the layer information is re-created completely. In the future, we could make this smarter so that manually edited properties are not lost. However, for now, I think it's not necessary.

philippotto:
> At the moment, the layer information is re-created completely. In the future, we could make this smarter so that manually edited properties are not lost. However, for now, I think it's not necessary.

Ok, it turns out: the future is today. I changed the code to only overwrite the resolutions for existing layers. Otherwise, callers would need to always pass the tight bbox.

philippotto:
@normanrz ping :)

normanrz:
A cli would be also great for this.

philippotto:
> A cli would be also great for this.

Good point. I added that now (the user can pass `--refresh` in which case `--name` and `--scale` are not necessary). I couldn't get argparse exclusive group to work with nested sub groups, which is why I had to resort to manual verification of the arguments.

Issue #76, opened 2019-05-02
Metadata dataset name must be non-null


normanrz:
If the user fails to supply a name for the dataset in `wkcuber` or `wkcuber.metadata`, `null` will be written into the json. This is not permitted by webKnossos. Therefore, the default should be something else, e.g. empty string or directory name.

PR #85, opened 2019-05-13
Set default wkw file name

MartinBuessemeyer:
This PR addresses #76.

I decided to take the directory name as the `name` if a `name` is not provided. I decided to make the change in `metadata.py/write_webknossos_metadata` because this will directly fix it as well when `wkcuber` is used as a library e.g. `segmentation-tools`.


MartinBuessemeyer:
Please take a look at `metadata.py ln 290`. Should I remove that `assert statement` since we got a default name?

philippotto:
> Please take a look at `metadata.py ln 290`. Should I remove that `assert statement` since we got a default name?

I'd leave the assertion there. In general, we want the user to specify a proper name. They will get immediate feedback if the name is missing (unlike to the library usage for which you provided a fallback :+1:).

PR #88, opened 2019-05-17
Add anisotropic parameter to downsampling tool

### LINKED ISSUE #64: Support anisotropic downscaling in wkcuber.downsampling
The default tool `python -m wkcuber ...` should also support anisotropic downscaling.
I think a new boolean flag `--anisotropic_downsampling` should be enough. The downsampling configuration should be autodetected based on the voxel size.

philippotto:
Additionally, I changed the default parameter for "compressed" to true. This is already the case for the main script. I'm not aware of any strong arguments against this? @normanrz 

Fixes #64.

### DIFF
--- a/wkcuber/downsampling.py
+++ b/wkcuber/downsampling.py
@@ -22,7 +22,7 @@ from .utils import (
     add_distribution_flags,
     get_executor_for_args,
     wait_and_ensure_success,
-    add_anisotropic_flag
+    add_anisotropic_flag,
 )
 
 DEFAULT_EDGE_LEN = 256
@@ -79,10 +79,10 @@ def create_parser():
     )
     group.add_argument(
         "--anisotropic_target_mag",
-        help="Specify an explicit anisotropic target magnification which should be " \
-            "created (e.g., --anisotropic_target_mag 2-2-1). Consider using --anisotropic " \
-            "instead which automatically creates multiple anisotropic magnifications depending " \
-            "on the dataset's scale",
+        help="Specify an explicit anisotropic target magnification which should be "
+        "created (e.g., --anisotropic_target_mag 2-2-1). Consider using --anisotropic "
+        "instead which automatically creates multiple anisotropic magnifications depending "
+        "on the dataset's scale",
         type=str,
     )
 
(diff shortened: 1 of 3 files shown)


wkcuber/downsampling.py
```diff
@@ -90,10 +95,11 @@ def create_parser():
     )
 
     parser.add_argument(
-        "--compress", action="store_true", help="Compress data during downsampling"
+        "--compress", action="store_true", help="Compress data during downsampling", default=True
```
valentin-pinkau: how do I deactivate compression?

wkcuber/downsampling.py
```diff
@@ -90,10 +95,14 @@ def create_parser():
     )
 
     parser.add_argument(
-        "--compress", action="store_true", help="Compress data during downsampling"
+        "--compress",
+        action="store_true",
+        help="Compress data during downsampling",
+        default=True,
```
valentin-pinkau: how do I deactivate compression if `store_true` and `default=True`?

wkcuber/downsampling.py
```diff
@@ -533,6 +542,28 @@ def detect_larger_and_smaller_dimension(scale):
...
+                "to find target magnifications for anisotropic downsampling. "
+                "Does the provided dataset have a datasource-properties.json file?"
+            )
+            raise exc
+
+        downsample_mags_anisotropic(
+            args.path,
+            args.layer_name,
+            from_mag,
+            max_mag,
```
valentin-pinkau: the `max_mag` is isotropic. How do I specify to stop at e.g. 1024-1024-512?

wkcuber/downsampling.py
philippotto: Very valid point. I inverted the argument (the same as the main script now).

wkcuber/downsampling.py
philippotto: The anisotropic parameter decides for itself which mags specifically should be created. It will stop when the current mag is equal or greater than the target mag. For the greater/equal comparison, the maximum dim of the both mags are used. Therefore, you can only control that the "poorest" resolution should not exceed 1024. I added a comment about that to the argparse help (see my latest commit). I hope this is enough for now.

Issue #77, opened 2019-05-03
Export tiff stack


normanrz:
It would be cool to have another tool that exports a tiff stack from a wkw dataset.

philippotto:
@MartinBuessemeyer Isn't this done? Please make sure to always close your issues after you have finished them. You can write `fixes #77` in the corresponding PR and then the issue is closed as soon as your PR is merged.

PR #91, opened 2019-05-20
Export wkw as tiff

MartinBuessemeyer:
This PR will add a script that exports wkw files to tiff files. It addresses #77.

Tests will be added soon.

### DIFF
--- a/tests/test_export_wkw_as_tiff.py
+++ b/tests/test_export_wkw_as_tiff.py
@@ -1,0 +1,49 @@
+from wkcuber.export_wkw_as_tiff import export_tiff_stack, wkw_name_and_bbox_to_tiff_name, create_parser
+import os
+from PIL import Image
+from wkcuber.mag import Mag
+import wkw
+import numpy as np
+
+def test_export_tiff_stack():
+    parser = create_parser()
+    args_list = ["--source_path", os.path.join("testdata", "WT1_wkw"),
+                 "--destination_path", os.path.join("testoutput", "WT1_wkw"),
+                 "--layer_name", "color",
+                 "--name", "text_export",
+                 "--axis", "z",
+                 "--bbox", "0,0,0,100,100,5",
+                 "--mag", "1"]
+    args = parser.parse_args(args_list)
+
+    bbox = [int(s.strip()) for s in args.bbox.split(",")]
+    bbox = {"topleft": bbox[0:3], "size": bbox[3:6]}
+
+    export_tiff_stack(
+        wkw_file_path=args.source_path,
+        wkw_layer=args.layer_name,
+        bbox=bbox,
+        mag=Mag(args.mag),
+        destination_path=args.destination_path,
+        name=args.name,
+        axis=args.axis,
+        args=args,
+    )
+
+    test_wkw_file_path = os.path.join("testdata", "WT1_wkw", "color", Mag(1).to_layer_name())
+    with wkw.Dataset.open(test_wkw_file_path) as dataset:
+        slice_bbox = bbox
+        slice_bbox["size"] = [slice_bbox["size"][0], slice_bbox["size"][1], 1]
+        for slice in range(bbox["size"][2]):
+            slice_bbox["offset"] = [slice_bbox["topleft"][0], slice_bbox["topleft"][1], bbox["topleft"][2] + slice]
+            tiff_path = os.path.join(args.destination_path, wkw_name_and_bbox_to_tiff_name(args.name, slice_bbox))
+
+            assert os.path.isfile(tiff_path), f"Expected a tiff to be written at: {tiff_path}."
+
+            test_image = np.array(Image.open(tiff_path))
+            test_image.transpose((1, 0))
+
+            correct_image = dataset.read(off=slice_bbox["topleft"], shape=slice_bbox["size"])
+
+            assert np.array_equal(correct_image, test_image), f"The tiff file {tiff_path} that was written is not " \
+                                                              f"equal to the original wkw_file."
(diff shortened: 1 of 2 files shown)


philippotto:
@MartinBuessemeyer Please assign me as a reviewer when the pr is ready for review (i.e., not WIP) :)

MartinBuessemeyer:
Test has been added

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,171 @@
...
+
+    image = Image.fromarray(tiff_data)
+    image.save(tiff_file_path)
+
+
+def export_tiff_stack(
+    wkw_file_path, wkw_layer, bbox, mag, destination_path, name, axis, args
+):
+    if not os.path.isdir(destination_path):
+        os.mkdir(destination_path)
```
philippotto: You can use `os.makedirs(destination_path, exist_ok=True)` to get rid of the if condition.

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,171 @@
...
+        default="color",
+    )
+
+    parser.add_argument("--name", "-n", help="Name of the tiffs", default=None)
+
+    parser.add_argument(
+        "--axis",
+        "-a",
+        help="The axis that the image should be generated along. "
+        "Thus choosing z will print x,y slices.",
```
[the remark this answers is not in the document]
philippotto: On a second thought, I'm not sure whether we really need this feature. Did you have a specific use case in mind when you wrote the code? I think, using z by default is sufficient for us (the cubing process also always assumes z as the default axis). I cannot think of a reason to use another axis :thinking:

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,171 @@
...
+            tiff_bbox["topleft"][2],
+        ]
+        tiff_bbox["size"] = [tiff_bbox["size"][0], 1, tiff_bbox["size"][2]]
+    else:
+        tiff_bbox["topleft"] = [
+            tiff_bbox["topleft"][0],
+            tiff_bbox["topleft"][1],
+            tiff_bbox["topleft"][2] + slice_number,
+        ]
+        tiff_bbox["size"] = [tiff_bbox["size"][0], tiff_bbox["size"][1], 1]
```
philippotto: As mentioned above, I think we can get rid of the axis-dependent code.

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,171 @@
...
+            axis_index = 2
+        num_slices = bbox["size"][axis_index]
+        slices = range(num_slices)
+        export_args = zip(
+            slices,
+            [bbox] * num_slices,
+            [destination_path] * num_slices,
+            [name] * num_slices,
+            [axis] * num_slices,
+            [dataset_path] * num_slices,
```
philippotto: How about joining these arguments into one tuple and multiplying that with num_slices? Like so `(bbox, destination_path, name, ...) * num_slices`

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,171 @@
...
+
+    return parser
+
+
+def wkw_name_and_bbox_to_tiff_name(
+    name: str, bbox: Dict[str, Tuple[int, int, int]]
+) -> str:
+    return (
+        f'{name}_topleft_{bbox["topleft"][0]}_{bbox["topleft"][1]}_{bbox["topleft"][2]}'
+        f'_size_{bbox["topleft"][0]}_{bbox["topleft"][1]}_{bbox["topleft"][2]}.tiff'
```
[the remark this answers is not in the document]
philippotto: I'm not sure whether it's a good idea to have the full bounding box in each file name. I feel, like it *suggests* that there is a lot of important information in the file names, but actually, it's not really important (and also, it doesn't differ between the slices). I'd be more comfortable with something simple, such as `{slice_index}.tiff`. That would be super easy to grasp for the user.

wkcuber/export_wkw_as_tiff.py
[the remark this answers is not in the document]
MartinBuessemeyer: I thought this tool is used to export some part of a wkw-file to tiff files and since you might not always just look at x,y-slices, I decided to add that feature.

wkcuber/export_wkw_as_tiff.py
MartinBuessemeyer: Nice :+1:

wkcuber/export_wkw_as_tiff.py
MartinBuessemeyer: you need to add the additional brackets. Then it is working. `[(bbox, destination_path, name, dataset_path)] * num_slices)
`

MartinBuessemeyer:
Adapted the Feedback. Thanks for providing it :)

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,132 @@
...
+    add_distribution_flags(parser)
+
+    return parser
+
+
+def wkw_name_and_bbox_to_tiff_name(name: str, slice_index: int) -> str:
+    if name is None or name == "":
+        return f"{slice_index}.tiff"
+    else:
+        return f"name_{slice_index}.tiff"
```
philippotto: This should probably be `return f"{name}_{slice_index}.tiff"`.

tests/test_export_wkw_as_tiff.py
```diff
@@ -0,0 +1,50 @@
...
+import wkw
+import numpy as np
+
+def test_export_tiff_stack():
+    parser = create_parser()
+    args_list = ["--source_path", os.path.join("testdata", "WT1_wkw"),
+                 "--destination_path", os.path.join("testoutput", "WT1_wkw"),
+                 "--layer_name", "color",
+                 "--name", "text_export",
+                 "--axis", "z",
```
philippotto: The axis parameter does not exist anymore, right? Also see line 28.

MartinBuessemeyer:
Since the tiling extension is already implemented, this PR is obsolete.

tests/test_export_wkw_as_tiff.py
MartinBuessemeyer: That's why the tests failed :D

PR #93, opened 2019-05-24
Export wkw as tiff with tiling support

MartinBuessemeyer:
This PR will add a script that exports wkw files to tiff files and supports tiling. It addresses #77 and replaces #91.

There are still no tests for the tiling.


wkcuber/export_wkw_as_tiff.py
```diff
@@ -58,56 +78,95 @@ def wkw_name_and_bbox_to_tiff_name(name: str, slice_index: int) -> str:
...
-        return f"name_{slice_index}.tiff"
+        return f"{name}_{slice_index}.tiff"
+
+
+def wkw_slice_to_image(data_slice: np.ndarray):
+    # discard the z dimension
+    data_slice = data_slice.squeeze(axis=3)
+    if data_slice.shape[0] == 1:
+        # discard greyscale dimension
+        data_slice = data_slice.squeeze(axis=0)
```
philippotto: What would happen if you didn't do this? Would it create a multi-channel tiff file even though there is only one channel? I would have thought that this isn't a problem :thinking:

README.md
```diff
@@ -11,6 +11,7 @@ Created with [Python3](https://www.python.org/).
 ## Features
 
 * `wkcuber`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
+* `wkcuber.export_wkw_as_tiff`: Convert WKW datasets to a tiff stack (includes tiling to `z/y/x.tiff` folder structure)
```
philippotto: "includes" could make it sound as if tiff export would always be tiling. I'd write `... (writing as tiles to a ... folder structure is also supported)`

wkcuber/export_wkw_as_tiff.py
```diff
@@ -58,56 +78,95 @@ def wkw_name_and_bbox_to_tiff_name(name: str, slice_index: int) -> str:
...
     tiff_bbox["topleft"] = [
         tiff_bbox["topleft"][0],
         tiff_bbox["topleft"][1],
         tiff_bbox["topleft"][2] + slice_number,
     ]
     tiff_bbox["size"] = [tiff_bbox["size"][0], tiff_bbox["size"][1], 1]
 
-    tiff_file_name = wkw_name_and_bbox_to_tiff_name(name, slice_number)
+    if tiling_size is None:
+        with wkw.Dataset.open(dataset_path) as dataset:
```
philippotto: You can hoist the with statement to be above the if. Then, you only need it once instead of twice (see line 128).

wkcuber/export_wkw_as_tiff.py
```diff
@@ -58,56 +78,95 @@ def wkw_name_and_bbox_to_tiff_name(name: str, slice_index: int) -> str:
...
+                for x_tile_index in range(
+                    1, ceil((tiff_bbox["size"][0]) / tiling_size[0]) + 1
+                ):
+                    tile_tiff_filename = f"{x_tile_index}.tiff"
+                    tile_bbox["topleft"][0] += tiling_size[0]
 
-    image = Image.fromarray(tiff_data)
-    image.save(tiff_file_path)
+                    tile_tiff_data = dataset.read(
+                        tile_bbox["topleft"], tile_bbox["size"]
```
philippotto: The handling of the `tile_bbox` can be improved a bit, I think :) Mainly, I'd avoid mutating the bounding box variable, since it makes deep copy and the restoration necessary (additionally, it scatters the bbox computation across the entire loops, even though it could be done in one statement). Instead, I'd compute the `tile_bbox` from scratch where it's necessary (line 143 probably) using the `*_tile_index` variables (which I would make zero-based so that the range expressions are easier to read).

Roughly, it would look like this:
```python
tile_bbox_size = [tiling_size[0], tiling_size[1], 1]
for y_tile_index in range(ceil((tiff_bbox["size"][1]) / tiling_size[1])):
  tile_tiff_path = os.path.join(
    dest_path, str(slice_number), str(y_tile_index + 1)
  )
  os.makedirs...
  for x_tile_index in range(ceil((tiff_bbox["size"][0]) / tiling_size[0])):
    tile_tiff_filename = f"{x_tile_index + 1}.tiff"
    tile_bbox_topleft = [tiff_bbox[0] + tile_bbox_size[0] * x_tile_index, ...]
    ...
```

wkcuber/export_wkw_as_tiff.py
```diff
@@ -48,6 +59,15 @@ def create_parser():
         "--mag", "-m", help="The magnification that should be read", default=1
     )
 
+    parser.add_argument(
+        "--tiling_slice_size",
+        "-tiling_size",
+        help="This will activate tiling. Thus each z-slice will be divided"
```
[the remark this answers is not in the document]
philippotto: You can also do `group = parser.add_mutually_exclusive_group()` and add the tile parameters to that group. This, ensures that the user cannot pass both arguments (argparse will complain then).

wkcuber/export_wkw_as_tiff.py
MartinBuessemeyer: The tests do fail if I remove the check and I can remember that the method only accepts  `m x n` or `m x n x 3` images but I cannot find the spot in the doc anymore.

wkcuber/export_wkw_as_tiff.py
MartinBuessemeyer: Ok I have no excuse here why I did it that way but yes your way is definitely better than mine ans is way more readable. Thx :D

MartinBuessemeyer:
I finished implementing the feedback, the tests and fixing a handful of bugs that still existed 0.0

tests/test_export_wkw_as_tiff.py
```diff
@@ -0,0 +1,168 @@
...
+                                                      tile_bbox["topleft"][1] + tile_bbox["size"][1] * y_tile_index,
+                                                      tile_bbox["topleft"][2] + data_slice_index],
+                                                 shape=tile_bbox["size"])
+
+                    correct_image = np.squeeze(correct_image)
+
+                    assert np.array_equal(correct_image, test_image), f"The tiff file {tiff_path} that was written " \
+                        f"is not equal to the original wkw_file."
+
+                    def test_export_tiff_stack_tile_size():
```
normanrz: test in test?

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,218 @@
...
+
+
+def export_tiff_stack(
+    wkw_file_path, wkw_layer, bbox, mag, destination_path, name, tiling_slice_size, args
+):
+    os.makedirs(destination_path, exist_ok=True)
+    dataset_path = os.path.join(wkw_file_path, wkw_layer, mag.to_layer_name())
+
+    with get_executor_for_args(args) as executor:
+        num_slices = bbox["size"][2]
```
normanrz: I think it would make sense to iterate over batches of slices. Since wkw is organized in 32**3 cubes, 32 slices have to be read even if you're only interested in a single slice. Therefore, it is more efficient to read in 32 slices at once and write them out one after the other. The batch size should be 32 by default and adjustable through args. See `wkcuber.cubing` for reference.

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,265 @@
...
+        "-l",
+        help="Name of the layer that will be converted to a tiff stack",
+        default="color",
+    )
+
+    parser.add_argument("--name", "-n", help="Name of the tiffs", default="")
+
+    parser.add_argument(
+        "--bbox",
+        "-bbox",
```
normanrz: No need for a single-dash argument, since `-b` is already taken.

wkcuber/export_wkw_as_tiff.py
```diff
@@ -0,0 +1,265 @@
...
+        "--tiles_per_dimension",
+        "-t",
+        help='For very large datasets, it is recommended to enable tiling which will ensure that each slice is exported to multiple images (i.e., tiles). As a parameter you should provide the amount of tiles per dimension in the form of "x,y".'
+        'Also see at "--tile_size" to specify the absolute size of the tiles.',
+        default=None,
+    )
+
+    tiling_option_group.add_argument(
+        "--tile_size",
+        "-tile_size",
```
normanrz: Same here. Remove the single-dash arg.

wkcuber/export_wkw_as_tiff.py
normanrz: As a general rule: the single-dash args should only be single character as well (e.g. `-t`). If that's not possible then leave it out.

MartinBuessemeyer: Thanks. I didn't know that.

MartinBuessemeyer:
Thanks. I added the feedback.

philippotto:
@dependabot merge

philippotto:
@MartinBuessemeyer is this ready to merge from your side?

MartinBuessemeyer:
Yes, shall I? @philippotto

philippotto:
https://media.giphy.com/media/3ofT5Uh65LptIEWwY8/giphy.gif 

Yes :) If a PR has been accepted and there are no todos left, you can merge that PR (unless for the webknossos repository, but you are currently not working on that :)).

PR #94, opened 2019-05-27
Add pad option to cube images with differing size

philippotto:


philippotto:
@fm3 This is ready for review now. I cubed some tiffs with not common extents and the output looked ok.

Issue #92, opened 2019-05-23
Refactor lib interface for (anisotropic) downsampling


philippotto:
Right now, the anisotropic downsampling is easy to use when invoked via the CLI. However, when using wkcuber as a library, the difference between anisotropic and isotropic downsampling is a bit cumbersome. Ideally, I'd like it to be a simple flag like this:

`downsample(path, max_mag,..., anisotropic=True)`

Also, the scale should be derived by the path automatically so that the call side doesn't have to look this up.

MartinBuessemeyer:
`Also, the scale should be derived by the path automatically so that the call side doesn't have to look this up.`

I assume that you mean the `from_mag` with scale right?

PR #98, opened 2019-07-02
Better downsample lib interface

MartinBuessemeyer:
This PR tackles #92

I implemented the interface with the method `downsample_mags`. I renamed the old `downsample_mags` to `downsample_mags_isotropic`. The new `downsample_mags` takes the same arguments and an optional`anisotropic` argument. Thus it is backwards compatible with the previous `downsample_mags`.

The interface is now able to detect `from_mag`,  `layer_name` and `scale` from the file `path` and the `datasourceporperties.json`. and supplies default values for almost all arguments thus `downsample_mags(path, anisotropic=True)` is a valid invocation.

Additionally I removed some unused imports that my IDE indicated.

### DIFF
--- a/tests/scripts/anisotropic_downsampling.sh
+++ b/tests/scripts/anisotropic_downsampling.sh
@@ -10,6 +10,7 @@ docker run \
   --buffer_cube_size 128 \
   --layer_name color \
   /testoutput/tiff
+
 docker run \
   -v "${PWD}/testoutput:/testoutput" \
   --rm \
@@ -25,3 +26,20 @@ docker run \
 [ -d testoutput/tiff/color/4-4-1 ]
 [ $(find testoutput/tiff/color/2-2-1 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
 [ $(find testoutput/tiff/color/4-4-1 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
+
+docker run \
+  -v "${PWD}/testoutput:/testoutput" \
+  --rm \
+  scalableminds/webknossos-cuber:${CIRCLE_BUILD_NUM} \
+  wkcuber.downsampling \
+  --jobs 2 \
+  --from 4-4-1 \
+  --max 16 \
+  --buffer_cube_size 128 \
+  --layer_name color \
+  /testoutput/tiff
+[ -d testoutput/tiff/color/2-2-1 ]
+[ -d testoutput/tiff/color/4-4-1 ]
+[ $(find testoutput/tiff/color/2-2-1 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
+[ $(find testoutput/tiff/color/4-4-1 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
+
--- a/tests/scripts/anisotropic_downsampling.sh
+++ b/tests/scripts/anisotropic_downsampling.sh
@@ -39,8 +39,9 @@ docker run \
   --layer_name color \
   --scale 11.24,11.24,25 \
   /testoutput/tiff
-[ -d testoutput/tiff/color/2-2-1 ]
-[ -d testoutput/tiff/color/4-4-1 ]
-[ $(find testoutput/tiff/color/2-2-1 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
-[ $(find testoutput/tiff/color/4-4-1 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
+[ -d testoutput/tiff/color/8-8-2 ]
+[ -d testoutput/tiff/color/16-16-4 ]
+[ $(find testoutput/tiff/color/8-8-2 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
+[ $(find testoutput/tiff/color/16-16-4 -mindepth 3 -name "*.wkw" | wc -l) -eq 1 ]
+
 
--- a/tests/scripts/anisotropic_downsampling.sh
+++ b/tests/scripts/anisotropic_downsampling.sh
@@ -37,6 +37,7 @@ docker run \
   --max 16 \
   --buffer_cube_size 128 \
   --layer_name color \
+  --scale 11.24,11.24,25 \
   /testoutput/tiff
 [ -d testoutput/tiff/color/2-2-1 ]
 [ -d testoutput/tiff/color/4-4-1 ]
--- a/tests/test_downsampling.py
+++ b/tests/test_downsampling.py
@@ -1,204 +1,204 @@
-# import logging
-# import numpy as np
-# from wkcuber.downsampling import (
-#     InterpolationModes,
-#     downsample_cube,
-#     downsample_cube_job,
-#     cube_addresses,
-#     get_next_anisotropic_mag,
-# )
-# import wkw
-# from wkcuber.mag import Mag
-# from wkcuber.utils import WkwDatasetInfo, open_wkw
-# from wkcuber.downsampling import _mode, non_linear_filter_3d
-# import shutil
-#
-# WKW_CUBE_SIZE = 1024
-# CUBE_EDGE_LEN = 256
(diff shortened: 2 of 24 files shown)


wkcuber/__main__.py
```diff
@@ -78,7 +75,7 @@ def create_parser():
         )
 
     else:
-        downsample_mags(
+        downsample_mags_isotropic(
```
normanrz: I think the default should be anisotropic downsampling.

MartinBuessemeyer: Yes I think so too but this may change the result of the other code since now all data sets that were scaled isotropic before will now be scaled anisotropic.

normanrz: This `__main__.py` is only run from the command line not from other code.

tests/test_downsampling.py
```diff
@@ -1,204 +1,204 @@
...
-        [(10.5, 24, 10.5),      Mag((2, 1, 2)), Mag((4, 1, 4))],
-        [(24, 10.5, 10.5),      Mag((1, 2, 2)), Mag((1, 4, 4))],
-        [(10.5, 10.5, 10.5),    Mag(2),         Mag(4)],
-    ]
-    for i in range(len(mag_tests)):
-        assert mag_tests[i][2] == get_next_anisotropic_mag(mag_tests[i][1], mag_tests[i][0]), "The next anisotropic" \
-                                                                                          " Magnification of {} with " \
-                                                                                          "the size {} should be {}"\
-            .format(mag_tests[i][1], mag_tests[i][0], mag_tests[i][2])
+# import logging
```
normanrz: Comments?

wkcuber/downsampling.py
```diff
@@ -3,16 +3,14 @@
...
 from argparse import ArgumentParser
-from math import floor, log2
-from os import path, listdir
+import os
 from scipy.ndimage.interpolation import zoom
 from itertools import product
-from functools import lru_cache
 from enum import Enum
-from .mag import Mag
+from wkcuber.mag import Mag
```
normanrz: Why no relative imports?

wkcuber/downsampling.py
```diff
@@ -447,6 +445,61 @@ def downsample_mag(
...
+
+    if anisotropic:
+        if not scale:
+            try:
+                scale = read_datasource_properties(args.path)["scale"]
+            except Exception as exc:
+                logging.error(
+                    f"Could not get the scale from the datasource-properties.json. Probably your path is wrong. "
+                    "If you do not provide the layer_name or from_mag, they need to be included in the path."
+                    "(e.g. dataset/color/1). Other wise the path should just point at the dataset directory."
```
normanrz: Otherwise

wkcuber/downsampling.py
```diff
@@ -447,6 +445,61 @@ def downsample_mag(
...
+    scale = getattr(args, "scale", None) if args else None
+    if not layer_name or not from_mag:
+        layer_name = os.path.basename(os.path.dirname(path))
+        from_mag = Mag(os.path.basename(path))
+        path = os.path.dirname(os.path.dirname(path))
+        if anisotropic:
+            scale = read_datasource_properties(path)["scale"]
+
+    if anisotropic:
+        if not scale:
```
normanrz: `if scale is None`

wkcuber/utils.py
```diff
@@ -62,14 +62,15 @@ def add_verbose_flag(parser):
...
 
 
-def add_anisotropic_flag(parser):
+def add_isotropic_flag(parser):
     parser.add_argument(
-        "--anisotropic",
-        "-a",
+        "--isotropic",
+        "-iso",
         help="Activates Anisotropic downsampling. It will detect which dimension is the smallest and the largest. "
```
normanrz: Please adjust help text.

tests/test_downsampling.py
MartinBuessemeyer: commented the test out since it takes to much time.

wkcuber/downsampling.py
MartinBuessemeyer: I'm having issues with the remote docker interpreter path. It does not really find the relative files. Using absolute imports are fixing the issue. I reverted the change.

tests/scripts/anisotropic_downsampling.sh
```diff
@@ -25,3 +26,21 @@ docker run \
...
+  wkcuber.downsampling \
+  --jobs 2 \
+  --from 4-4-1 \
+  --max 16 \
+  --buffer_cube_size 128 \
+  --layer_name color \
+  --scale 11.24,11.24,25 \
+  /testoutput/tiff
+[ -d testoutput/tiff/color/2-2-1 ]
+[ -d testoutput/tiff/color/4-4-1 ]
```
philippotto: Shouldn't you assert that `8-8-1` was created?

wkcuber/downsampling.py
```diff
@@ -447,6 +452,61 @@ def downsample_mag(
...
+    max_mag: Mag = Mag(32),
+    interpolation_mode: str = "default",
+    cube_edge_len: int = DEFAULT_EDGE_LEN,
+    compress: bool = True,
+    args=None,
+    anisotropic: bool = True,
+):
+    scale = getattr(args, "scale", None) if args else None
+    if not layer_name or not from_mag:
+        layer_name = os.path.basename(os.path.dirname(path))
```
philippotto: So, if one of the attributes `layer_name` and `from_mag` was not provided, you infer both? This means, that if I only provide one of these args, the provided one will be overwritten, nevertheless. This is a bit weird. I guess the reasoning behind this is that the user can either specify the dataset path WITH or without layername/mag.

Maybe add an assertion which enforces that either both or none of the parameters (layername and mag) should be provided.

wkcuber/downsampling.py
```diff
@@ -506,6 +566,9 @@ def downsample_mags_anisotropic(
 def get_next_anisotropic_mag(mag, scale):
     max_index, min_index = detect_larger_and_smaller_dimension(scale)
     mag_array = mag.to_array()
+    logging.info(scale)
```
philippotto: Remove again?

wkcuber/utils.py
```diff
@@ -62,14 +62,14 @@ def add_verbose_flag(parser):
...
+def add_isotropic_flag(parser):
     parser.add_argument(
-        "--anisotropic",
-        "-a",
-        help="Activates Anisotropic downsampling. It will detect which dimension is the smallest and the largest. "
-        "The largest dimension will only be downsampled by 2 if it would be smaller or equal to the smallest "
-        "dimension in the next downsampling step.",
-        dest="anisotropic",
+        "--isotropic",
+        "-iso",
```
philippotto: Single-dash parameters should only have a single character. Doublecheck whether `-i` is still available.

wkcuber/utils.py
```diff
@@ -62,14 +62,14 @@ def add_verbose_flag(parser):
...
+        help="Activates isotropic downsampling. The default is anisotropic downsampling. "
+        "Therefore it will always downsample each dimension with the factor 2.",
```
philippotto: `Isotropic downsampling will always...`. Otherwise, it might be a bit unclear, since the previous sentence refers to anisotropic downsampling.

tests/scripts/anisotropic_downsampling.sh
MartinBuessemeyer: True. I forgot to change that but it needs to be `8-8-2` and `16-16-4`

wkcuber/utils.py
MartinBuessemeyer: `-i` is taken by `interpolation_mode`.

wkcuber/downsampling.py
```diff
@@ -109,8 +107,15 @@ def create_parser():
         action="store_true",
     )
 
+    parser.add_argument(
```
philippotto: I think there are two other occurrences of this code. Can you extract it similar to how `add_isotropic_flag` is defined in the utils?

PR #101, opened 2019-07-19
auto-bump version number according to git tag

philippotto:
see https://github.com/scalableminds/cluster_tools/pull/52 which does the same for cluster_tools


normanrz:
The CI fails, because the docker build doesn't copy the `.git` folder, so setuptools-scm can't access the tag information.

philippotto:
The `--squash` argument for docker build is still experimental, unfortunately. `.git` is 50 mb on my local machine. I guess it's fine for now.

Issue #90, opened 2019-05-20
Allow to provide input path pattern when cubing tiled tiffs


philippotto:
Similar to how `in_plane_stitching.js` (in segmentation-tools) accepts an `input_path_pattern`.

philippotto:
@MartinBuessemeyer Please remember to close your issues when they are completed.

PR #108, opened 2019-07-30
Provide path patterns for tile cubing

MartinBuessemeyer:
This PR adds the `input_path_pattern` for tiled cubing from the `in_plane_stitching.py`. It does not require the dimensions arguments. It detects the dimensions and then orders the files. Additionally it allows to cube Images with color channels.

How to test:
- python3 -m wkcuber.tile_cubing /testdata/temca2 /testoutput/temca2_test/

Issues:
- addresses #90


philippotto:
@MartinBuessemeyer Is this ready for review? Just wondering, because I'm assigned as a reviewer, but there's still the `WIP` in the title?

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,233 @@
...
+    y_missing_number_length_offset = floor(log10(max(y_value, 1))) + 1
+    z_missing_number_length_offset = floor(log10(max(z_value, 1))) + 1
+    upper_bound_x = x_decimal_length - x_missing_number_length_offset
+    upper_bound_y = y_decimal_length - y_missing_number_length_offset
+    upper_bound_z = z_decimal_length - z_missing_number_length_offset
+
+    # try to find the file with all combinations of number lengths
+    for z_missing_number_length in range(upper_bound_z):
+        for y_missing_number_length in range(upper_bound_y):
+            for x_missing_number_length in range(upper_bound_x):
```
philippotto: So, with the default pattern of `{zzzzzzzzzz}/{yyyyyyyyyy}/{xxxxxxxxxx}`, this method will do 10^3 = 1000 iterations to check whether a certain path exists? And this will be done for files? This seems quite inefficient to me. Why is it necessary to support every combination of padding? Can't we simply assume that the user either gives a pattern like `{z}/{y}/{x}` (single digit --> no padding for all indices) or a pattern like `{zzz}/{yyy}/{xxx}` (multiple digit --> precise padding to create a length of three). Then, we would only do one file access per tile instead of 1000?

I think it's a fair assumption that the user-provided data has to be consistent (i.e., the padding is consistent or there is no padding).

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,233 @@
...
+) -> int:
+    x_min = None
+    y_min = None
+    z_min = None
+    x_max = 0
+    y_max = 0
+    z_max = 0
+    arbitrary_file = None
+    file_count = 0
+    for x in range(x_decimal_length):
```
philippotto: See my other comment regarding the padding. The three for-loops should be avoidable, right?

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,233 @@
...
+                    occurrences = re.findall("({x+}|{y+}|{z+})", file_path_pattern)
+                    index_offset_caused_by_brackets_and_specific_length = 0
+                    for occurrence in occurrences:
+                        # update the offset since the pattern and the file path have different length
+                        occurrence_begin_index = (
+                            file_path_pattern.index(occurrence)
+                            - index_offset_caused_by_brackets_and_specific_length
+                        )
+                        index_offset_caused_by_brackets_and_specific_length += 2
+                        if occurrence[1] == "x":
```
philippotto: It should be possible to unify the three branches of this if-statement, since they are almost identical (except for variable names). If you pack x_min, y_min, z_min etc. into one dict `min_values = { "x": ..., "y": .... }` you can set the appropriate values dynamically.

Overall, this method seems quite long and I didn't grasp it fully yet. Might be easier to understand after refactoring the if-statement together with my other recommendation to reduce padding-tolerance.
We can also discuss this in person, if you want.

wkcuber/tile_cubing.py
MartinBuessemeyer: Yes that's true. I could not come up with a regex that works with glob and lets you look for numbers with a dynamic length.

But since Norman wanted everything to be backwards compatible this is the solution I came up with. Please take a look at the tests. The data set does not meet the criteria, that all digits have the same length. 
Do you have a better solution that is backwards compatible?

wkcuber/tile_cubing.py
MartinBuessemeyer: Nice Idea :+1:

wkcuber/tile_cubing.py
philippotto: > But since Norman wanted everything to be backwards compatible this is the solution I came up with. Please take a look at the tests. The data set does not meet the criteria, that all digits have the same length.

Which test case are you referring to? From what I see, all the test data, uses an unpadded naming schema. This means, if an index is `5`, the file name is `5.tiff` and if the index is `511`, we know it's `511.tiff`. The provided pattern would simply be `{x}/...`. If the data is padded like `005.tiff` and `511.tiff`, the pattern should be `{xxx}`.
I think, only supporting these two options (consistent padding or consistent non-padding) is fair. I'll ask Norman about :)
Still, to not break the interface, we would need to recognize which of the two modes the user provided (padded or unpadded). I'll think about this.

> I could not come up with a regex that works with glob and lets you look for numbers with a dynamic length.

Maybe we shouldn't use glob then? We could list all files and apply the regular expression by ourselves. But let's discuss this in person, maybe.

philippotto: I talked to Norman and we agreed on going with "either unpadded or consistently padded". So, you don't need to iterate through all combinations and can simply do two file operations per look up. Alternatively, one could traverse over the provided path, list all files and match them against a regexp, but one would need to be careful to not traverse unrelated folders. To get this issue done, the first solution is more sensible :)

Can you implement this please? Will re-review again afterwards :)

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,204 @@
...
+                        )
+                        min_dimensions[current_dimension] = min(
+                            min_dimensions[current_dimension] or coordinate_value,
+                            coordinate_value,
+                        )
+                        max_dimensions[current_dimension] = max(
+                            max_dimensions[current_dimension], coordinate_value
+                        )
+
+    return (
```
philippotto: Please add an assertion that verifies that `min_dimensions` does not include None anymore.

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,204 @@
...
+                        current_dimension = occurrence[1]
+                        occurrence_end_index = (
+                            occurrence_begin_index
+                            + decimal_length[current_dimension]
+                            - current_decimal_length[current_dimension]
+                        )
+                        index_offset_caused_by_brackets_and_specific_length += current_decimal_length[
+                            current_dimension
+                        ]
+                        coordinate_value = int(
```
philippotto: The above code seems quite complicated to me. If I understand correctly, you want to extract the actual values for x, y and z and you already know how many digits each component has. Maybe the following approach is shorter and/or easier to understand?
```python
# Turn a pattern {xxx}/{yyy}/{zzzzzz} for given dimension counts into (e.g., 2, 2, 3) into
# something like xx/yy/zzz (note that the curly braces are gone)
applied_fpp = replace_coordinates(file_path_pattern, {"x": x, "y": y, "z": z})

# For each dimension, look up where it starts within the applied pattern.
# Use that index to look up the actual value within the file name
for current_dimension in ["x", "y", "z"]:
	idx = applied_fpp.index(current_dimension)
	value = int(file_name[idx:idx + decimal_length[current_dimension]])
	assert value not None
	min_dimensions[current_dimension] = min(...)
	max_dimensions[current_dimension] = min(...)
```

(I didn't test the above code, but I hope the gist is correct :))

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,204 @@
...
+    z_decimal_length: int,
+) -> Tuple[int, int, int, int, int, int, str, int]:
+    arbitrary_file = None
+    file_count = 0
+    decimal_length = {
+        "x": x_decimal_length,
+        "y": y_decimal_length,
+        "z": z_decimal_length,
+    }
+    current_decimal_length = {"x": 0, "y": 0, "z": 0}
```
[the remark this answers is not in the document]
philippotto: Ah, you use it to index into the values using the dimension as a variable. Maybe add a comment for that.

wkcuber/tile_cubing.py
```diff
@@ -66,119 +210,139 @@ def tile_cubing_job(target_wkw_info, z_batches, source_path, batch_size, tile_si
...
-                        else:
-                            buffer.append(
-                                np.zeros(tile_size, dtype=target_wkw_info.dtype)
+                for x in range(x_min, x_max + 1):
+                    for y in range(y_min, y_max + 1):
+                        ref_time2 = time.time()
+                        buffer = []
+                        for z in z_batch:
+                            # Read file if exists or use zeros instead
+                            file = find_file_with_dimensions(
```
philippotto: `file_name`

wkcuber/tile_cubing.py
```diff
@@ -1,60 +1,194 @@
...
 
-    all_source_files.sort()
-    return all_source_files
+    return decimal_lengths
+
+
+def detect_interval_for_dimensions(
+    file_path_pattern: str, decimal_lengths: Dict[str, int]
+) -> Tuple[Dict[str, int], Dict[str, int], str, int]:
+    arbitrary_file = None
```
philippotto: This variable is not used i I see it correctly? You could assign to this variable for the first value of `file_name` and return `arbitrary_file` instead of `file_name`. In my opinion, it's a bit weird to return the loop variable `file_name` outside of the loop (in other languages, this wouldn't be possible since the variable is often only defined in the loop). But that's my personal take on that. Either remove the above variable or change it as described :)

MartinBuessemeyer:
I merged the master into this branch and saw, that a new method was introduced (`prepare_slices_for_wkw`) that does that stacking of the buffer. But it does not seem compatible with my implementation. Is this fine?
To be precise, I am removing dimension with the shape of 1. It looks like that the Image-Reader always adds an extra dimension that should be used for the z dimension (see `image-readers.py `line 129). Thus the Image-Reader returns an image with the shape of `(x, y, channel_count, z=1)`. Therefore I remove that z dimension. Shall I change my script accordingly or is it fine to just leave it as it is now.

philippotto:
> I merged the master into this branch and saw, that a new method was introduced (`prepare_slices_for_wkw`) that does that stacking of the buffer. But it does not seem compatible with my implementation. Is this fine?

I think it should be possible to adapt your code so that `prepare_slices_for_wkw` can stay as it is on master.

> To be precise, I am removing dimension with the shape of 1. It looks like that the Image-Reader always adds an extra dimension that should be used for the z dimension (see image-readers.py line 129).

The linked code ensures that there is always an axis for the *channel count*. An axis for z exists, as well.

> Thus the Image-Reader returns an image with the shape of (x, y, channel_count, z=1). Therefore I remove that z dimension. Shall I change my script accordingly or is it fine to just leave it as it is now.

I'd suggest to adapt your code so that it can handle shapes with (x, y, channel_count, z). I see two options for doing this:
- remove the z-dimension in the beginning and add it back before calling `prepare_slices_for_wkw` or
- adapt the code so that all parts can deal with the 4 axes

You can decide which option you want to implement. In both cases, your code should be deal with data shapes where there is an axis for the channel count so that multi-channel data can also be "tile-cubed".

PR #109, opened 2019-08-02
ensure_wkw with compression

normanrz:
When using `downsample_mag(...)` with `compress=True`, the header isn't written with the compression flag and thus no compression happens.


normanrz:
Seems like the reference data in git is actually not compressed^^

PR #116, opened 2019-08-09
Make scale parameter required

philippotto:
...and remove scale parameter from downsampling tool.

- a default of 1,1,1 for the scale is almost never correct. the most likely scenario is that you forget to set the parameter and then your downsampling is isotropic. so, you have to fix the parameter in the dataset and downsample again
- the scale parameter for the downsampling tool shouldn't be necessary. the datasource-properties.json should always contain that scale


philippotto:
@normanrz can you force the merge? :)

PR #120, opened 2019-08-16
Map futures

### LINKED ISSUE #119: Convert jobs to array jobs
It seems as if slurm is more robust with array jobs than with a ton of single jobs. Also, the array jobs are easier to control and kill. Would be good to switch to those.

normanrz:
Using `executor.map_to_futures` instead of `executor.submit` everywhere.
Fixes #119


tests/test_export_wkw_as_tiff.py
```diff
@@ -9,118 +9,195 @@
...
         slice_bbox["size"] = [slice_bbox["size"][0], slice_bbox["size"][1], 1]
         for data_slice_index in range(bbox["size"][2]):
 
             for y_tile_index in range(ceil(bbox["size"][1] / tile_bbox["size"][1])):
-                for x_tile_index in range(ceil(tile_bbox["size"][0] / tile_bbox["size"][0])):
-                    tiff_path = os.path.join(destination_path, f"{data_slice_index + 1}", f"{y_tile_index + 1}", f"{x_tile_index + 1}.tiff")
-
-                    assert os.path.isfile(tiff_path), f"Expected a tiff to be written at: {tiff_path}."
+                for x_tile_index in range(
+                    ceil(tile_bbox["size"][0] / tile_bbox["size"][0])
```
philippotto: Do you know where the format changes come from? Did the black version change?

normanrz: I'm not sure the tests had been formatted by black at all.

PR #126, opened 2019-08-27
fixed float/double to float32, float64 conversion

### LINKED ISSUE #100: float/double instead of float32/float64
webknossos expects the datatypes to be called `float` and `double` instead of `float32` or `float64` in the `datasource-properties.json`. This needs to be converted both ways in `wkcuber.metadata`

MartinBuessemeyer:
This PR fixes the bug described in #100. Additionally I fixed two more bugs. An argument was missing in the tests. `read_read_metadata_for_layer` could not read a layer with the `elementClass` of `uint24` which can be produced by writing a layer with dtype `uint8` and 3 channels. I fixed it by returning `uint8` when there is no `numpy-dtype` that matches `elementClass`. 

**Steps to test:**
       run the tests

**Issues**:
      fixes #100

### DIFF
--- a/tests/scripts/simple_anisotropic_tiff_cubing.sh
+++ b/tests/scripts/simple_anisotropic_tiff_cubing.sh
@@ -10,6 +10,7 @@ docker run \
   --batch_size 8 \
   --layer_name color \
   --max 8 \
+  --scale "11.24,11.24,25"\
   --name awesome_data \
   /testdata/tiff /testoutput/tiff3
 [ -d testoutput/tiff3/color ]
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -17,9 +17,9 @@ def test_element_class_convertion():
     prediction_wkw_info = WkwDatasetInfo(
         test_wkw_path, prediction_layer_name, np.float32, 1
     )
-    ensure_wkw(prediction_wkw_info, num_channels=1)
+    ensure_wkw(prediction_wkw_info, num_channels=3)
 
-    write_costum_layer(test_wkw_path, "prediction", np.float32, num_channels=1)
+    write_custom_layer(test_wkw_path, "prediction", np.float32, num_channels=3)
     write_webknossos_metadata(
         test_wkw_path,
         "test_metadata",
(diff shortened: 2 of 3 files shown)


tests/test_metadata.py
```diff
@@ -0,0 +1,70 @@
...
+    read_datasource_properties,
+    read_metadata_for_layer,
+)
+
+
+def test_element_class_convertion():
+    test_wkw_path = os.path.join("testoutput", "test_metadata")
+    prediction_layer_name = "prediction"
+    prediction_wkw_info = WkwDatasetInfo(
+        test_wkw_path, prediction_layer_name, np.float32, 1
```
philippotto: Prediction layers usually have three channels which you could reflect here.

tests/scripts/simple_anisotropic_tiff_cubing.sh
```diff
@@ -10,6 +10,7 @@ docker run \
   --batch_size 8 \
   --layer_name color \
   --max 8 \
+  --scale "11.24,11.24,25"\
```
philippotto: Do you know why the CI passed even before this change? Isn't this called in the CI?

wkcuber/metadata.py
```diff
@@ -149,6 +161,14 @@ def read_metadata_for_layer(wkw_path, layer_name):
     return layer_info, dtype, bounding_box, origin
 
 
+def convert_dype_to_element_class(dtype):
+    convertion_map = {np.float32: "float", np.float64: "double"}
```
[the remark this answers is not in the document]
MartinBuessemeyer: That does not work any more after merging `known_nupy_types`.

tests/scripts/simple_anisotropic_tiff_cubing.sh
MartinBuessemeyer: Sry I don't know.

tests/test_metadata.py
MartinBuessemeyer: The `1` is the `Mag` and the `WKWDatasetInfo` does not contain a `num_channel`-variable.

Issue #97, opened 2019-06-18
Downsample image stacks without writing mag1


normanrz:
When cubing image stacks, it would be great to be able to create downsampled versions (eg 2-2-1) without writing mag 1 to disk.

MartinBuessemeyer:
I have a Question towards this issue:
What is exactly meant with "downsampled versions"?
1. Way of understanding: The Image-Stack has already the format Mag 2-2-1 and therefore it should be written to Mag 2-2-1 instead of Mag 1. Therefore this would (more or less) just require another command line argument.
2. Way of understanding: The Image-Stack has the format Mag 1 and therefore should be directly downsampled to Mag 2-2-1 without Mag 1 being saved to the disk. Then only Mag 2-2-1 should be saved to the disk. Therefore this would require another command line argument and direct "in memory" downsampling.

normanrz:
I meant 2

PR #132, opened 2019-09-19
Downsample tiff cubing in memory 

MartinBuessemeyer:
Adds the option of downsampling a cubed image with `cubing.py` to directly in memory to another `Magnification`. It can be activated by providing the `--traget_mag` argument and the interpolation mode can be set with `--interpolation_mode`.

**Steps to test:**
- `in_memory_downsampled_cubing.sh` tests the new feature and is already included in the test execution.

**Addesses:**
- #97


tests/scripts/in_memory_downsampled_cubing.sh
```diff
@@ -0,0 +1,17 @@
...
+  -v "${PWD}/testoutput:/testoutput" \
+  --rm \
+  scalableminds/webknossos-cuber:${CIRCLE_BUILD_NUM} \
+  wkcuber.cubing \
+  --jobs 2 \
+  --batch_size 8 \
+  --layer_name color \
+  --target_mag 2-2-1 \
+  --interpolation_mode default \
+  /testdata/tiff /testoutput/tiff4
```
philippotto: Instead of `tiff4` I'd use something like `in_memory_downsampled_tiff`.

wkcuber/cubing.py
```diff
@@ -57,6 +60,13 @@ def create_parser():
         action="store_true",
     )
 
+    parser.add_argument(
+        "--target_mag",
+        help="Automatically downsamples the cubed images to the provided mag before writing to disk. The magnification can be provided like 2-2-1.",
```
philippotto:
```suggestion
        help="Automatically downsamples the cubed images to the provided magnification before writing to disk. The magnification can be provided like 2-2-1.",
```

wkcuber/cubing.py
```diff
@@ -175,7 +229,16 @@ def cubing_job(args):
 
 def cubing(source_path, target_path, layer_name, dtype, batch_size, args=None) -> dict:
 
-    target_wkw_info = WkwDatasetInfo(target_path, layer_name, dtype, 1)
+    target_mag = Mag(getattr(args, "target_mag", 1))
```
philippotto: Why not `target_mag = Mag(args.target_mag)` and then defaulting to target_mag to 1 in the argparse definition?

wkcuber/cubing.py
```diff
@@ -175,7 +229,16 @@ def cubing_job(args):
...
+    target_wkw_info = WkwDatasetInfo(target_path, layer_name, dtype, target_mag)
+    if target_mag != Mag(1):
+        interpolation_mode = parse_interpolation_mode(
+            getattr(args, "interpolation_mode", "DEFAULT"), target_wkw_info.layer_name
```
philippotto: same here

wkcuber/cubing.py
```diff
@@ -156,8 +170,48 @@ def cubing_job(args):
                     ]
 
                 buffer = prepare_slices_for_wkw(slices, num_channels)
+                if downsampling_needed:
+                    logging.info(f"Downsampling buffer of size {buffer.shape}.")
```
philippotto: I'd extract lines 174 to 211 into a separate function (e.g., `downsample_unpadded_data`).

wkcuber/cubing.py
```diff
@@ -156,8 +170,48 @@ def cubing_job(args):
...
+                    padding_size_for_downsampling = [
+                        (
+                            0,
+                            (
+                                required_multiple_of
+                                - (buffer_dimension_size % required_multiple_of)
+                            )
+                            % required_multiple_of,
```
philippotto: Why is this modulo operation necessary? `buffer_dimension_size % required_multiple_of` evaluates to a value with `value <= required_multiple_of`. That's why `required_multiple_of - (buffer_dimension_size % required_multiple_of)` has to be `< requried_multiple_of` which is why `% required_multiple_of` won't make any difference. Or am I missing something?

philippotto: `padding_size_for_downsampling` can be calculated more concisely by leveraging vectorization with numpy. Something along these lines:

```python
target_mag_np = np.array(target_mag.to_array())
buffer_shape_np = np.array(buffer.shape[1:])
padding_size_for_downsampling = target_mag_np - (buffer_shape_np % target_mag_np)
```

wkcuber/cubing.py
```diff
@@ -156,8 +170,48 @@ def cubing_job(args):
...
+                        )
+                    ]
+                    logging.info([(0, 0)] + padding_size_for_downsampling)
+                    buffer = np.pad(
+                        buffer,
+                        pad_width=[(0, 0)] + padding_size_for_downsampling,
+                        mode="constant",
+                    )
+                    downsampled_buffer_shape = [
+                        current_dimension_size // dimension_decrease
```
philippotto: Similarly, if both these variables are numpy arrays, `downsampled_buffer_shape = current_dimension_size // dimension_decrease` is enough for this expression. You might need to convert back from np to python lists.

wkcuber/cubing.py
MartinBuessemeyer: Thanks for the numpy approach. It makes it way easier to read. :+1:  But I think that the second % is important. Imagine `buffer_shape=10` and `target_mag=2` then `buffer_shape % target_mag = 0` and therefore the `padding_size=2` even if no padding is needed. Therefore the second % is used to catch the case that there is no padding needed.

wkcuber/cubing.py
MartinBuessemeyer: Very nice :wink:

wkcuber/cubing.py
philippotto: Ah, now, I get it. When the first mod operation yields 0, the subtraction yields `padding_size` which has to be turned to zero. Good point :+1: However, I think, the overall operation can still be expressed more concise with: `-buffer_shape_np % required_multiple_of` :)

wkcuber/cubing.py
```diff
@@ -173,9 +193,41 @@ def cubing_job(args):
                 raise exc
 
 
+def downsample_unpadded_data(buffer, target_mag, interpolation_mode):
```
philippotto: Nice :+1: I'd suggest to move this inside the downsampling.py module. Afterwards, feel free to merge :)

PR #139, opened 2019-10-01
Input nifti

### LINKED ISSUE #63: Support NifTi // NII Files
Cuber should be able to convert NiFTi files (*.nii). Typically used in MRT

leowe:
Closes #63


wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+        "Converting of {} took {:.8f}s".format(
+            source_nifti_path, time.time() - ref_time
+        )
+    )
+
+    write_webknossos_metadata(
+        target_path,
+        source_nifti_path.split("/")[-2],
+        scale="10,10,10",
+        exact_bounding_box={"topleft": offset, "size": size},
```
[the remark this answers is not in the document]
hotzenklotz: I made the appropriate changes in commit a9d796a

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+        "Converting of {} took {:.8f}s".format(
+            source_nifti_path, time.time() - ref_time
+        )
+    )
+
+    write_webknossos_metadata(
+        target_path,
+        source_nifti_path.split("/")[-2],
+        scale="10,10,10",
+        exact_bounding_box={"topLeft": offset, "width": size[0], "height": size[1], "depth": size[2]},
```
hotzenklotz: @leowe Do you know the shape of the nifit files? Is it width first, then height etc? Or the other way around?

hotzenklotz:
@leowe  Please finish this PR and add some documentation.

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+
+def convert_folder_nifti(
+    source_folder_path, target_path, color_file, segmentation_file
+):
+    files = [f for f in os.listdir(source_folder_path) if f.endswith(".nii")]
```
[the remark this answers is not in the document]
hotzenklotz: Ne, ich meine: https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob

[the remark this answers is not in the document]
normanrz: Is there a difference to using `files = glob.glob("**/*.nii")`?

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+
+    return (data / factor).astype(np.dtype(target_dtype))
+
+
+def convert_nifti(source_nifti_path, target_path, layer_name, dtype, mag=1):
+    target_wkw_info = WkwDatasetInfo(target_path, layer_name, dtype, mag)
+    ensure_wkw(target_wkw_info)
+
+    ref_time = time.time()
+    # ignoring affine transformation
```
philippotto: Can you elaborate on this comment? What transformation? Do you mean a potential translation which is ignored by setting the offset to `(0, 0, 0)`? In that case, you can remove the comment, I think. Or change it to `Assume no translation` to be more precise.

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+        source_nifti = nib.load(source_nifti_path)
+        cube_data = np.array(source_nifti.get_fdata())
+
+        if len(source_nifti.shape) == 3:
+            size = list(source_nifti.shape)
+            cube_data = cube_data.reshape((1,) + source_nifti.shape)
+
+        elif len(source_nifti.shape) == 4:
+            size = list(source_nifti.shape[:-1])
+            cube_data = np.transpose(cube_data, (3, 0, 1, 2))
```
philippotto: Why do you assume `3` as the channel count? Shouldn't `source_nifti.shape[-1]` include the channel count?

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+    logging.debug(
+        "Converting of {} took {:.8f}s".format(
+            source_nifti_path, time.time() - ref_time
+        )
+    )
+
+    write_webknossos_metadata(
+        target_path,
+        source_nifti_path.split("/")[-2],
+        scale="10,10,10",
```
philippotto: The scale should be a required (without a default) argparse parameter to ensure that the written metadata is correct.

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+
+    parser.add_argument(
+        "--dtype",
+        "-d",
+        help="Target datatype (e.g. uint8, uint16, uint32).",
+        default="uint8",
+    )
+
+    parser.add_argument(
+        "--convert_folder",
```
philippotto: I'd remove this argument and change the description of `source_path` to `Path to NIFTY file or to a directory if multiple NIFTI files should be converted. In the latter case, also see --color_file and --segmentation_file.`

In the code, you simply check whether the path points to a file or a dir.

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+    )
+
+    parser.add_argument(
+        "--segmentation_file",
+        help="When converting folder, name of file to become segmentation layer",
+        default=None,
+    )
+
+    add_verbose_flag(parser)
+    add_distribution_flags(parser)
```
philippotto: This line can be removed, since your code doesn't make use of these distribution flags. I guess, it's also not necessary, since nifti files are typically rather small, right?

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,177 @@
...
+def to_target_datatype(data: np.ndarray, target_dtype) -> np.ndarray:
+
+    if data.dtype == np.dtype("float32") or data.dtype == np.dtype("float64"):
+        factor = data.max()
+
+    else:
+        factor = np.iinfo(data.dtype).max
+
+    if target_dtype != np.dtype("float32"):
+        factor = factor / np.iinfo(target_dtype).max
```
philippotto: I find this logic a bit hard to grasp, since the calculation is spread over to ifs. How about merging them like this:

```python
    if data.dtype == np.dtype("float32"):
        factor = data.max()
    elif data.dtype == np.dtype("float64"):
        factor = data.max() / np.iinfo(target_dtype).max
    else:
        factor = np.iinfo(data.dtype).max / np.iinfo(target_dtype).max
```

However, I'm not sure, whether this makes 100% sense. If the input is in float32 (and the output too), the output will be normalized between -1 and 1, right? However, there shouldn't be any change happening to the data from my understanding?

wkcuber/convert_nifti.py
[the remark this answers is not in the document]
philippotto: > Is there a difference to using `files = glob.glob("**/*.nii")`?

Seems to be equivalent, but more object oriented (since it can be called on a path object). The table at the end of @hotzenklotz's link shows a good comparison between the two interfaces (https://docs.python.org/3/library/pathlib.html#correspondence-to-tools-in-the-os-module).

wkcuber/convert_nifti.py
leowe: Nifti saves the transformation matrix in the affine attribute.

wkcuber/convert_nifti.py
leowe: You're right. I think the changes should do it.

wkcuber/convert_nifti.py
leowe: It should be width and then height. So it should be right.

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,179 @@
...
+    )
+
+    parser.add_argument(
+        "--segmentation_file",
+        help="When converting folder, name of file to become segmentation layer",
+        default=None,
+    )
+
+    parser.add_argument(
+        "--scale", help="The scale of the NIFTI input in the form X,Y,Z.", default=None,
```
hotzenklotz: I am pretty sure `scale` is read as a string from the command-line, e.g. `"10, 10, 30"`. Instead it should be a list, e.g. `[10, 10, 30]`.

hotzenklotz: Please add some type assertion to `write_webknossos_metadata` to prevent this in the future. 

See `(from .metadata import write_webknossos_metadata)`

normanrz: Not sure, if I asked before: Isn't the scale stored in the nifti header?

philippotto: The `main` and the `metadata` entry point also use a scale parameter which needs to be passed in the form of `10,10,10` (without any brackets). Therefore, convert_nifti should do the same.
However, you can import `add_scale_flag` from `utils.py` for that. Converting the scale from string to a list is done via `scale = tuple(float(x) for x in args.scale.split(","))` (see main method in `__main__`). It would be cool if you could make the scale-parsing part of the argparse definition. Something like this would do the trick:

```
def parse_scale(scale):
  try:
    scale = tuple(float(x) for x in args.scale.split(","))
    return scale
  except Exception:
        raise argparse.ArgumentTypeError("The scale could not be parsed")

def add_scale_flag(parser, required=True):
    parser.add_argument(
        "--scale",
        "-s",
        help="Scale of the dataset (e.g. 11.2,11.2,25). This is the size of one voxel in nm.",
        required=required,
        type=parse_scale,
    )
```

Then, `args.scale` doesn't need to be parsed for each usage of `add_scale_flag`.

hotzenklotz: @leowe BTW, I already fixed the `datasource-properties.json` files for stroke-vision

leowe: Okay, so there is a parameter in the nifty header that can store the scale of the dataset (`pixdim`). Should I make the scale flag optional and try to infer the scale from the nifty header and when the parameter is given simply overwrite the inferred scale?

philippotto: Sounds reasonable!

leowe: I now implemented it as discussed.

PR #140, opened 2019-10-02
add recubing script

leowe:
* add recubing script which writes an existing dataset in a new one where you can specify the wkw_file_len (which is e.g. helpful when it was written with a file length of 1 and can therefore not be downsampled)


wkcuber/recubing.py
```diff
@@ -0,0 +1,138 @@
...
+                        outer_bounding_box_tl[1] + y,
+                        outer_bounding_box_tl[2] + z,
+                    ]
+
+                    logging.info("Writing at {}".format(top_left))
+
+                    data_cube = source_wkw_dataset.read(
+                        top_left, (wkw_cube_size, wkw_cube_size, wkw_cube_size)
+                    )
+                    target_wkw_dataset = open_wkw(target_wkw_info)
```
philippotto: This `open_wkw` call should be below line 106 in a `with` context (similar to source_wkw_dataset). Otherwise, you open the dataset for each cube (and it's not closed).

wkcuber/recubing.py
```diff
@@ -0,0 +1,138 @@
...
+    source_wkw_info = WkwDatasetInfo(source_path, layer_name, 1, source_wkw_header)
+
+    ensure_wkw(target_wkw_info)
+
+    bounding_box = infer_bounding_box(source_wkw_info)
+    bottom_right = [
+        coord + size for coord, size in zip(bounding_box[0], bounding_box[1])
+    ]
+
+    wkw_cube_size = wkw_file_len * BLOCK_LEN
```
philippotto: Instead, of `BLOCK_LEN` is better to use `target_wkw_header.block_len` since it's the proper value (and not an imported, static number).

wkcuber/recubing.py
```diff
@@ -0,0 +1,138 @@
...
+        outer_bounding_box_br[0] - outer_bounding_box_tl[0],
+        outer_bounding_box_br[1] - outer_bounding_box_tl[1],
+        outer_bounding_box_br[2] - outer_bounding_box_tl[2],
+    ]
+
+    with open_wkw(source_wkw_info) as source_wkw_dataset:
+        for x in range(0, outer_bounding_box_size[0], wkw_cube_size):
+            for y in range(0, outer_bounding_box_size[1], wkw_cube_size):
+                for z in range(0, outer_bounding_box_size[2], wkw_cube_size):
+                    logging.info("Reading at offset {}, {}, {}".format(x, y, z))
```
philippotto: That logging statement can be removed.

wkcuber/utils.py
```diff
@@ -47,6 +47,33 @@ def ensure_wkw(target_wkw_info):
     target_wkw.close()
 
 
+def infer_bounding_box(dataset: WkwDatasetInfo):
```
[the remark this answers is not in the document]
leowe: Yes I can do that. `detect_bbox` defines parse_cube_file_name again. Should I simply import it there from `utils`?

wkcuber/convert_nifti.py
```diff
@@ -0,0 +1,110 @@
+import time
```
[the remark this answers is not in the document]
leowe: You're right.

fredynaruton (approved):

Issue #133, opened 2019-09-23
Downsampling of datasets with non-default file_len is broken


philippotto:
When using a `file_len != 32`, the downsampling code breaks at at least one spot:

```
                assert (
                    wkw_cubelength % cube_edge_len == 0
                ), "buffer_cube_size must be a divisor of wkw cube length"
```

For example, a `file_len` of `1` will make the cube_length `32`, but the default `cube_edge_len` is 256. If `cube_edge_len` is not specified, it should default to a value which always "just works".

If one passes `cube_edge_len == 32` (and uses live compression which is turned on by default), the downsampling fails, too, since it will try to read only one cube at a time and write that into the new magnification. However, for live compression, cubes have to be written as a whole. The proper fix for this is probably to re-use the same `file_len` attribute for the new magnification.

PR #141, opened 2019-10-04
Fix downsampling bug if non-default file_len is used

rschwanhold:
fixes: #133

### DIFF
--- a/wkcuber/downsampling.py
+++ b/wkcuber/downsampling.py
@@ -168,7 +168,10 @@ def downsample(
             wkw.Header.BLOCK_TYPE_LZ4HC if compress else wkw.Header.BLOCK_TYPE_RAW
         )
         ensure_wkw(
-            target_wkw_info, block_type=header_block_type, num_channels=num_channels, file_len=source_wkw.header.file_len
+            target_wkw_info,
+            block_type=header_block_type,
+            num_channels=num_channels,
+            file_len=source_wkw.header.file_len,
         )
 
     with get_executor_for_args(args) as executor:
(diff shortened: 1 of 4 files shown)


wkcuber/downsampling.py
```diff
@@ -148,15 +148,15 @@ def downsample(
     logging.debug(
         "Found source cubes: count={} size={} min={} max={}".format(
             len(source_cube_addresses),
-            (cube_edge_len,) * 3,
+            (buffer_edge_len,) * 3,
```
rschwanhold: If no explicit `buffer_edge_len` is specified, the value is `None` at this point. This leads to debug messages like: `DEBUG Found source cubes: count=1 size=(None, None, None) min=(0, 0, 0) max=(0, 0, 0)`. 
How should we handle this?

wkcuber/downsampling.py
```diff
@@ -168,7 +168,10 @@ def downsample(
             wkw.Header.BLOCK_TYPE_LZ4HC if compress else wkw.Header.BLOCK_TYPE_RAW
         )
         ensure_wkw(
-            target_wkw_info, block_type=header_block_type, num_channels=num_channels
+            target_wkw_info,
```
philippotto: Nice stuff! I created the follow up issue we talked about: https://github.com/scalableminds/webknossos-cuber/issues/142

wkcuber/downsampling.py
philippotto: I suggest to extract https://github.com/scalableminds/webknossos-cuber/pull/141/files#diff-363a076e73484d6f38615890702e0506R224-R226 into a `determine_buffer_edge_len` method and then open the source dataset before logging these values. Then, you can read the file_len and print the buffer_edge_len afterwards properly.

philippotto:
@normanrz the license/cla check doesn't finish. Can you have a look?

Issue #137, opened 2019-09-27
Allow providing a specific target mag for downsampling anisotropically


philippotto:
Right now, the downsampling interface allows the following:

- create one new magnification with a certain mag (for example, create 2-2-1 or 2-2-2 from 1-1-1)
- downsample isotropically to a certain target mag (for example create 2, 4, 8, 16 from 1)
- downsample anisotropically "automatically" (all magnifications are determined automatically)

However, for alignment we would like to have the following mode:
- create downsampled magnifications until a certain target magnification (for example, the target magnification could be 16-16-1 and then all mags between should be created without considering the scale).

PR #143, opened 2019-10-08
Allow providing a specific target mag for downsampling anisotropically

rschwanhold:
fixes: #137

### DIFF
--- a/wkcuber/downsampling.py
+++ b/wkcuber/downsampling.py
@@ -90,8 +90,8 @@ def create_parser():
 
     group.add_argument(
         "--anisotropic_target_mag",
-        help="Specify an explicit anisotropic target magnification which should be "
-        "created (e.g., --anisotropic_target_mag 2-2-1). Consider using --anisotropic "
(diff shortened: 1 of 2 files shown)


rschwanhold:
Should I implement tests for this?

wkcuber/downsampling.py
```diff
@@ -38,6 +38,12 @@ def determine_buffer_edge_len(dataset):
     return min(DEFAULT_EDGE_LEN, dataset.header.file_len * dataset.header.block_len)
 
 
+def calculate_scale(target_mag):
```
philippotto: Very elegant solution :+1: It didn't cross my mind that you could simply reuse `downsample_mags_anisotropic` by calculating an appropriate "fake scale".

However, I'd rename `calculate_scale` to `calculate_virtual_scale_for_target_mag` or something like that. Also you could add a comment that explains that this scale only serves the purpose of being used for `downsample_mags_anisotropic`. Otherwise, it could be confusing why the actual dataset scale is not used.

rschwanhold:
I've looked at a downsampled dataset in webknossos and it looks good.

Issue #142, opened 2019-10-04
Refactor WkwDatasetInfo to include all wkw-header-related properties


philippotto:
Right now, WkwDatasetInfo is used to pass dataset descriptors around, but doesn't include all wkw-header-specific values (it can only hold `voxel_type` currently). That's why, calls to 
 `open_wkw` and `ensure_wkw` currently consume the rest of the header-related variables. Ideally, the info tuple should be able to describe everything which is needed to create a wkw dataset. Thus, also the header. It could look like this:

```
WkwDatasetInfo = namedtuple(
    "WkwDatasetInfo", ("dataset_path", "layer_name", "mag", "header")
)
```

where header may contain voxel_type, num_channels, version, block_len, file_len and block_type.

PR #144, opened 2019-10-10
Refactor WkwDatasetInfo to include all wkw-header-related properties

rschwanhold:
fixes: #142

### DIFF
--- a/wkcuber/cubing.py
+++ b/wkcuber/cubing.py
@@ -1,6 +1,7 @@
 import time
 import logging
 import numpy as np
+import wkw
 from argparse import ArgumentParser
 from os import path
 from natsort import natsorted
@@ -121,7 +122,6 @@ def cubing_job(args):
         source_file_batches,
         batch_size,
         image_size,
-        num_channels,
         pad,
     ) = args
     if len(z_batches) == 0:
@@ -129,7 +129,7 @@ def cubing_job(args):
 
     downsampling_needed = target_mag != Mag(1)
 
-    with open_wkw(target_wkw_info, num_channels=num_channels) as target_wkw:
+    with open_wkw(target_wkw_info) as target_wkw:
         # Iterate over batches of continuous z sections
         # The batches have a maximum size of `batch_size`
         # Batched iterations allows to utilize IO more efficiently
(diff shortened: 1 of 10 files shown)


wkcuber/downsampling.py
```diff
@@ -462,10 +474,15 @@ def downsample_mag(
 ):
     interpolation_mode = parse_interpolation_mode(interpolation_mode, layer_name)
 
-    source_wkw_info = WkwDatasetInfo(path, layer_name, None, source_mag.to_layer_name())
+    source_wkw_info = WkwDatasetInfo(
+        path, layer_name, source_mag.to_layer_name(), wkw.Header(np.uint8)
+    )
```
rschwanhold: Previously the `voxel_type` was `None`. I am not sure why this should ever be `None` so I changed it to `uint8`. Is that correct?

wkcuber/utils.py
```diff
@@ -33,21 +33,16 @@
 logger = getLogger(__name__)
 
 
-def open_wkw(info, **kwargs):
-    if hasattr(info, "dtype"):
-        header = wkw.Header(np.dtype(info.dtype), **kwargs)
-    else:
-        header = wkw.Header(np.uint8, **kwargs)
```
rschwanhold: Is it correct that this `if` is no longer needed, because the header always has a `voxel_type` attribute?

wkcuber/cubing.py
```diff
@@ -195,8 +197,17 @@ def cubing_job(args):
 
 def cubing(source_path, target_path, layer_name, dtype, batch_size, args=None) -> dict:
```
rschwanhold: The `dtype` attribute is the argument that the user can pass per command line. This is for example `"uint8"`. How to convert this string into an `VALID_VOXEL_TYPE`?

wkcuber/utils.py
philippotto: Yes, this should be correct!

wkcuber/downsampling.py
philippotto: Hmm, semantically, it's not 100% correct. The thing is, that we don't know which voxel_type is actually used in the `source_wkw` info before opening the dataset. That's why, it wasn't specified before. Using `uint8` is merely a guess, which will work, since wkw will ignore it, but it's not really correct.
My suggestion would be as follows: Instead of providing a `wkw.Header` to `WkwDatasetInfo` simply pass `None`. In `ensure_wkw` you can assert that the header must not be `None` (since it's needed for the creation of a dataset).

So, the rule of thumb would be to only pass a header when the dataset might need to be created. If it's already there, don't pass a header. Does this make sense?

wkcuber/cubing.py
philippotto: There is a `convert_element_class_to_dtype` function in `metadata.py` which should be what you are looking for. Do you know why this wasn't a problem before this PR?

PR #151, opened 2019-11-07
Detect layers by looking at header.wkw

### LINKED ISSUE #148: Detect layers by looking for header.wkw
When a layer is not named `color`, `segmentation` or `prediction` it is recognized as a wkw layer. However, it should be possible to detect a layer simply by looking for a `header.wkw` file. Then we can detect arbitrary layer names.

rschwanhold:
closes #148

### DIFF
--- a/wkcuber/metadata.py
+++ b/wkcuber/metadata.py
@@ -12,6 +12,8 @@ from typing import Optional
 from .mag import Mag
 from typing import List
 from .utils import add_verbose_flag, setup_logging, add_scale_flag
+from pathlib import Path
+from os.path import *
 
 
 def get_datasource_path(dataset_path):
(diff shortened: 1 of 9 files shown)


wkcuber/metadata.py
```diff
@@ -12,6 +12,8 @@
 from .mag import Mag
 from typing import List
 from .utils import add_verbose_flag, setup_logging, add_scale_flag
+from pathlib import Path
+from os.path import *
```
philippotto: I consider `import *` as an anti-pattern as it hides the fact where functions are taken from. For example, the code itself doesn't really tell me that `glob` stems from `os.path`. Additionally, there is the risk that this wildcard import shadows other variables in this namespace. Therefore, I'd advise to state explictly what you want to import. Alternatively, you can just do `import os.path` and then use glob like `os.path.glob`, I think.

wkcuber/metadata.py
```diff
@@ -309,13 +311,22 @@ def detect_segmentation_layer(
...
             dataset_path, "segmentation", max_id, compute_max_id, exact_bounding_box
         )
+    available_layer_names = set(
+        [
+            basename(normpath(Path(x).parent.parent))
+            for x in glob.glob(dataset_path + "/*/*/header.wkw")
+        ]
+    )
+    for layer_name in available_layer_names:
+        # color and segmentation are already checked explicitly to ensure downwards compatibility
```
philippotto:
```suggestion
        # color and segmentation are already checked explicitly to ensure downwards compatibility (some older datasets don't have the header.wkw file)
```

wkcuber/metadata.py
rschwanhold: I see what you mean. Actually glob is not from `os.path` (it was already imported).

philippotto: > Actually glob is not from `os.path` (it was already imported).

that drives the point home, too :laughing:

hotzenklotz: #advertisment Me likey `pathlib.rglob` https://docs.python.org/3/library/pathlib.html#pathlib.Path.rglob

PR #163, opened 2019-11-27
fixes layer detections

### LINKED ISSUE #162: cubeLength cannot be `null`
The metadata detection may yield `null` for `cubeLength`. This leads webKnossos to not load the dataset at all. I would suggest to drop the layer-mag entry in the json for when the cubeLength cannot be detected.

See https://github.com/scalableminds/webknossos-cuber/blob/master/wkcuber/metadata.py#L193-L197

normanrz:
Fixes #162


normanrz:
How about this?
- If a standard layer folder exists, but is mal-formed: assertion
- If there are extra folders, but are mal-formed: warning

philippotto:
Yes, this sounds good :+1: Then there needs to be an try-except block in https://github.com/scalableminds/webknossos-cuber/pull/163/commits/5075daee95166249943159e7fc0ce602f732e470#diff-dfa45265a3a1fa3577bd21df42ba981aR351, right?

Issue #99, opened 2019-07-03
High-level dataset library


normanrz:
I would like to have a more high-level library for webknossos datasets. The library should have the following features

- [ ] Provide a handle for datasets (with multiple layers and mags)
- [ ] Property access to the datasets metadata
- [ ] Read/write data from specified layers and mags
- [ ] Provide a handle to a specified bounding box in the data
- [ ] The data handle should have methods for aligned chunking

As an example, this is code I'd like to write:
```python
from wkcuber import WKDataset

def work_on_chunk(slice):
  print("offset={} shape={}".format(slice.offset, slice.shape))

# Create handle
with WKDataset.open("/path/to/dataset") as ds:
  # Metadata access
  print("Color layer dtype={} bounding_box={}".format(ds["color"].dtype, ds["color"].bounding_box))
  
  # Write data
  ds["color"][Mag(2,2,1)][10:20, 10:20, 10:20] = np.zeros((10, 10, 10))

  # Create a data handle
  slice = ds["color"][Mag(2,2,1)].slice(offset=(10, 10, 10), size=(1024, 1024, 1024))

  # Read data from the data handle
  slice[0:10, 0:10, 0:10]

  # Chunking in the data handle
  with SlurmExecutor() as exec:
    slice.map_chunked(work_on_chunk, chunk_alignment=(32, 32, 32), executor=exec)
```

For this, I think we can re-use some code from the segmentation-tools. Notably the `BoundingBox` class and the `distribute.py`

Further extensions would be to also allow the manipulative steps from the wkcuber to this api. For example:
```python
ds["color"][Mag(1)].downsample(Mag(2, 2, 1), executor=exec)
ds["color"][Mag(2, 2, 1)].compress(executor=exec)
```

philippotto:
Maybe this is a good issue for @rschwanhold? I think we should discuss/think through the new interface thoroughly before implementing it, since it will probably stick with us for quite some time.

normanrz:
Makes sense. How would we spec that? I think it makes sense to have it in written form.

philippotto:
> I think it makes sense to have it in written form.

Definitely. I think, we should write down all user-facing classes + method signatures.

philippotto:
@rschwanhold Talk to me when you got capacities for this issue :)

Another interesting aspect for this is issue are tiff datasets for which we could define a dedicated format.

PR #165, opened 2019-11-28
High-level dataset library

rschwanhold:
closes: #99 
- [x] implement a way to create/open datasets
- [x] implement a way to add layers to datasets
- [x] implement a way to add mags to layers
- [x] update the `datasource-properties.json` when layers/mags are added/removed
- [x] Read/write data from specified layers and mags
- [x] Provide a handle to a specified bounding box in the data
- [x] update `datasource-properties.json` when data is written and therefore the boundingbox changes
- [x] implement a library for accessing tiff data
- [x] clean up
- [x] implement tests


rschwanhold:
@philippotto, @normanrz I've pushed the current state of my code. There are still some things that I need to work on, but you might already take a look at it.

I realized that the directory of `testdata/simple_wk_dataset` is really big, therefore I didn't commit it. The other tests should work.

tests/test_dataset.py
```diff
@@ -0,0 +1,296 @@
...
+
+    copytree(old_dataset_path, new_dataset_path)
+
+    slice = WKSlice(mag_filename, size=(100, 100, 100))
+
+    with slice.open():
+        data = slice.read((10, 10, 10))
+        assert not np.array_equal(data, np.zeros((3, 10, 10, 10)))
+
+        slice.write(np.zeros((3, 10, 10, 10), dtype=np.uint8))
```
philippotto: I'd rather generate some random numbers, write these and then compare these against the generated numbers. You can use numpy for that. I also recommend to initialize the seed of numpy random with a static value so that the test is deterministic (e.g., `np.random.seed(1234)`).

tests/test_dataset.py
```diff
@@ -0,0 +1,296 @@
...
+    new_dataset_path = path.join(dirname, "../testoutput/simple_tiff_dataset/")
+    mag_filename = path.join(new_dataset_path, "color/1")
+
+    delete_dir(new_dataset_path)
+
+    copytree(old_dataset_path, new_dataset_path)
+
+    slice = TiffSlice(mag_filename, size=(100, 100, 100))
+
+    with slice.open():  # no need to pass a header because the default fit for this test
```
philippotto: So, the default is uint8, right? If I'd try to write uint16 data, would it throw an exception?

tests/test_dataset.py
```diff
@@ -0,0 +1,296 @@
...
+    copytree(old_dataset_path, new_dataset_path)
+
+    slice = TiffSlice(mag_filename, size=(100, 100, 100))
+
+    with slice.open():
+        try:
+            slice.write(
+                np.zeros((200, 200, 5), dtype=np.uint8)
+            )  # this is bigger than the bounding_box
+            raise AssertionError(
```
philippotto: Awesome :+1:

philippotto: There is no test pendant for writing outside of bounds for a WKSlice, is there?

tests/test_dataset.py
```diff
@@ -0,0 +1,296 @@
...
+    # 10 images (z-layers), each 250x250, dtype=np.uint8
+    data = np.zeros((3, 250, 250, 10), dtype=np.uint8)
+    for h in range(10):
+        for i in range(250):
+            for j in range(250):
+                data[0, i, j, h] = i
+                data[1, i, j, h] = j
+                data[2, i, j, h] = 100
+
+    ds_tiff.get_layer("color").get_mag("1").write(data)
```
philippotto: There is no assertion in this test. Maybe just read out again the same bounding box and assert that the output is equal to `data`?

normanrz:
I had a high-level overview and I like it a lot :+1:
What I am missing is something like `get_or_create` for layers, mags etc. Otherwise, the client code has do to these existance checks all the time.
```
ds = WKDataset.create("very_new_dataset", scale=(1, 1, 1))
ds.get_layer("color", "color", create=True).get_mag(1)
ds.get_layer("color", "color", create=True).get_mag(Mag(2,2,1))
```

tests/test_dataset.py
```diff
@@ -0,0 +1,520 @@
...
+    assert path.exists("../testoutput/WKDataset_advanced/color/1")
+    assert path.exists("../testoutput/WKDataset_advanced/color/2-2-1")
+
+    assert len(ds.properties.data_layers) == 1
+    assert len(ds.properties.data_layers["color"].wkw_resolutions) == 2
+
+    assert ds.properties.data_layers["color"].element_class == np.dtype(np.uint16)
+    assert (
+        ds.properties.data_layers["color"].wkw_resolutions[0].cube_length == 64 * 64
+    )  # mag "1"
```
philippotto: Can the `mag=1` value be read from `wkw_resolutions[0]`? Also, `wkw_magnifications[0]` might be a better name? What do you think?

wkcuber/test_api.py
```diff
@@ -0,0 +1,194 @@
...
+
+from wkcuber.api.Dataset import WKDataset, TiffDataset
+from wkcuber.api.Slice import TiffSlice
+import numpy as np
+from skimage import io
+
+if __name__ == "__main__":
+
+    ds = WKDataset.create("/home/robert/Desktop/temp/simple_wk_dataset", [1])
+    mag = ds.add_layer("color", "color", num_channels=3).add_mag(
```
hotzenklotz: Something that I noticed during the demo earlier:  I found it hard to read/understand`ds.add_layer("color", "color"`. One `color` refers to a category and the other to a name, right? Perhaps the categories can be constants/enums. e.g. `Dataset.ColorType` or `DS.ColorLayer` or `DS.COLOR` or something.

`mag = ds.add_layer(Dataset.ColorType, "color", num_channels=3).add_mag(...`

tests/test_dataset.py
```diff
@@ -0,0 +1,520 @@
...
+
+    assert path.exists("../testoutput/TiffDataset/color/1")
+    assert path.exists("../testoutput/TiffDataset/color/2-2-1")
+
+    assert len(ds.properties.data_layers) == 1
+    assert len(ds.properties.data_layers["color"].wkw_resolutions) == 2
+
+
+def test_open_wk_dataset():
+    ds = WKDataset.open("../testdata/simple_wk_dataset")
```
philippotto: This path (and the ones below) use snake_case and the above paths used PascalCase. Could you make this consistent?

tests/test_dataset.py
```diff
@@ -0,0 +1,520 @@
...
+
+    with tiff_slice.open():
+        np.random.seed(1234)
+        write_data = np.random.rand(5, 5, 5).astype(np.uint8)
+
+        tiff_slice.write(np.zeros((5, 5, 5), dtype=np.uint8))
+
+        data = tiff_slice.read((5, 5, 5))
+        assert data.shape == (1, 5, 5, 5)  # this dataset has only one channel
+        assert np.array_equal(data, np.expand_dims(write_data, 0))
```
philippotto: How can `write_data be equal to `data` when `write_data` wasn't written to the dataset in this test? Does it rely on a previous test?

tests/test_dataset.py
```diff
@@ -0,0 +1,520 @@
...
+def test_wk_read_padded_data():
+    filename = "../testoutput/Empty_WKDataset"
+    delete_dir(filename)
+
+    mag = (
+        WKDataset.create(filename, scale=(1, 1, 1))
+        .add_layer("color", "color", num_channels=3)
+        .add_mag("1")
+    )
+    # there are no tiffs yet, however, this should not fail but pad the data with zeros
```
philippotto:
```suggestion
    # there is no data yet, however, this should not fail but pad the data with zeros
```

tests/test_dataset.py
```diff
@@ -0,0 +1,520 @@
...
+
+
+def test_read_and_write_of_properties():
+    destination_path = "../testoutput/read_write_properties/"
+    delete_dir(destination_path)
+    source_file_name = "../testdata/simple_tiff_dataset/datasource-properties.json"
+    destination_file_name = destination_path + "datasource-properties.json"
+
+    imported_properties = TiffProperties.from_json(source_file_name)
+    imported_properties.path = destination_file_name
```
philippotto: Is this a typical usage of the `TiffProperties` class? When would you set the `path` property?

tests/test_dataset.py
```diff
@@ -0,0 +1,520 @@
...
+def test_num_channel_mismatch_assertion():
+    delete_dir("../testoutput/WKDataset")
+
+    ds = WKDataset.create("../testoutput/WKDataset", scale=(1, 1, 1))
+    mag = ds.add_layer("color", "color", num_channels=1).add_mag(
+        "1"
+    )  # num_channel=1 is also the default
+
+    np.random.seed(1234)
+    write_data = np.random.rand(3, 10, 10, 10).astype(np.uint8)  # 3 channel
```
philippotto:
```suggestion
    write_data = np.random.rand(3, 10, 10, 10).astype(np.uint8)  # 3 channels
```

tests/test_dataset.py
philippotto: Since the `open` call doesn't need a corresponding `close` call, what do you think about changing the interface to a simple constructor? Like `WkDataset("...")`? I'd also support @normanrz suggestion for allowing a get-or-create behavior via an additional argument.

wkcuber/api/Dataset.py
```diff
@@ -0,0 +1,175 @@
...
+        # normalize the value of dtype in case the parameter was passed as a string
+        dtype = np.dtype(dtype)
+
+        if layer_name in self.layers.keys():
+            raise IndexError(
+                "Adding layer {} failed. There is already a layer with this name".format(
+                    layer_name
+                )
+            )
+        self.layers[layer_name] = self.__create_layer__(layer_name, dtype, num_channels)
```
philippotto: As far as I know, two underscores in method names are intended for python-internals. A single underscore should be better in such cases. Also, a trailing underscore can be omitted. So, `self._create_layer` would be my suggestion.

wkcuber/api/Dataset.py
```diff
@@ -0,0 +1,175 @@
...
+        self.properties.add_layer(layer_name, category, dtype.name, num_channels)
+        return self.layers[layer_name]
+
+    def get_or_add_layer(
+        self, layer_name, category, dtype=np.dtype("uint8"), num_channels=1
+    ):
+        if layer_name in self.layers.keys():
+            assert self.properties.data_layers[layer_name].category == category, (
+                "Cannot get_or_add_layer: The layer %s already exists, but the dytpes do not match"
```
philippotto:
```suggestion
                "Cannot get_or_add_layer: The layer %s already exists, but the categories do not match"
```

Also, I recommend using f-strings like this:

```
f"Cannot get_or_add_layer: The layer {layer_name} already exists, ...
```

Also, you could include the differing categories in that string. That's always very helpful when the assertion is thrown and you need to debug it :)

wkcuber/api/Dataset.py
```diff
@@ -0,0 +1,175 @@
...
+                % layer_name
+            )
+            assert self.layers[layer_name].dtype == np.dtype(dtype), (
+                "Cannot get_or_add_layer: The layer %s already exists, but the dytpes do not match"
```
philippotto:
```suggestion
                "Cannot get_or_add_layer: The layer %s already exists, but the dtypes do not match"
```

also, f-strings + printing the dtypes.

wkcuber/api/Dataset.py
```diff
@@ -0,0 +1,175 @@
...
+                % layer_name
+            )
+            assert self.layers[layer_name].num_channels == num_channels, (
+                "Cannot get_or_add_layer: The layer %s already exists, but the number of channels do not match"
```
philippotto: Same here.

wkcuber/api/Dataset.py
```diff
@@ -0,0 +1,175 @@
...
+
+    def __create_layer__(self, layer_name, dtype, num_channels):
+        return WKLayer(layer_name, self, dtype, num_channels)
+
+
+class TiffDataset(AbstractDataset):
+    @classmethod
+    def open(cls, dataset_path):
+        properties = TiffProperties.from_json(
+            join(dataset_path, "datasource-properties.json")
```
philippotto: Maybe extract that file name into a constant variable.

wkcuber/api/Layer.py
```diff
@@ -0,0 +1,142 @@
...
+    def __init__(self, name, dataset, dtype, num_channels):
+        self.name = name
+        self.dataset = dataset
+        self.dtype = dtype
+        self.num_channels = num_channels
+        self.mags = {}
+
+        full_path = join(dataset.path, name)
+        try:
+            mkdir(full_path)
```
philippotto: Will this raise an error (which is caught and printed then), when the layer already exists? When opening an existing dataset, it would be expected that the folder exists, right? Maybe `makedirs(..., exist_ok=True)` would make sense?

wkcuber/api/Layer.py
```diff
@@ -0,0 +1,142 @@
...
+                "Deleting mag {} failed. There is no mag with this name".format(mag)
+            )
+
+        del self.mags[mag]
+        self.dataset.properties.delete_mag(self.name, mag)
+        # delete files on disk
+        full_path = join(self.dataset.path, self.name, mag)
+        rmtree(full_path)
+
+    def __create_dir_for_mag__(self, mag):
```
philippotto: `_create_dir_for_mag`

wkcuber/api/Layer.py
```diff
@@ -0,0 +1,142 @@
...
+        full_path = join(self.dataset.path, self.name, mag)
+
+        try:
+            mkdir(full_path)
+        except OSError:
+            logging.info("Creation of MagDataset {} failed".format(full_path))
+
+    def __assert_mag_does_not_exist_yet__(self, mag):
```
philippotto: `_assert_mag_does_not_exist_yet`

wkcuber/api/Layer.py
```diff
@@ -0,0 +1,142 @@
...
+
+        return self.mags[mag]
+
+    def get_or_add_mag(self, mag, block_len=32, file_len=32, block_type=1):
+        # normalize the name of the mag
+        mag = Mag(mag).to_layer_name()
+
+        if mag in self.mags.keys():
+            assert self.mags[mag].header.block_len == block_len, (
+                "Cannot get_or_add_mag: The mag %s already exists, but the block lengths do not match"
```
philippotto: See my other feedback regarding the assertion messages.

philippotto: What happens when I want to `get` a mag (from which I know that it exists), but I don't know with which block_len etc. it was created? Then, I don't have a possibility to use that method, right (since the default block_len value might not match the existing mag)?
I think, you probably need to adapt the default values to `None` and map that to actual default values *only when constructing a new mag*. The same is probably true for all `get_or_create` methods.

wkcuber/api/Layer.py
```diff
@@ -0,0 +1,142 @@
...
+        else:
+            return self.add_mag(mag, block_len, file_len, block_type)
+
+    def setup_mag(self, mag):
+        # This method is used to initialize the mag when opening the Dataset. This does not create e.g. folders.
+
+        # normalize the name of the mag
+        mag = Mag(mag).to_layer_name()
+
+        self.__assert_mag_does_not_exist_yet__(mag)
```
philippotto: Why do you assert this here? The comment above sounds as if this method is also used when the mag already exists? Wouldn't this fail when instantiating an existing dataset?

wkcuber/api/MagDataset.py
```diff
@@ -0,0 +1,135 @@
...
+
+class MagDataset:
+    def __init__(self, layer, name):
+        self.mag = {}
+        self.layer = layer
+        self.name = name
+        self.header = self.get_header()
+
+        file_path = join(self.layer.dataset.path, self.layer.name, self.name)
+        size = self.layer.dataset.properties.data_layers[
```
philippotto: What's the difference between `self.layer` and `self.layer.dataset.properties.data_layers[self.layer.name]`? Adding python type annotations might be helpful here, since the types also act as documentation.

wkcuber/api/MagDataset.py
```diff
@@ -0,0 +1,135 @@
...
+            self.layer.name
+        ].get_bounding_box_size()
+        offset = self.layer.dataset.properties.data_layers[
+            self.layer.name
+        ].get_bounding_box_offset()
+        self.slice = self.__get_slice_type__()(
```
philippotto: 1) Why can't you use `get_slice()` here? If that method accesses `self.header` then the signature of `get_slice` should match your requirements here.
2) Maybe it's a good idea to add an `is_bounded` parameter to `get_slice`? That way, you don't need to mutate the slice object afterwards.

wkcuber/api/MagDataset.py
```diff
@@ -0,0 +1,135 @@
...
+        return self.slice.read(size, offset)
+
+    def write(self, data, offset=(0, 0, 0)):
+        self.__assert_valid_num_channels__(data.shape)
+        self.slice.write(data, offset)
+        layer_properties = self.layer.dataset.properties.data_layers[self.layer.name]
+        current_offset = layer_properties.get_bounding_box_offset()
+        current_size = layer_properties.get_bounding_box_size()
+
+        if self.slice.check_bounds(current_offset, current_size):
```
philippotto: Why do you call `check_bounds` if the slice is never bound for MagDatasets anyway?

wkcuber/api/MagDataset.py
```diff
@@ -0,0 +1,135 @@
...
+            new_offset = (
+                offset
+                if current_offset == (-1, -1, -1)
+                else tuple(min(x) for x in zip(current_offset, offset))
+            )
+            total_size = tuple(max(x) for x in zip(current_size, data.shape[-3:]))
+            self.slice.size = total_size
+            # no need to update the offset of the slice, because it is always the absolute offset (0, 0, 0)
+
+            self.layer.dataset.properties.set_bounding_box_size_of_layer(
```
philippotto: Maybe provide a method like `set_bounding_box_of_layer` to which you pass `total_size` and `new_offset`. This makes the code here more concise and the properties are only updated once on disk, if I understand this correctly.

wkcuber/api/MagDataset.py
```diff
@@ -0,0 +1,135 @@
...
+            voxel_type=self.layer.dtype,
+            num_channels=self.layer.num_channels,
+            version=1,
+            block_len=self.block_len,
+            file_len=self.file_len,
+            block_type=self.block_type,
+        )
+
+    def get_slice(self, mag_file_path, size, global_offset):
+        return WKSlice(mag_file_path, self.get_header(), size, global_offset)
```
philippotto: Couldn't you use `self.header` directly instead of `self.get_header()`? Alternatively, you could change the `get_header` method to a `initialize_header` method which sets `self.header` only when it wasn't initialized yet (`if self.header is None: self.header = wkw.Header...`).

tests/test_dataset.py
rschwanhold: Good catch. The random values of `np.random.rand` are between 0 and 1. Converting them to uint8 results in zeros. This was also the case in other tests.

tests/test_dataset.py
rschwanhold: Yes, the mag can be read from `wkw_resolutions[0]`. I have added this to the test

tests/test_dataset.py
rschwanhold: No, this is not the typical usage of the `TiffProperties`. The developer does not need to do this, because this is handled by the Dataset. However, I do it explicitly in this test, to test if reading and writing properties work.

wkcuber/api/Slice.py
```diff
@@ -0,0 +1,120 @@
...
+    def check_bounds(self, offset, size):
+        for s1, s2, off in zip(self.size, size, offset):
+            if s2 + off > s1 and self.is_bounded:
+                return False
+        return True
+
+    def assert_bounds(self, offset, size):
+        if not self.check_bounds(offset, size):
+            raise AssertionError(
+                "Writing out of bounds: The parameter 'size' {} is not compatible with the attribute 'size' {}".format(
```
philippotto: Can we make this error message more comprehensible? For example, `the passed parameter "size" exceeds the size of the current slice` or something similar?

wkcuber/api/Slice.py
```diff
@@ -0,0 +1,120 @@
...
+
+from wkcuber.api.TiffData.TiffMag import TiffMag
+
+
+class Slice:
+    def __init__(
+        self,
+        path_to_mag_dataset,
+        header,
+        size=(1024, 1024, 1024),
```
philippotto: Why does this default to 1024? What do you think about making this parameter non-optional?

rschwanhold:
@philippotto the build never starts. Do I know what I can do against it?

PR #175, opened 2019-12-17
Poetry

normanrz:
Use poetry as dependency manager instead of pip/setuptools.

My motivation for this PR was the unification of dependencies in `requirements.txt` and `setup.py` which were out-of-sync.


pyproject.toml
```diff
@@ -0,0 +1,29 @@
+[tool.poetry]
+name = "wkcuber"
+version = "0.0.0"
```
philippotto: The correct version is read from the GH releases page through `setuptools_scm`, right? Can we omit this property completely here? Otherwise, maybe add comment mentioning `setuptools_scm`.

PR #182, opened 2020-02-06
Enable Compression for NIFTI Conversion Script

hotzenklotz:
PR changes: 
- Enables wkw Compression for NIFTI Conversion Script
- Reduce wkw file length to 256 (instead of 1024) to account for smaller MRI datasets


hotzenklotz:
@leowe Do you have any objections to writing compressed wkw files for the NIFTI conversion?

PR #184, opened 2020-02-29
High level dataset api chunking

rschwanhold:
- [x] Wait until #178 is merged

- After this PR is merged, create a new PR in voxelytics in order to use the `BoundingBox`-code that was copied into webknossos-cuber in this PR


wkcuber/api/View.py
```diff
@@ -99,6 +123,20 @@ def open(self):
...
+
+        assert chunk_size is not None
+
+        if 0 in chunk_size:
+            raise AssertionError(
+                f"The passed parameter 'chunk_size' {chunk_size} contains at least one 0. This is not allowed."
+            )
+        if (np.array(chunk_size) % file_dim).any():
+            raise AssertionError(
+                f"The passed parameter 'chunk_size' {chunk_size} must be a multiple of the file size {file_dim}"
```
normanrz: Does that mean when using files with cube length 1024, only multiples of 1024 are valid chunk sizes? Given the structure of WKW, I would rather enforce the chunk size to be a power of 2 and ≥ 32.

rschwanhold: I changed the behavior of this method as you suggested.

rschwanhold:
@normanrz 
In the last commit, I added the function `get_view` for `View`; renamed the parameter `global_offset` of `get_view` to `offset`; and adjusted the method `_check_chunk_size`.

I also noticed that it was possible to get a `View` which is bounded (meaning that the size of the view must never be exceeded)  with a size that is bigger than the size of the dataset itself. In fact, this was the case in many test cases. I fixed this by changing the size of the view or making the view not bounded. (This is why there are many new code changes)

PR #190, opened 2020-03-05
Use the most recent BoundingBox implementation from voxelytics

rschwanhold:
The class `BoundingBox`, which was added in #184, does not reflect the state of the current master of voxelytics (the code in #184 was copied from an older commit).

### DIFF
--- a/wkcuber/api/bounding_box.py
+++ b/wkcuber/api/bounding_box.py
@@ -1,9 +1,16 @@
-from typing import Tuple, Union, List, Dict, Generator, Optional
+# mypy: allow-untyped-defs
 import json
+from typing import Dict, Generator, Iterable, List, Optional, Tuple, Union, NamedTuple
+
 import numpy as np
 from wkcuber.mag import Mag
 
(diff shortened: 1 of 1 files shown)


wkcuber/api/bounding_box.py
```diff
@@ -1,60 +1,83 @@
...
 import json
+from typing import Dict, Generator, Iterable, List, Optional, Tuple, Union, NamedTuple
+
 import numpy as np
 from wkcuber.mag import Mag
 
 Shape3D = Union[List[int], Tuple[int, int, int], np.ndarray]
 
 
+class BoundingBoxNamedTuple(NamedTuple):
```
rschwanhold: In the code in the voxelytics repo this class belongs to `voxelytics.configs.datasources.schema.py` and was imported here. I added this class here to avoid that wkcuber gets dependent on voxelytics.

philippotto: Great! If you do the PR in voxelytics, you can remove BoundingBoxNamedTuple by importing it from wkcuber.

PR #205, opened 2020-04-27
Fix bounding box of mag != 1 when exporting wkw to tiff

### LINKED ISSUE #203: Export to tiff with mag!=1 has wrong bbox
When exporting tiffs in a mag > 1, it reads the correct data but keeps the bbox of mag 1. So there is way too much 0-data.
In general I think it makes sense to treat bboxes as in mag 1 (also when specifying an explicit bbox as cli argument), but of course they need to be handled properly when extracting higher mag data.

downscaled_but_wrong_bbox

philippotto:
I also swapped the tested dataset to `simple_wk_dataset` instead of `WT1_wkw.tar.gz`. The latter seems to be empty for me?? When importing it into webKnossos it's only black :thinking: 

fixes #203

### DIFF
--- a/test.sh
+++ b/test.sh
@@ -1,3 +1,4 @@
+set -euxo pipefail
 export CIRCLE_BUILD_NUM=latest
 tests/scripts/build_docker_image.sh
 tests/scripts/decompress_test_data.sh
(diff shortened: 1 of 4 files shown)


normanrz:
Did you also test whether it works when not supplying the `--bbox` cli arg?

philippotto:
> Did you also test whether it works when not supplying the `--bbox` cli arg?

Yes, I just tested this.

PR #211, opened 2020-05-28
Assert that bounding box and mag are aligned

georgwiese:
Towards fixing https://github.com/scalableminds/voxelytics/issues/1650, I added the following to our `BoundingBox` class:
- `BoundingBox.in_mag()` does not to any implicit rounding (hence does not have a `ceil` attribute anymore). Instead, it fails hard if the bounding box cannot unambiguously be converted to a given mag.
- `BoundingBox.align_with_mag()` implements what you'll typically want to do in that case: round the bounding box up (i.e., make it larger) until it is mag aligned.

These are a few test cases:
```
>>> BoundingBox((1, 1, 1), (10, 10, 10)).align_with_mag(Mag(2))
BoundingBox(topleft=(0, 0, 0), size=(12, 12, 12))
>>> BoundingBox((1, 1, 1), (9, 9, 9)).align_with_mag(Mag(2))
BoundingBox(topleft=(0, 0, 0), size=(10, 10, 10))
>>> BoundingBox((1, 1, 1), (9, 9, 9)).align_with_mag(Mag(4))
BoundingBox(topleft=(0, 0, 0), size=(12, 12, 12))
>>> BoundingBox((1, 2, 3), (9, 9, 9)).align_with_mag(Mag(2))
BoundingBox(topleft=(0, 2, 2), size=(10, 10, 10))
>>> BoundingBox((1, 2, 3), (9, 9, 9)).in_mag(Mag(2))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/georg/coding/webknossos-cuber/wkcuber/api/bounding_box.py", line 194, in in_mag
    ), f"topleft {self.topleft} is not aligned with the mag {mag}. Use BoundingBox.align_with_mag()."
AssertionError: topleft [1 2 3] is not aligned with the mag 2. Use BoundingBox.align_with_mag().
>>> BoundingBox((2, 2, 2), (9, 9, 9)).in_mag(Mag(2))
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/Users/georg/coding/webknossos-cuber/wkcuber/api/bounding_box.py", line 197, in in_mag
    ), f"bottomright {self.bottomright} is not aligned with the mag {mag}. Use BoundingBox.align_with_mag()."
AssertionError: bottomright [11 11 11] is not aligned with the mag 2. Use BoundingBox.align_with_mag().
>>> BoundingBox((2, 2, 2), (10, 10, 10)).in_mag(Mag(2))
[...]

PR #217, opened 2020-06-09
Floored version of align_with_mag()

georgwiese:
Follow-up to #211 

When bounding boxes are not aligned, we will also sometimes want a floored version of `align_with_mag()`. For example, when we read annotated training data, we want to make sure we never read voxels that are not annotated.

### DIFF
--- a/wkcuber/api/bounding_box.py
+++ b/wkcuber/api/bounding_box.py
@@ -209,7 +209,8 @@ class BoundingBox:
 
         np_mag = np.array(mag.to_array())
 
-        align = lambda point, round_fn: return round_fn(point / np_mag).astype(np.int) * np_mag
+        align = lambda point, round_fn: round_fn(point / np_mag).astype(np.int) * np_mag
+
         if ceil:
             topleft = align(self.topleft, np.floor)
             bottomright = align(self.bottomright, np.ceil)
--- a/wkcuber/api/bounding_box.py
+++ b/wkcuber/api/bounding_box.py
@@ -204,7 +204,7 @@ class BoundingBox:
     def align_with_mag(self, mag: Mag, ceil=False):
         """Rounds the bounding box, so that both topleft and bottomright are divisible by mag.
 
-        :argument ceil: If true, we read more, otherwise less.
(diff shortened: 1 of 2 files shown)


wkcuber/api/bounding_box.py
```diff
@@ -201,12 +201,21 @@ def in_mag(self, mag: Mag) -> "BoundingBox":
             size=(self.size // np_mag).astype(np.int),
         )
 
-    def align_with_mag(self, mag: Mag):
-        """Rounds the bounding box up, so that both topleft and bottomright are divisible by mag."""
+    def align_with_mag(self, mag: Mag, ceil=False):
+        """Rounds the bounding box, so that both topleft and bottomright are divisible by mag.
+
+        :argument ceil: If true, we read more, otherwise less.
```
philippotto, line 207:
```suggestion
        :argument ceil: If true, the bounding box is enlarged when necessary. If false, it's shrinked when necessary.
```

wkcuber/api/bounding_box.py
```diff
@@ -201,12 +201,21 @@ def in_mag(self, mag: Mag) -> "BoundingBox":
...
-        topleft = (self.topleft // np_mag).astype(np.int) * np_mag
-        bottomright = np.ceil(self.bottomright / np_mag).astype(np.int) * np_mag
+
+        if ceil:
+            topleft = np.floor(self.topleft / np_mag).astype(np.int) * np_mag
+            bottomright = np.ceil(self.bottomright / np_mag).astype(np.int) * np_mag
+        else:
+            topleft = np.ceil(self.topleft / np_mag).astype(np.int) * np_mag
+            bottomright = np.floor(self.bottomright / np_mag).astype(np.int) * np_mag
+
```
jstriebel, line 218:
```suggestion

        align = lambda point, round_fn: return round_fn(point / np_mag).astype(np.int) * np_mag
        if ceil:
            topleft = align(self.topleft, np.floor)
            bottomright = align(self.bottomright, np.ceil)
        else:
            topleft = align(self.topleft, np.ceil)
            bottomright = align(self.bottomright, np.floor)
```

jstriebel: Just one suggestion, everything else LGTM :+1:

PR #222, opened 2020-07-15
Add various improvements to nifti conversion and fix tiff-api transpose bug

philippotto:
- rewrite nifti conversion tool to use high-level dataset api
- support tiff output in nifti conversion tool
- fix transpose bug in high-level dataset api for tiff
- add `--use_orientation_header` flag so that nifti output is oriented correctly (under the assumption that the header is correct which is not always the case)
- add `--enforce_bounding_box` flag to enforce a certain output bounding box when converting nifti files
- automatically detect binary layers when converting nifti files and cast these to 0 and 1, instead of 0 and 255
- change default file_len (so that the output of nifti is not in the magnitude of gigabytes anymore) and refactor corresponding constants


wkcuber/api/TiffData/TiffMag.py
```diff
@@ -414,11 +414,11 @@ def open(cls, file_name):
         return cls(file_name)
 
     def read(self) -> np.array:
-        return io.imread(self.file_name)
+        return io.imread(self.file_name).transpose()
 
     def write(self, pixels):
         os.makedirs(os.path.dirname(self.file_name), exist_ok=True)
-        io.imsave(self.file_name, pixels, check_contrast=False)
+        io.imsave(self.file_name, pixels.transpose(), check_contrast=False)
```
philippotto, line 421: cc @rschwanhold: I'm confident that this fix is correct, but wanted to let you know, anyway (and feel free to doublecheck!). The bug is especially tricky to find, since it's intransparent when only using the API (since input and output are transposed, thus, effectively cancelling each other out). One has to inspect the tiff files by hand to see whether something is flipped or not. Of course, that's even harder to do, when the tests generate procedural images ^^

wkcuber/convert_nifti.py
```diff
@@ -54,13 +60,41 @@ def create_parser():
...
+        "--write_tiff",
+        help="Output tiff dataset instead of wkw.",
+        default=False,
+        action="store_true",
+    )
+
+    parser.add_argument(
+        "--use_orientation_header",
+        help="Use orientation information from header to interpret the input data (should be tried if output orientation seems to be wrong).",
+        default=False,
```
philippotto, line 73: I set this to False for now, since other nifti files did not seem to use this header correctly If the output is flipped, one can try to toggle this flag.

wkcuber/convert_nifti.py
```diff
@@ -72,129 +106,200 @@ def to_target_datatype(data: np.ndarray, target_dtype) -> np.ndarray:
...
-                )
+    is_probably_binary = np.unique(cube_data).shape[0] == 2
+    assume_segmentation_layer = False  # Since webKnossos does not support multiple segmention layers, this is hardcoded to False right now.
+
+    max_cell_id_args = (
+        {"largest_segment_id": int(np.max(cube_data) + 1)}
+        if assume_segmentation_layer
+        else {}
+    )
+    category_type = "segmentation" if assume_segmentation_layer else "color"
```
hotzenklotz, line 143: Just to clarify, this means all Niftis will be converted to wK color layers, right? Even the ground truth segmentations, right? (Those used o be a wK `segmentation` layer...? I guess it does not matter much. The ML can probably read GT from either color or segmentation layers.

wkcuber/convert_nifti.py
```diff
@@ -72,129 +106,200 @@ def to_target_datatype(data: np.ndarray, target_dtype) -> np.ndarray:
...
+def main():
     args = create_parser().parse_args()
     setup_logging(args)
 
     source_path = Path(args.source_path)
 
+    bbox_to_enforce = None
+    if args.enforce_bounding_box is not None:
+        bbox_tuple = tuple(int(x) for x in args.enforce_bounding_box.split(","))
+        print("bbox_tuple", bbox_tuple)
```
hotzenklotz, line 276: Ups, `print`

wkcuber/convert_nifti.py
```diff
@@ -72,129 +106,200 @@ def to_target_datatype(data: np.ndarray, target_dtype) -> np.ndarray:
...
 
-if __name__ == "__main__":
+def main():
     args = create_parser().parse_args()
     setup_logging(args)
 
     source_path = Path(args.source_path)
 
+    bbox_to_enforce = None
+    if args.enforce_bounding_box is not None:
```
hotzenklotz, line 274: bounding box CLI parsing seems like a common task. It might make sense to move it into wkcuber utils. See how `add_scale` and `parse_scale` are done. Just an idea.

hotzenklotz:
@philippotto One question concerning the flipping and canonical transformation matrix: When we last looked at the mulder data, the lesion segmentation should have been on the right-hand side of the brain. Is this the case, when using this script?

wkcuber/convert_nifti.py
philippotto: > Just to clarify, this means all Niftis will be converted to wK color layers, right? Even the ground truth segmentations, right?

Yes! And this has been like this before, so it's not a step back. I intended to auto-detect this and set the correct category, but then ran into problems, because WK does not support multiple segmentation layers. Then, I changed it back to always use color as category, but used the inferred type information to improve the type casting, so that we don't have the 0 vs. 255 issue with segmentation layers.

philippotto:
> @philippotto One question concerning the flipping and canonical transformation matrix: When we last looked at the mulder data, the lesion segmentation should have been on the right-hand side of the brain. Is this the case, when using this script?

Unfortunately, the lesion will be on the left side when only using `--use_orientation_header`. However, this matches papaya's rendering, so I'd leave it as is.
To make these pesky cases easier in the future, I just added another `--flip_axes` argument with which we can easily correct misinformation in the header. For example, `--flip_axes 1,2` would flip the x and y axis during conversion.

PR #225, opened 2020-07-20
Add possibility to change the boundingbox of a layer

### LINKED ISSUE #223: Allow explicitly setting a layer's bounding box
For example, when one cuts data away and knows that the bounding box needs to be shrunken, there is no way of setting the box explicitly with the api.

rschwanhold:
- fixes #223


tests/test_dataset.py
```diff
@@ -1146,3 +1146,37 @@ def test_tiled_tiff_dataset_get_or_create():
...
+    layer = ds.get_layer("color")
+    mag = layer.get_mag("1")
+
+    bbox_offset = ds.properties.data_layers["color"].get_bounding_box_offset()
+    bbox_size = ds.properties.data_layers["color"].get_bounding_box_size()
+    assert bbox_size == (265, 265, 10)
+    original_data = mag.read(bbox_size)
+    assert original_data.shape == (1, 265, 265, 10)
+
+    layer.set_bounding_box(bbox_offset, (100, 100, 10))  # decrease boundingbox
```
rschwanhold, line 1168: @philippotto do you prefer one method that can change the offset and the size at the same time or should I split this into two methods?

philippotto: I think, it would be cool to have all three methods :)

wkcuber/api/Layer.py
```diff
@@ -60,6 +64,28 @@ def _assert_mag_does_not_exist_yet(self, mag):
...
+        self.set_bounding_box_offset(offset)
+        self.set_bounding_box_size(size)
+
+    def set_bounding_box_offset(self, offset: Tuple[int, int, int]):
+        size = self.dataset.properties.data_layers["color"].get_bounding_box_size()
+        self.dataset.properties._set_bounding_box_of_layer(
+            self.name, tuple(offset), tuple(size)
+        )
+        for _, mag in self.mags.items():
+            mag.view.global_offset = offset
```
rschwanhold, line 79: What is the expected behavior for views that were created before the boundingbox was changed?
I have now updated the size and offset of the views that are representing the different magnifications of the layer. However, if the user creates a view of a smaller boundingbox explicitly (with `ds.get_view(...)`)  (before the bbox was changed) then the "global boundingbox" of the view does not get updated. The reason is, that te layer itself does not keep track of all Views that have been created. If this is a problem, the View might need to look up the "global boundingbox" from the properties.json each time the `read` or `write` method is called.

philippotto: Good point! I think, it's alright if a view sticks to the global bbox which was present at the time when the view was created (that would also mean that voxel addresses stay consistent throughout the lifetime of a view, right?). Setting the bbox explicitly shouldn't be a frequent option. Also it should be done either before creating a view or after working with a view. In these cases, no harm can be done, anyway. So, it sounds fine to me :+1:

PR #231, opened 2020-08-07
Add check-equality task

philippotto:
wkcuber.check_equality can be used to compare two datasets for equalness (for example, after compressing a dataset, one can assert that all bits are where they are supposed to be). See readme or test for usage examples.

### DIFF
--- a/wkcuber/compress.py
+++ b/wkcuber/compress.py
@@ -47,8 +47,6 @@ def create_parser():
         "--mag", "-m", nargs="*", help="Magnification level", default=None
     )
 
-    parser.add_argument("--verify", default=False, action="store_true")
-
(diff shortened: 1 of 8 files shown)


wkcuber/compress.py
```diff
@@ -46,6 +47,8 @@ def create_parser():
         "--mag", "-m", nargs="*", help="Magnification level", default=None
     )
 
+    parser.add_argument("--verify", default=False, action="store_true")
```
georgwiese, line 50: What happens with this flag? Can't find any use for it?

Also, please add a description.

philippotto: Sorry that is an obsolete artifact. I removed it again.

PR #236, opened 2020-08-20
add check for division by zero

### LINKED ISSUE #214: NIFTI Division by zero error
When converting a nifti file only consisting of zeros, the to_target_datatype function divides by 0.

leowe:
Add a simple check to prevent a division by zero. Since this case can happen in every case in the conditional in the function, this was the easiest solution.

closes #214

### DIFF
--- a/wkcuber/convert_nifti.py
+++ b/wkcuber/convert_nifti.py
@@ -102,6 +102,9 @@ def to_target_datatype(
     else:
         factor = np.iinfo(data.dtype).max / np.iinfo(target_dtype).max
 
+    if factor == 0:
+        factor = 1
+
     return (data / factor).astype(np.dtype(target_dtype))
 
 


wkcuber/convert_nifti.py
```diff
@@ -102,6 +102,9 @@ def to_target_datatype(
     else:
         factor = np.iinfo(data.dtype).max / np.iinfo(target_dtype).max
 
+    if factor == 0:
+        factor = 1
```
philippotto, line 106: I think, this can only happen when `data.max() == 0`, right? (at least with the current implementation)

I'd suggest to change the if-conditional accordingly, to make this fallback more strict. In general, dividing-by-zero is a bug and just doing "nothing" (i.e., dividing by 1) might not always be the correct way to deal with edge cases. If the implementation changes at some point and introduces a new divide-by-zero bug, the `if factor == 0` would just silently shadow it (which might be correct or wrong).

I'd suggest the following:
```
if data.max() == 0:
  logging.warning("Don't rescale data since maximum is 0")
  factor = 1
```

Do you agree?

Issue #234, opened 2020-08-12
Dont interpret coordinates relative to dataset bounding box


philippotto:
I just debugged some code of mine and finally found out that the high-level dataset api does not directly use the passed offsets (for reading and writing). Instead, the offset is interpreted relative to the global bounding box which is specified in the datasource-properties.json. I think this is problematic and not the behavior which is expected in most cases.

For the "Views" which are bounded to a specific bounding box, it seems reasonable to stick to the "relativeness", but when accessing the entire dataset, I'd expect to use the global coordinate system. Do you agree, @normanrz @rschwanhold ? This would also align with the "old wkw api access".

- [x] discuss
- [x] double-check whether existing code already relied on the old behavior
- [x] change implementation

rschwanhold:
I wasn't aware, that the "old wkw api" did it differently. I agree with you that we should change this.

That raises the question for me: What is the purpose of the offset in the properties.json? Are there other use-cases than to restrict a view?

philippotto:
> I agree with you that we should change this.

Great! Could you do this? :)

> That raises the question for me: What is the purpose of the offset in the properties.json? Are there other use-cases than to restrict a view?

In general, it's used to know "where the data is". Either in webKnossos when viewing, but also when doing automatic processing (e.g, segmenting the data) it's used to read the correct bounding box.

rschwanhold:
@philippotto I would like to get your opinion on how the bounding box of views should be calculated. (Maybe I gave too much attention to the offset in the properties.json)
- Current behavior:
  - the `MagDataset` has a `View` which has its origin at the offset that is defined in the properties.json
  - `MagDataset.get_view` takes two optional parameters (`size` and `offset`) which are combined with the size and offset from the properties.json
    - the offset of the new view is the addition of the offset from the properties and the passed parameter
    - if the size is specified as a parameter, that size is used for the new view (otherwise the value from the properties is used)
  - A `View` itself also has a function `get_view` (to get a sub-view). The offset of the new view is relative to the view that created it

- Possible new behavior:
  - the `MagDataset` has a `View` which has its origin always at (0, 0, 0)
  - `get_view` takes two optional parameters (`size` and `offset`) 
    - if the parameters `size` and `offset` are not specified, the size and offset from the properties.json are used
    - what is the expected behavior if the parameters are specified?
      - should the offset be interpreted relative to (0, 0, 0) or to the offset from the properties.json?
[...]

PR #237, opened 2020-08-26
Interpret offsets for MagDataset.get_view as absolute values

rschwanhold:
The parameter `offset` for `MagDataset.get_view` was previously interpreted relative to the offset in the datasource-properties.json. Now, the parameter is interpreted as an absolute value.

I did not find any usages of `get_view` in voxelytics (and only my own usages in wkcuber). Therefore, I conclude that there are no implementations that rely on the old behavior.

fixes: #234

### DIFF
--- a/wkcuber/api/MagDataset.py
+++ b/wkcuber/api/MagDataset.py
@@ -79,17 +79,24 @@ class MagDataset:
             offset = offset_in_properties
 
         if size is None:
-            size = np.array(size_in_properties) - np.array(offset)
+            size = np.array(size_in_properties) - (
+                np.array(offset) - np.array(offset_in_properties)
+            )
 
-        # assert that the parameter size is valid
-        for s1, s2, off in zip(size_in_properties, size, offset):
-            if s2 + off > s1 and is_bounded:
-                raise AssertionError(
-                    f"The combination of the passed parameter 'size' {size} and {offset} are not compatible with the "
-                    f"size ({size_in_properties}) from the properties.json."
-                )
+        # assert that the parameters size and offset are valid
+        if is_bounded:
+            for off_prop, off in zip(offset_in_properties, offset):
+                if off < off_prop:
+                    raise AssertionError(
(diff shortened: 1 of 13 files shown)


wkcuber/api/MagDataset.py
```diff
@@ -68,30 +68,38 @@ def get_view(self, size=None, offset=None, is_bounded=True):
...
+            size = np.array(size_in_properties) - (
+                np.array(offset) - np.array(offset_in_properties)
+            )
+
+        # assert that the parameters size and offset are valid
+        if is_bounded:
+            for off_prop, off in zip(offset_in_properties, offset):
+                if off < off_prop:
+                    raise AssertionError(
+                        f"The passed parameter 'offset' {offset} is outside the bounding box from the properties.json."
```
philippotto, line 91:
```suggestion
                        f"The passed parameter 'offset' {offset} is outside the bounding box from the properties.json. Use is_bounded=False if you intend to write outside out the existing bounding box."
```

wkcuber/api/MagDataset.py
```diff
@@ -68,30 +68,38 @@ def get_view(self, size=None, offset=None, is_bounded=True):
...
+                    )
+            for s1, s2, off in zip(size_in_properties, size, offset):
+                if s2 + off > s1:
+                    raise AssertionError(
+                        f"The combination of the passed parameter 'size' {size} and 'offset' {offset} are not compatible with the "
+                        f"size ({size_in_properties}) from the properties.json."
```
philippotto, line 97:
```suggestion
                        f"The combination of the passed parameter 'size' {size} and 'offset' {offset} are not compatible with the "
                        f"size ({size_in_properties}) from the properties.json.  Use is_bounded=False if you intend to write outside out the existing bounding box."
```

philippotto (approved):
Excellent! Thank you for taking care of this issue. My two suggestions probably need some formatting afterwards :see_no_evil:

Issue #235, opened 2020-08-13
High-level API cannot interpret multi-channel data ("uint24")


philippotto:
```data type "uint24" not understood``` is thrown when an RGB layer should be opened.

philippotto:
In the past, datasource-properties.json contained "uint24" for three-channel layers. However, the new API probably would expect uint8 (since one channel consists of 8 bit). Not sure how we want to unify this..

daniel-wer:
This affects voxelytics quite a bit as most connect datasets contain a uint24 prediction layer. Those datasets cannot be opened by the high-level API.

Issue #229, opened 2020-08-05
Make read interface of MagDataset consistent with WkwDataset


philippotto:
wkw `read` takes offset and size
MagDataset `read` takes size and offset

Probably, MagDataset (and the View) need to be changed. Then, we would need to release a breaking change and update it everywhere where it's used. Not sure, whether we can assert somehow that the order of the parameters is correct?

PR #239, opened 2020-09-22
Dataset API changes: parameter 'dtype' is now interpreted as 'dtype_per_layer' and parameter list for 'read(...)' was reordered to match wkw.Dataset.read(...)

rschwanhold:
- The Dataset API now interprets the dtype per layer (not per channel)
- Reorder parameter list for `MagDataset.read(...)` and `View.read(...)` to match the order of `wkw.Dataset.read(...)`
- fixes: #235 
- fixes: #229

### DIFF
--- a/wkcuber/api/Dataset.py
+++ b/wkcuber/api/Dataset.py
@@ -5,6 +5,7 @@ from os.path import join, normpath, basename
 from pathlib import Path
 import numpy as np
 import os
+import re
 
 from wkcuber.api.Properties.DatasetProperties import (
     WKProperties,
@@ -13,6 +14,29 @@ from wkcuber.api.Properties.DatasetProperties import (
 )
 from wkcuber.api.Layer import Layer, WKLayer, TiffLayer, TiledTiffLayer
 from wkcuber.api.View import View
+
+
+def is_int(s: str) -> bool:
+    try:
+        int(s)
+        return True
+    except ValueError:
+        return False
+
+
+def dtype_per_layer_to_dtype_per_channel(dtype_per_layer, num_channels):
+    if dtype_per_layer is None:
+        return None
+    # split the dtype into the actual type and the number of bits
+    # example: "uint24" -> ["uint", "24"]
+    dtype_per_layer_parts = re.split('(\d+)', str(dtype_per_layer))
+    # calculate number of bits for dtype_per_channel
+    dtype_per_channel_parts = [(str(int(int(part) / num_channels)) if is_int(part) else part) for part in
+                               dtype_per_layer_parts]
+    try:
+        return np.dtype(''.join(dtype_per_channel_parts))
+    except TypeError as e:
+        raise TypeError("Converting dtype_per_layer to dtype_per_channel failed. Double check if the dtype_per_layer value is correct. " + str(e))
 
 
 class AbstractDataset(ABC):
@@ -72,14 +96,11 @@ class AbstractDataset(ABC):
             )
         return self.layers[layer_name]
 
-    def add_layer(self, layer_name, category, dtype=None, num_channels=None, **kwargs):
-        if dtype is None:
-            dtype = np.dtype("uint8")
+    def add_layer(self, layer_name, category, dtype_per_layer=None, num_channels=None, **kwargs):
         if num_channels is None:
             num_channels = 1
-
-        # normalize the value of dtype in case the parameter was passed as a string
-        dtype = np.dtype(dtype)
+        if dtype_per_layer is None:
+            dtype_per_layer = "uint" + str(8*num_channels)
 
         if layer_name in self.layers.keys():
             raise IndexError(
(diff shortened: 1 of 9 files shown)


wkcuber/api/Dataset.py
```diff
@@ -15,6 +16,34 @@
...
+    if dtype_per_layer is None:
+        return None
+    # split the dtype into the actual type and the number of bits
+    # example: "uint24" -> ["uint", "24"]
+    dtype_per_layer_parts = re.split("(\d+)", str(dtype_per_layer))
+    # calculate number of bits for dtype_per_channel
+    dtype_per_channel_parts = [
+        (str(int(int(part) / num_channels)) if is_int(part) else part)
+        for part in dtype_per_layer_parts
+    ]
```
rschwanhold, line 37: Let's say someone creates a layer with 3 channels.
If that person would use `"int"` as `dtype_per_layer`, then `dtype_per_channel` would still be `int` and therefore valid. However, the normal `int` has 32 bits, which means that one channel should have `32/3` bits (which does not work). 
Should I simply add a check that the `dtype_per_layer` must have a number in it (e.g. `uint24`) if the layer has multiple channel?

philippotto: > Should I simply add a check that the dtype_per_layer must have a number in it (e.g. uint24) if the layer has multiple channel?

Yes, I think, it is fair to require the dtype to be specific (i.e., also contain a number).

wkcuber/api/Dataset.py
```diff
@@ -72,40 +101,58 @@ def get_layer(self, layer_name) -> Layer:
...
+    def add_layer(
+        self, layer_name, category, dtype_per_layer=None, num_channels=None, **kwargs
+    ):
         if num_channels is None:
             num_channels = 1
-
-        # normalize the value of dtype in case the parameter was passed as a string
-        dtype = np.dtype(dtype)
+        if dtype_per_layer is None:
+            dtype_per_layer = "uint" + str(8 * num_channels)
```
philippotto, line 110: Maybe extract the 8 into a `DEFAULT_BIT_DEPTH` constant at the top of the module.

philippotto (changes requested):
Awesome! I like how you changed the dtype very clearly to `dtype_per_channel` and `dtype_per_layer`.

I have two suggestions:
- allow to pass a numpy dtype when adding a layer (strings need to work still for funky dtypes as uint48, but when specifying `uint16`, it would be nice to be able to pass `np.uint16`)
- allow to pass `dtype_per_channel` when adding a layer. I think, often that will feel more intuitive (also, that's how it was done before). Then, the code should assert that not both arguments are passed.

Also, we need to think about API backwards compatibility. One could keep the `dtype` argument and alias it to `dtype_per_channel` and optionally emit a warning that one should use the clearer/newer argument name `dtype_per_channel` directly. I think, that could be a good migration path. Do you agree?

wkcuber/api/Dataset.py
```diff
@@ -72,40 +130,88 @@ def get_layer(self, layer_name) -> Layer:
...
+                raise TypeError(
+                    "Cannot add layer. The specified 'dtype_per_channel' must be a valid dtype. "
+                    + str(e)
+                )
+            dtype_per_layer = dtype_per_channel_to_dtype_per_layer(
+                dtype_per_channel, num_channels
+            )
+        elif dtype_per_layer is not None:
+            if type(dtype_per_layer) is not str:
+                dtype_per_layer = str(np.dtype(dtype_per_layer))
```
rschwanhold, line 162: An alternative to these lines would be to wrap a try-catch block around it and simply try to convert `dtype_per_layer` to a valid dtype. This would make the assertion that there is a digit in the dtype redundant (see `convert_dtypes`) because `str(np.dtype("int"))` results in `int64`.
What do you prefer?

rschwanhold:
> Also, we need to think about API backwards compatibility. One could keep the `dtype` argument and alias it to `dtype_per_channel` and optionally emit a warning that one should use the clearer/newer argument name `dtype_per_channel` directly. I think, that could be a good migration path. Do you agree?

Regarding the alias:
Do you mean something like this (making use of the kwargs) ...
```
def add_layer(self, layer_name, category, dtype_per_layer=None, dtype_per_channel=None, num_channels=None, **kwargs):
	if "dtype" in kwargs.keys():
		dtype_per_layer = kwargs["dtype"]
	...
```
... or would you rename `dtype_per_channel` to `dtype`?

Also, should I move `dtype_per_channel` behind `num_channels` so that `num_channels` is still the fourth parameter?

philippotto:
> Also, should I move `dtype_per_channel` behind `num_channels` so that `num_channels` is still the fourth parameter?

Hm, this is a good point. Adding an alias to stay backwards-compatible and then changing the parameter order is not really consistent :D However, we should thrive for a clean parameter in my opinion.
Then, it might make sense to simply make a hard change and bump the major version of wkcuber. When updating the lib, we should pay attention that this interface is used correctly. Also, the release note should contain a comment about this. Additionally, we could remove `dtype` completely and say that the clearer `dtype_per_channel` or `dtype_per_layer` should be used instead. 

@normanrz What do you think about making a major bump to wkcuber? Last time you were against it, if I remember correctly. Not sure why though. Seems like a good way to indicate the breaking changes.

wkcuber/api/Dataset.py
philippotto: > An alternative to these lines would be to wrap a try-catch block around it and simply try to convert `dtype_per_layer` to a valid dtype. This would make the assertion that there is a digit in the dtype redundant (see `convert_dtypes`) because `str(np.dtype("int"))` results in `int64`.

Sounds good!

normanrz:
> @normanrz What do you think about making a major bump to wkcuber? Last time you were against it, if I remember correctly. Not sure why though. Seems like a good way to indicate the breaking changes.

We are still on a 0.x.x version which accepts breaking changes. I think you can do the breaking change and change all occurrences in our different projects.

rschwanhold:
Should I use the opportunity and also resolve #229, so that we do not need to make another breaking change in the near future?

philippotto:
> Should I use the opportunity and also resolve #229, so that we do not need to make another breaking change in the near future?

Yes, that would be awesome! Please include this change in the PR title, too, then, since it's quite important and should show up in the commit message, once this PR is squashed and merged.

philippotto (approved):
Great :) After merging this PR, please release a new version and integrate it into our other project(s) :)

philippotto:
> Also, we need to think about API backwards compatibility. One could keep the dtype argument and alias it to dtype_per_channel and optionally emit a warning that one should use the clearer/newer argument name dtype_per_channel directly. I think, that could be a good migration path. Do you agree?

@rschwanhold If I'm not mistaken, this slipped through the cracks. Can you open another PR to fix this? Simply asserting that `dtype` is not passed (with a hint about the new options), should be enough.

Issue #201, opened 2020-04-20
api: Support adding a layer via symlink


philippotto:
Use case: I want to add a layer to a DS which already exists somewhere else. When done manually, one would add a symlink into the DS folder which points to the desired layer. Afterwards, one has to manipulate the datasource-properties.json file. If the DS api could do this automatically, this would be awesome.

To get the necessary properties, one could look for a `datasource-properties.json` file in `..` (relative to the symlink target).

PR #240, opened 2020-09-22
Support adding a layer via symlink

rschwanhold:
- add method to the `Dataset` class for adding layers of foreign datasets via a symlink
- fixes: #201

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1284,7 +1284,11 @@ def test_add_symlink_layer():
     delete_dir("./testoutput/simple_wk_dataset_copy")
     copytree("./testdata/simple_wk_dataset/", "./testoutput/simple_wk_dataset_copy/")
 
-    original_mag = WKDataset("./testoutput/simple_wk_dataset_copy/").get_layer("color").get_mag("1")
+    original_mag = (
+        WKDataset("./testoutput/simple_wk_dataset_copy/")
+        .get_layer("color")
+        .get_mag("1")
+    )
 
     ds = WKDataset.create("./testoutput/wk_dataset_with_symlink", scale=(1, 1, 1))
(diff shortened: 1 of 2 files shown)


rschwanhold:
The properties of the layer are copied to the `datasource-properties.json` of the new dataset. 
Example:
`ds1` is the original dataset with a layer `color`.
`ds2` has a symlink to the `color` layer of `ds1`
If you write data to `color` through `ds2`, then only the bounding box in the properties of `ds2` will be updated.
Is this a problem?

philippotto:
> The properties of the layer are copied to the `datasource-properties.json` of the new dataset.
> Example:
> `ds1` is the original dataset with a layer `color`.
> `ds2` has a symlink to the `color` layer of `ds1`
> If you write data to `color` through `ds2`, then only the bounding box in the properties of `ds2` will be updated.
> Is this a problem?

As I see it, there are three options:
- Try to be clever and update the `datasource-properties.json` in both datasets --> However, this won't work, if one mutates the original dataset (because that dataset doesn't know it's being linked to).
- Try to forbid mutating a symlinked layer --> However, this also does not cover the case where the original dataset is mutated.
- Don't do anything and let the user be responsible for what they are doing.

Since option 1 and 2 are not perfect, I'd simply go with option 3. Until now, we also simply symlink layers and don't really care about the problem of keeping the bounding box consistent. Since we usually don't mutate the symlinked layer, anyway, this should be fine in my opinion.

However, maybe somebody also can also weigh in with their opinion? For example, @jstriebel ?

philippotto (commented):
Cool PR! Code looks good and :+1: for the test. Let's wait for another opinion regarding your last question, though.

jstriebel:
Agree with @philippotto :+1: Leaving this to the user would probably be fine!

philippotto (approved):
Nice one :sunglasses:

Issue #209, opened 2020-05-07
Allow to write a subset of a compressed wkw dataset


jstriebel:
It would be great if one could write a subset of a compressed dataset in the API. This could be achieved by first reading the whole file, then overwriting the respective subset and the writing the whole file.

PR #241, opened 2020-09-23
Allow to write a subset of a compressed wkw dataset

rschwanhold:
- fixes: #209

### DIFF
--- a/wkcuber/api/View.py
+++ b/wkcuber/api/View.py
@@ -48,9 +48,7 @@ class View:
         )
 
         if self._is_compressed() and allow_compressed_write:
-            absolute_offset, data = self._handle_compressed_write(
-                absolute_offset, data
-            )
+            absolute_offset, data = self._handle_compressed_write(absolute_offset, data)
 
         if not was_opened:
             self.open()
--- a/wkcuber/api/View.py
+++ b/wkcuber/api/View.py
@@ -119,7 +119,7 @@ class View:
             relative_offset = np.array(chunk.topleft) - np.array(self.global_offset)
             view = self.get_view(size=chunk.size, relative_offset=relative_offset)
             view.is_bounded = True
-            job_args.append((view, job_args_per_chunk,))
+            job_args.append((view, job_args_per_chunk))
 
         # execute the work for each chunk
         wait_and_ensure_success(executor.map_to_futures(work_on_chunk, job_args))
--- a/wkcuber/api/View.py
+++ b/wkcuber/api/View.py
@@ -1,7 +1,7 @@
 import math
 
 import numpy as np
-from wkw import Dataset
+from wkw import Dataset, wkw
 
 from wkcuber.api.TiffData.TiffMag import TiffMag
 from wkcuber.api.bounding_box import BoundingBox
@@ -46,6 +46,8 @@ class View:
         absolute_offset = tuple(
             sum(x) for x in zip(self.global_offset, relative_offset)
         )
+
+        absolute_offset, data = self._handle_compressed_write(absolute_offset, data)
 
         if not was_opened:
             self.open()
@@ -120,6 +122,9 @@ class View:
     def _check_chunk_size(self, chunk_size):
         raise NotImplementedError
 
+    def _handle_compressed_write(self, absolute_offset, data):
+        return absolute_offset, data
+
     def __enter__(self):
         return self
 
@@ -155,6 +160,40 @@ class WKView(View):
             raise AssertionError(
                 f"The passed parameter 'chunk_size' {chunk_size} must be a multiple of (32, 32, 32)."
             )
+
+    def _handle_compressed_write(self, absolute_offset, data):
+        if (
+            self.header.block_type == wkw.Header.BLOCK_TYPE_LZ4
+            or self.header.block_type == wkw.Header.BLOCK_TYPE_LZ4HC
+        ):
+            # data is compressed
+
+            # calculate aligned bounding box
+            file_bb = np.full(3, self.header.file_len * self.header.block_len)
+            margin_to_top_left = np.array(absolute_offset) % file_bb
+            aligned_offset = np.array(absolute_offset) - margin_to_top_left
+            bottom_right = np.array(absolute_offset) + np.array(list(data.shape)[-3:])
+            margin_to_bottom_right = file_bb - (bottom_right % file_bb)
+            aligned_bottom_right = bottom_right + margin_to_bottom_right
+            aligned_shape = aligned_bottom_right - aligned_offset
+
+            if tuple(aligned_offset) != tuple(absolute_offset) or tuple(
+                aligned_shape
+            ) != tuple(list(data.shape)[-3:]):
+                # the data is not aligned
+                # read the aligned bounding box
+                aligned_data = self.read(offset=aligned_offset, size=aligned_shape)
+                index_slice = [slice(None, None)] + [
+                    slice(start, end)
+                    for start, end in zip(
+                        margin_to_top_left, bottom_right - aligned_offset
+                    )
+                ]
+                # overwrite the specified data
+                aligned_data[tuple(index_slice)] = data
+                return tuple(aligned_offset), aligned_data
+
+        return absolute_offset, data
 
 
 class TiffView(View):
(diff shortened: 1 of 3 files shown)


wkcuber/api/View.py
```diff
@@ -120,6 +122,9 @@ def for_each_chunk(self, work_on_chunk, job_args_per_chunk, chunk_size, executor
     def _check_chunk_size(self, chunk_size):
         raise NotImplementedError
 
+    def _handle_compressed_write(self, absolute_offset, data):
+        return absolute_offset, data
```
rschwanhold, line 126: The Dataset API does not support compressed tiffs yet (as far as I know). Therefore, I only overwrite this method for the `WKView`

wkcuber/api/View.py
```diff
@@ -156,6 +161,40 @@ def _check_chunk_size(self, chunk_size):
...
+            margin_to_bottom_right = file_bb - (bottom_right % file_bb)
+            aligned_bottom_right = bottom_right + margin_to_bottom_right
+            aligned_shape = aligned_bottom_right - aligned_offset
+
+            if tuple(aligned_offset) != tuple(absolute_offset) or tuple(
+                aligned_shape
+            ) != tuple(list(data.shape)[-3:]):
+                # the data is not aligned
+                # read the aligned bounding box
+                aligned_data = self.read(offset=aligned_offset, size=aligned_shape)
```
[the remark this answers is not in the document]
jstriebel: I was also wondering about this, but I think it's fine to leave this to the user.

wkcuber/api/View.py
```diff
@@ -156,6 +161,40 @@ def _check_chunk_size(self, chunk_size):
...
+            margin_to_top_left = np.array(absolute_offset) % file_bb
+            aligned_offset = np.array(absolute_offset) - margin_to_top_left
+            bottom_right = np.array(absolute_offset) + np.array(list(data.shape)[-3:])
+            margin_to_bottom_right = file_bb - (bottom_right % file_bb)
+            aligned_bottom_right = bottom_right + margin_to_bottom_right
+            aligned_shape = aligned_bottom_right - aligned_offset
+
+            if tuple(aligned_offset) != tuple(absolute_offset) or tuple(
+                aligned_shape
+            ) != tuple(list(data.shape)[-3:]):
```
jstriebel, line 182: All of those are already tuples.

```suggestion
            if aligned_offset != absolute_offset or aligned_shape != data.shape[-3:]:
```

wkcuber/api/View.py
```diff
@@ -156,6 +161,40 @@ def _check_chunk_size(self, chunk_size):
...
+            self.header.block_type == wkw.Header.BLOCK_TYPE_LZ4
+            or self.header.block_type == wkw.Header.BLOCK_TYPE_LZ4HC
+        ):
+            # data is compressed
+
+            # calculate aligned bounding box
+            file_bb = np.full(3, self.header.file_len * self.header.block_len)
+            margin_to_top_left = np.array(absolute_offset) % file_bb
+            aligned_offset = np.array(absolute_offset) - margin_to_top_left
+            bottom_right = np.array(absolute_offset) + np.array(list(data.shape)[-3:])
```
jstriebel, line 175:
```suggestion
            absolute_offset_np = np.array(absolute_offset)
            margin_to_top_left = absolute_offset_np % file_bb
            aligned_offset = absolute_offset_np - margin_to_top_left
            bottom_right = absolute_offset_np + np.array(data.shape[-3:])
```

wkcuber/api/View.py
```diff
@@ -156,6 +161,40 @@ def _check_chunk_size(self, chunk_size):
...
+                # read the aligned bounding box
+                aligned_data = self.read(offset=aligned_offset, size=aligned_shape)
+                index_slice = [slice(None, None)] + [
+                    slice(start, end)
+                    for start, end in zip(
+                        margin_to_top_left, bottom_right - aligned_offset
+                    )
+                ]
+                # overwrite the specified data
+                aligned_data[tuple(index_slice)] = data
```
jstriebel, line 193:
```suggestion
                index_slice = (slice(None, None), *(
                    slice(start, end)
                    for start, end in zip(
                        margin_to_top_left, bottom_right - aligned_offset
                    )
                ))
                # overwrite the specified data
                aligned_data[index_slice] = data
```

wkcuber/api/View.py
```diff
@@ -47,6 +47,8 @@ def write(self, data, relative_offset=(0, 0, 0)):
             sum(x) for x in zip(self.global_offset, relative_offset)
         )
 
+        absolute_offset, data = self._handle_compressed_write(absolute_offset, data)
```
jstriebel, line 50: Very elegant solution :+1:

wkcuber/api/View.py
jstriebel, line 185: This also works, when reading and writing multiple chunks, right? They only have to be aligned, I guess.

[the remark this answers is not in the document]
jstriebel: Do you have an idea on how to solve this with a warning elegantly and without a lot of overhead? Maybe a small cache about the last write could already be enough.

jstriebel (commented):
Nice work! And great tests!
See my comments below for the nitpicks :nerd_face:

wkcuber/api/View.py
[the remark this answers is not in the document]
philippotto: What do you think about adding another parameter to the write function? For example, `write(..., allow_compressed_write=True)`. By default this parameter would be `False`  (and if the target is compressed throw an exception which tells the user to either align the input data or use that parameter with the note of the performance hit).
I think, making this opt-in would be reasonable, since it should only be done when one is sure that the performance hit is not critical. Emitting a warning won't work, in my opinion, since it won't be read mostly.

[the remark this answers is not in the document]
jstriebel: Sounds great!

rschwanhold: Very good point. It failed if the view (and therefore also the chunks) had the parameter `is_bounded` set to `True`. 
If the bounding box of the chunks was not aligned, `read(offset=aligned_offset, size=aligned_shape)` would try to access data outside of the bounding box (which is not allowed if `is_bounded=True`).

rschwanhold: I think we could run into concurrency problems: `View.for_each_chunk(...)` executes a function for each chunk. Let's say that the chunks are not aligned with the compressed files. Now each chunk reads the compressed file (not only its own bounding box) and then writes the compressed data.
Maybe we shouldn't allow this for chunks?!

wkcuber/api/View.py
```diff
@@ -47,7 +47,15 @@ def write(self, data, relative_offset=(0, 0, 0)):
...
-        absolute_offset, data = self._handle_compressed_write(absolute_offset, data)
+        if self._is_compressed():
+            if allow_compressed_write:
+                absolute_offset, data = self._handle_compressed_write(
+                    absolute_offset, data
+                )
+            else:
+                raise ValueError(
+                    "Cannot write unaligned compressed data. Align the data first or use the parameter 'allow_compressed_write', which might impact the performance."
+                )
```
jstriebel, line 58: Here we don't know yet if the data is unaligned, right? I think we should do nothing in this case

wkcuber/api/View.py
```diff
@@ -161,39 +169,50 @@ def _check_chunk_size(self, chunk_size):
...
+            # the data is not aligned
+            # read the aligned bounding box
+            try:
                 aligned_data = self.read(offset=aligned_offset, size=aligned_shape)
-                index_slice = [slice(None, None)] + [
+            except AssertionError as e:
+                raise AssertionError(
+                    f"Writing compressed data failed. The compressed file is not fully inside the bounding box of the view (offset={self.global_offset}, size={self.size}). "
+                    + str(e)
+                )
```
jstriebel, line 201: :+1:

wkcuber/api/View.py
jstriebel, line 58:
```suggestion
        if self._is_compressed() and allow_compressed_write:
            absolute_offset, data = self._handle_compressed_write(
                absolute_offset, data
            )
```

jstriebel (approved):
Nice, LGTM :tada:

PR #243, opened 2020-10-16
Dataset API: search for datasets also in the path with the long mag name

rschwanhold:
Problem:
- There is a dataset with Mag("2")
- The dataset API would only search on disk for a directory called `.../2` but not for `.../2-2-2`

Solution:
- The dataset API first checks if there is a directory named `.../2`. If not, `.../2-2-2` is also checked

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1,7 +1,7 @@
 import filecmp
 import json
 import os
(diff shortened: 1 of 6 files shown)


normanrz (approved):
lgtm!

PR #246, opened 2020-11-02
Forbid usage of the deprecated parameter 'dtype' for the Dataset API

rschwanhold:
In #239 the parameter `dtype` was converted into `dtype_per_channel` and `dtype_per_layer`. This PR asserts that the old parameter `dtype` is no longer used.


philippotto (approved):
Excellent! Also :+1: for the test :tada:

PR #249, opened 2020-11-16
Use mypy for type checking

### LINKED ISSUE #244: Use MyPy
To improve the static robustness of this library (and its usages), we should integrate MyPy.

rschwanhold:
## Description
fixes #244 

## Todo
- [x] Add typechecking to CI

### DIFF
--- a/wkcuber/api/Properties/LayerProperties.py
+++ b/wkcuber/api/Properties/LayerProperties.py
@@ -1,12 +1,15 @@
 from os.path import join, dirname, isfile
+from pathlib import Path
+from typing import Tuple, Type, Union, Any, Dict, List, Optional, cast
 
 from wkw import wkw
 
+from wkcuber.api.Properties.ResolutionProperties import Resolution
 from wkcuber.mag import Mag
 from wkcuber.api.bounding_box import BoundingBox
 
 
-def extract_num_channels(num_channels_in_properties, path, layer, mag):
+def extract_num_channels(num_channels_in_properties: Optional[int], path: Union[str, Path], layer: str, mag: Optional[Dict[str, int]]) -> int:
     # if a wk dataset is not created with this API, then it most likely doesn't have the attribute 'num_channels' in the
     # datasource-properties.json. In this case we need to extract the 'num_channels' from the 'header.wkw'.
     if num_channels_in_properties is not None:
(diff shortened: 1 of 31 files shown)


wkcuber/convert_nifti.py
```diff
@@ -111,18 +113,17 @@ def to_target_datatype(
...
-    layer_name,
-    dtype,
-    scale,
-    mag=1,
-    file_len=DEFAULT_WKW_FILE_LEN,
-    bbox_to_enforce=None,
-    write_tiff=False,
-    use_orientation_header=False,
-    flip_axes=None,
-):
```
rschwanhold, line 125: The parameter `mag` is never used (it gets overwritten). Therefore, I removed the parameter. Is that correct?

wkcuber/api/Properties/LayerProperties.py
```diff
@@ -161,23 +176,23 @@ def bounding_box(self) -> dict:
...
-        mappings=None,
-    ):
+        name: str,
+        category: str,
+        element_class: str,
+        data_format: str,
+        num_channels: int,
+        bounding_box: Dict[str, Union[int, Tuple[int, int, int]]] = None,
+        resolutions: List[Resolution] = None,
+        largest_segment_id: int = None,
```
rschwanhold, line 193: I think `largest_segment_id` is not optional for segmentation layers, but making it non-optional would require a different order of the parameters and then the order of `SegmentationLayerProperties` would be different from `LayerProperties`. 
(There is an assertion in DatasetProperties._add_layer for this)
Therefore, I would leave it this way.

rschwanhold (commented):
I had do add a few exceptions to handle the `None` case of `Optional`. Let me know if there is a better way to do this

jstriebel (approved):
Nice! I did some spot-checks, they LGTM :+1: I guess `mypy` does the rest ;-) It's awesome that you already checked this with the current mypy-repo, so we don't expect any new errors when doing https://github.com/scalableminds/voxelytics/issues/1982 here, too.

wkcuber/convert_nifti.py
jstriebel: Seems fine for me, let's just check who might call this in any dependent projects.

wkcuber/api/Properties/LayerProperties.py
jstriebel: Yes, I think you can leave it like this or have a default of `-1`, if that is useful? If you want to leave the signature as is, you can `assert largest_segment_id is not None` in the `__init__()` to mark the attribute as non-optional.

jstriebel:
> I had do add a few exceptions to handle the None case of Optional. Let me know if there is a better way to do this.

We often use assertions in voxelytics for that matter, e.g. `assert x is not None, "x must be set to an integer in case ABC"`.

philippotto:
Are there any blockers for this PR or can it be merged soon? :)

rschwanhold:
This will be merged today

Issue #251, opened 2020-12-03
Auto-detect what conversion to run


normanrz:
When running the `wkcuber` on a folder or file, the cuber should automatically detect what conversion it should run:

* Convert an image sequence
* Convert a single multi-section image (see #221, #63)
   * Note: Single tiff files can also have multiple channels 
* Do nothing if already WKW

We'll probably have more conversion runs in the future, so this should be extendable.

For this, it needs to recursively check the folders. For example a user might point to folder `foo`, which contains folder `bar` that contains a sequence of images:

```
- foo
   - bar
      - 0001.tif
      - 0002.tif
      ...
```

PR #252, opened 2020-12-03
read multi-page tiffs

fm3:
work-in-progress pr to support reading multi-page tiff images


.github/workflows/main.yml
```diff
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7]
+        python-version: [3.7]
```
normanrz, line 11: What would be required to also support 3.6 and 3.8?

wkcuber/cubing.py
```diff
@@ -157,7 +157,7 @@ def cubing_job(args):
                 for z, file_name in zip(z_batch, source_file_batch):
                     # Image shape will be (x, y, channel_count, z=1)
```
rschwanhold, line 158: Is this comment still up to date?

wkcuber/cubing.py
```diff
@@ -210,13 +210,19 @@ def cubing_job(args):
...
 
 def cubing(source_path, target_path, layer_name, dtype, batch_size, args) -> dict:
-
     source_files = find_source_filenames(source_path)
 
     # All images are assumed to have equal dimensions
     num_x, num_y = image_reader.read_dimensions(source_files[0])
     num_channels = image_reader.read_channel_count(source_files[0])
-    num_z = len(source_files)
+    num_z_slices_per_file = image_reader.read_z_slices_per_file(source_files[0])
```
rschwanhold, line 218: Does each file have the same number of z_slices or is there only a single file with many z_slices?

wkcuber/cubing.py
```diff
@@ -157,7 +157,7 @@ def cubing_job(args):
                 for z, file_name in zip(z_batch, source_file_batch):
                     # Image shape will be (x, y, channel_count, z=1)
                     image = read_image_file(
-                        file_name, target_wkw_info.header.voxel_type
+                        file_name, target_wkw_info.header.voxel_type, z
```
rschwanhold, line 160: Is my understanding correct that, in case of non-multi-paged tiffs, the `z` is simply ignored because the `z` is also encoded in the filename?

rschwanhold (commented):
I have added a few questions

.github/workflows/main.yml
youri-k: As far as I can tell, Python 3.8 just works. However, Python 3.6 is not supported by `tifffile` anymore.

wkcuber/cubing.py
youri-k: As mentioned in the comment above, it is assumed that all images have equal dimensions, so I assume that for the number of z slices as well.

wkcuber/cubing.py
youri-k: Correct

.github/workflows/main.yml
normanrz: Then, we should check with the downstream projects (mostly voxelytics) that 3.6 will not be used anymore.

normanrz: And add 3.8 to the CI for testing.

wkcuber/image_readers.py
```diff
@@ -107,23 +117,54 @@ def read_channel_count(self, _file_name):
         return 1
 
 
-class ImageReader:
+class TiffImageReader(ImageReader):
+    def read_array(self, file_name, dtype, z_slice):
+        tif_file = TiffFile(file_name)
```
normanrz, line 122: In a recent script that I wrote, I had to transpose the image from `tifffile` because it uses inverted dimensions as compared to wkw. I can provide a test file.

normanrz:
@youri-k is this ready for review?

youri-k:
Sadly, it's not quite ready since it only reads the first channel of data. I think a better channel handling belongs to #251, so I will change this to always read all channels of data. Than it should be good to go

youri-k:
@normanrz @rschwanhold This should be good to go now :slightly_smiling_face:

normanrz (approved):
lgtm :+1:

wkcuber/image_readers.py
```diff
@@ -107,25 +123,83 @@ def read_channel_count(self, _file_name: str) -> int:
...
+    if index == -1:
+        return 1
+    else:
+        return tif_series.shape[index]  # pylint: disable=unsubscriptable-object
+
+
+class TiffImageReader(ImageReader):
+    def read_array(self, file_name: str, dtype: np.dtype, z_slice: int) -> np.ndarray:
+        with TiffFile(file_name) as tif_file:
+            num_channels = find_count_of_axis(tif_file, "C")
```
rschwanhold, line 139: Here you can use `read_channel_count(tif_file)`

poetry.lock
```diff
@@ -573,116 +560,130 @@ examples = ["matplotlib (>=2.1.1)", "scikit-image (>=0.13)", "pandas (>=0.25.0)"
...
 
 [package.extras]
 docs = ["sphinx", "jaraco.packaging (>=3.2)", "rst.linker (>=1.9)"]
-testing = ["pytest (>=3.5,<3.7.3 || >3.7.3)", "pytest-checkdocs (>=1.2.3)", "pytest-flake8", "pytest-cov", "jaraco.test (>=3.2.0)", "jaraco.itertools", "func-timeout", "pytest-black (>=0.3.7)", "pytest-mypy"]
+testing = ["pytest (>=3.5,!=3.7.3)", "pytest-checkdocs (>=1.2.3)", "pytest-flake8", "pytest-cov", "jaraco.test (>=3.2.0)", "jaraco.itertools", "func-timeout", "pytest-black (>=0.3.7)", "pytest-mypy"]
 
 [metadata]
-content-hash = "aca3ac4501ff2f0da5776d539e91b6912a3b8f01a936bbeeae48147af7695fcd"
-python-versions = "^3.6"
+lock-version = "1.1"
```
rschwanhold, line 684: A lot of the lock file changed, even though you only edited a few entries in the `pyproject.toml`. I had the same issue with my last PR. The reason for this is that you are probably using poetry version `1.1.x` (instead of `1.0.x`). The version of poetry is not explicitly specified in the Readme. You might want to update the Readme to use the correct poetry version.

rschwanhold (approved):
I have added one tiny suggestion. Other than that, it looks good.

philippotto:
> work-in-progress pr to support reading multi-page tiff images

I know that the PR was already merged, but could you update the PR description a bit, @youri-k. Would be interesting how exactly the multi-page tiff support works/behaves. Just wanted to create a new release, but I don't really know how to describe the changes from this PR :)

youri-k:
@philippotto I'm sorry I missed that again. I hope my explanation is understandable, but please don't hesitate to ask for more details.

Issue #187, opened 2020-03-03
Follow-up to high-level dataset API: support conversion to from one dataset type to another


rschwanhold:
For example, convert a `WKDataset` to a `TiffDataset`.  
There are already method-frames for this conversion, however, they are not implemented yet.

PR #259, opened 2020-12-18
Downsampling with the dataset api

rschwanhold:
fixes: #186 
fixes: #187


wkcuber/api/Layer.py
```diff
@@ -88,6 +105,74 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]):
...
+            from_mag: Mag,
+            max_mag: Mag,
+            interpolation_mode: str,
+            compress: bool,
+            scale: Tuple[float, float, float] = None,
+            buffer_edge_len=None,
+            args: Namespace = None,
+    ):
+        # if 'scale' is set, the data gets downsampled anisotropic
+        interpolation_mode = parse_interpolation_mode(interpolation_mode, self.name)
```
rschwanhold, line 125: I copied this from the old implementation, but I would assume that the second parameter should be the type of the layer instead of the name, right?

wkcuber/api/Layer.py
```diff
@@ -88,6 +105,74 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]):
...
+                    prev_mag_ds.get_view(),  # the offset and size of the view need to be set from inside the job
+                    mag_factors,
+                    interpolation_mode,
+                    buffer_edge_len,
+                    compress,
+                    target_mag_ds._get_file_dimensions(),
+                    job_count_per_log,
+                    target_mag_view.global_offset,
+                    num_chunks_per_dim,
+                    use_logging,
```
rschwanhold, line 161: In the old implementation, an counter `i` was used to determine which chunk should use logging. I am now using `.for_each_chunk(...)`, which passes the same arguments to all chunks. This is my attempt to get logging to work: a function `use_logging` is passed with which each chunks can calculate the counter `i` artificially and therefore know whether it should log or not. However the executor requires this function to be static (I think). Therefore I also have to pass a few additional variables to each chunk, so that they can call this function in a static way. This is a bit ugly.
What would be a better way to do this?

wkcuber/api/Layer.py
```diff
@@ -151,6 +236,10 @@ def setup_mag(self, mag):
             self.name, mag, wk_header.block_len * wk_header.file_len
         )
 
+    def _initialize_mag_from_other_mag(self, new_mag_name, other_mag, compress):
+        block_type = wkw.Header.BLOCK_TYPE_LZ4HC if compress else wkw.Header.BLOCK_TYPE_RAW
+        return self.add_mag(new_mag_name, other_mag.block_len, other_mag.file_len, block_type)
```
rschwanhold, line 241: The new implementation uses the same `block_len` as the source magnification. In the previous implementation this always defaults to 32

wkcuber/api/Layer.py
```diff
@@ -88,6 +96,131 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]):
...
+                    target_mag_ds._get_file_dimensions(),
+                    job_count_per_log,
+                )
+                prev_mag_ds.get_view().for_zipped_chunks(
+                    # this view is restricted to the bounding box specified in the properties
+                    downsample_cube_job,
+                    target_view=target_mag_view,
+                    job_args_per_chunk=job_args,
+                    source_chunk_size=np.array(target_mag_ds._get_file_dimensions()) * mag_factors,
+                    target_chunk_size=target_mag_ds._get_file_dimensions(),
```
rschwanhold, line 215: The idea is to use the file size of the target mag as a reference because concurrent writes need to be avoided. However, multiple of this would also be possible. For example the file dimension of a tiff is only (x, y, 1).

wkcuber/api/Dataset.py
```diff
@@ -286,6 +294,47 @@ def get_view(self, layer_name, mag, size, offset=None, is_bounded=True) -> View:
     def _create_layer(self, layer_name, dtype_per_channel, num_channels) -> Layer:
         raise NotImplementedError
 
+    def copy_dataset(self, empty_target_ds: "AbstractDataset", args=None):
```
rschwanhold, line 297: This has no direct connection with "downsampling", but I wrote this to convert a WKDataset to TiffDataset. I needed this for testing the new downsampling code.

wkcuber/api/Layer.py
```diff
@@ -99,20 +118,157 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]) -> None:
...
     mags: Dict[str, WKMagDataset]
 
-    def add_mag(
-        self,
-        mag: Union[str, Mag],
-        block_len: int = None,
-        file_len: int = None,
-        block_type: int = None,
-    ) -> MagDataset:
+    def add_mag(self, mag: Union[str, Mag], **kwargs: Any) -> MagDataset:
```
rschwanhold, line 267: Unfortunately, I had to change the signature of the method to add this method to the base class. This is necessary to pass the typecheck. I do not understand why this was not a problem in #249

wkcuber/api/Layer.py
```diff
@@ -99,20 +118,157 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]) -> None:
...
+            for i in range(0, 3):
+                # pad left / top / front
+                buffer_width = (np.array(current_offset) - aligned_offset)[i]
+                if buffer_width > 0:
+                    padding_shape = shape
+                    padding_shape[i + 1] = buffer_width
+                    mag.write(
+                        data=np.zeros(padding_shape, dtype=mag.get_dtype()),
+                        offset=aligned_offset,
+                    )
```
philippotto, line 186: I assume this only writes zeros, so that the bounding box is enlarged as a side-effect? Two questions:

- can we avoid writing zeros? (a wkw should return zeros anyway if it the target area is empty)
- should this be undone after the downsampling is finished so that the bounding box stays tight?

wkcuber/api/Layer.py
```diff
@@ -99,20 +118,157 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]) -> None:
...
+            // -all_mags_after_downsampling[-1].as_np()
+        )  # ceil div
+        # translate alignment into mag 1
+        aligned_offset_in_mag1 = lowest_mag * offset_in_lowest_mag
+        aligned_size_in_mag1 = lowest_mag * (
+            end_offset_in_lowest_mag - offset_in_lowest_mag
+        )
+
+        # pad the existing mags
+        for mag_name in existing_mags:
```
philippotto, line 168: It would be cool if the padding-related code was extracted in a separate function and unit-tested. Do you think that's possible without too much effort?

philippotto (commented):
Excellent work :+1: I left a couple of comments and questions, but the overall picture is looking very solid :tada:

wkcuber/api/Dataset.py
```diff
@@ -76,6 +77,16 @@ def dtype_per_channel_to_dtype_per_layer(
     )
 
 
+def copy_job(args: Tuple[View, View, int, Any]) -> None:
+    # Copy the data form one view to the other in a buffered fashion
+    (source_view, target_view, i, _) = args
+
+    data = source_view.read()
+    if len(data.shape) == 4 and data.shape[0] == 1:
+        data = data[0]  # remove channel dimension
```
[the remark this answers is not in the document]
rschwanhold: You mean the caller of the `write` method, right? I could move this check into the `write` method.

wkcuber/api/Dataset.py
```diff
@@ -307,6 +318,66 @@ def _create_layer(
     ) -> Layer:
         raise NotImplementedError
 
+    def copy_dataset(
+        self, empty_target_ds: "AbstractDataset", args: Namespace = None
```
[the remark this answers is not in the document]
rschwanhold: I have added the `Optional` now because it is semantically more correct, but I think mypy automatically understands that parameters with a default value of `None` are optional (at least my IDE understands it)

wkcuber/downsampling_utils.py
```diff
@@ -0,0 +1,351 @@
+import logging
```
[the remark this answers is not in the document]
rschwanhold: Not quite. Most of the code comes from `downsampling.py`. However, a few methods are modified:
- `determine_buffer_edge_len(...)` # modified
- `get_next_mag(...)` # modified (this were two separate functions)
- `downsample_cube_job(...)` # modified (here are most of the changes)

wkcuber/api/Layer.py
```diff
@@ -99,20 +118,157 @@ def set_bounding_box_size(self, size: Tuple[int, int, int]) -> None:
...
+        self,
+        from_mag: Mag,
+        max_mag: Mag,
+        interpolation_mode: str,
+        compress: bool,
+        scale: Tuple[float, float, float] = None,
+        buffer_edge_len: int = None,
+        args: Namespace = None,
+    ) -> None:
+        # if 'scale' is set, the data gets downsampled anisotropic
```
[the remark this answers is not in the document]
rschwanhold: What do you mean with "scale of the layer"?
Let say the layer has 2 existing magnifications ("1", "2"). The user can now downsample the data anisotropic (`from_mag="2"`, ..., `scale="1-1-2"`). This would create `4-4-2`-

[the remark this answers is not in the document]
philippotto: I think, the `scale` argument here is the scale which is persisted in the datasource-properties.json. So, for example it could be `[11.24, 11.24, 24]` or something equally odd/fractional as it's referring to the nm size of a voxel. Since this is stored in the dataset's properties, the layer could probably look this up easily via its dataset.

Speaking closer to the code: Have a look at the following in downsampling.py:

```python
    if anisotropic:
        if scale is None:
            try:
                scale = read_datasource_properties(path)["scale"]
            except Exception as exc:
                logging.error(
                    "Could not get the scale from the datasource-properties.json. Probably your path is wrong. "
                    "If you do not provide the layer_name or from_mag, they need to be included in the path."
                    "(e.g. dataset/color/1). Otherwise the path should just point at the dataset directory."
                    "the path: %s",
                    path,
                )

    WKDataset(path).get_layer(layer_name).downsample(
        from_mag=from_mag,
        max_mag=max_mag,
        interpolation_mode=interpolation_mode,
        compress=compress,
        scale=scale,
        buffer_edge_len=buffer_edge_len,
        args=args,
    )
```

Instead of extracting the scale from the outside, the actual `downsample` method could do so. It would only need the boolean information `anisotropic` (or `use_scale`; I don't have a strong opinion here). Otherwise, if some other code wants to downsample a layer anisotropically, they would need to look up the scale manually.

wkcuber/api/Dataset.py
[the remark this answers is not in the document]
philippotto: Exactly! Sounds like a good solution.

wkcuber/downsampling_utils.py
```diff
@@ -0,0 +1,351 @@
...
+        chunk_xzy_idx[2] * num_chunks_per_dim[0] * num_chunks_per_dim[1]
+        + chunk_xzy_idx[1] * num_chunks_per_dim[1]
+        + chunk_xzy_idx[0]
+    )
+
+    return i % job_count_per_log == 0
+
+
+def determine_buffer_edge_len(dataset: wkw.Dataset) -> int:
+    if hasattr(dataset.header, "file_len") and hasattr(dataset.header, "block_len"):
```
philippotto, line 48: This probably cannot happen, but conceptionally only one of these attributes could be set (e.g., block_len == 1). file_len would still default to 32 in which case the effective cube_length will be 32. However, the current code wouldn't take that length into account (since the header would need to have both attributes defined) and simply return the default edge length.

From my understanding, the header always has both properties defined (in the wkw library, they simply default to 32). So, the actual question is probably: why do we need the `hasattr` calls here? If it's only to satisfy mypy, we could replace this with `assert hasattr(dataset.header,  "file_len") and ...`.

wkcuber/api/Layer.py
philippotto: Yes, you are completely right. Good catch! Seems like mode-filtering was used for all layers which weren't named `color` in the past :see_no_evil:

wkcuber/downsampling_utils.py
rschwanhold: I haven't push the latest commit. The type of dataset is actually `Union[wkw.Dataset, TiffMag]`. `TiffMag` does not have those attributes

philippotto: Ah, I see. Can you do an is-instance-of check and differentiate via an if-else between the two cases? Might be a tiny bit cleaner.

wkcuber/api/Layer.py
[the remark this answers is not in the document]
rschwanhold: Ok, very interesting. I like your idea and will implement it. I just always assumed that you could specify any scale as parameter (independent of the the scale in the properties)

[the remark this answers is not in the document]
philippotto: > I just always assumed that you could specify any scale as parameter (independent of the the scale in the properties)

Yes, this might have been the case, but I'd argue that this wasn't really intended during the design of the old interface. Instead, it can be quite error prone. If one wants to create a custom mag hierarchy, one has to do this manually in my opinion.

wkcuber/api/Layer.py
rschwanhold: 1. I haven't tested this, but I assume that it would be sufficient for `WKDataset` to update the boundingbox (without padding the data) (because, as you pointed out, wkw handles this). ~~ I think this would not work for `TiffDataset`.~~ The `TiffDataset` does not add padding during reading, but during writing.
2. This cannot be undone. The amount of padding is dependent on the lowest magnification (so that for every two adjacent magnifications the ratio of their offsets and sizes is equal to `scale`). Resetting the boundingbox would result in rounding problems in lower mags

wkcuber/api/View.py
```diff
@@ -158,6 +158,79 @@ def for_each_chunk(
...
+        source_chunks = BoundingBox(source_offset, self.size).chunk(
+            source_chunk_size_np, list(source_chunk_size_np)
+        )
+        target_chunks = BoundingBox(target_offset, calculated_target_size).chunk(
+            target_chunk_size, list(target_chunk_size)
+        )
+
+        for i, (source_chunk, target_chunk) in enumerate(
+            zip(source_chunks, target_chunks)
+        ):
```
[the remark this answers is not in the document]
rschwanhold: I agree with you that the current implementation is not optimal. 
I think the mapping of chunks from one view to the other should be bijective (no source/target chunks should appear twice). This allows two use cases:
1. Both, the `target_view` and the `source_view`, have the same size and their `chunk_sizes` also match (e.g. copying data from A to B). (This is a special case of 2.)
2. The `source_view` and `target_view` can have different sizes (e.g. downsampling: the `source_view` is larger than `target_view` by a factor of `scale`). In this case, the ratio of the `chunk_size` of the views must be the same as the ratio between the sizes of the views.

This ensures that number of chunks for both views is the same. Therefore, `zip` works as expected.
I would propose, that I update the assertions in the function to the behavior described above.

Your given example is in my opinion not valid because `small_chunk_1` and `small_chunk_2` could be combined to one chunk, by choosing a different `target_chunk_size`. Is there a use case where your example would be nice and the behavior of above does not work?

Another thing that concerns me a little with this method is that the user has to be a bit careful with choosing the chunk_sizes: Let's say that data should be written to the `target_view`. The work on the different chunks is executed in parallel. This means that the `target_chunk_size` must be a aligned with the files on disk to avoid concurrent writes. Currently there is no such assertion because the `View` has no information about how the data is stored on disk.

wkcuber/api/Layer.py
rschwanhold: Only updating the properties does work. I have tested it locally, but before merging we should test everything on a large dataset and look at the results in webknossos. The large dataset should have an offset, so that as many edge cases as possible are covered (like this 'padding').

wkcuber/api/View.py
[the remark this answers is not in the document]
philippotto: Cool, that clears some things up for me :+1: Somehow I was stuck with the concept that the chunk size would be the cube size in which the wkw file is. With that assumption, it was hard to wrap my head around how the mapping would be bijective. However, your explanation makes total sense and I also think that this abstraction is more elegant to use.

Could you add that bit to the docstring of the method? Mainly, that the mapping is bijective and how this can be used to do stuff, such as downsampling.

> I would propose, that I update the assertions in the function to the behavior described above.

Yes, that would be good.

> Currently there is no such assertion because the View has no information about how the data is stored on disk.

Hm, I think, that would be a good thing to have. Otherwise, we will run into this sooner or later and have a hard time debugging this. Can we allow the `View` to access the header information somehow?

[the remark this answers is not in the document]
rschwanhold: I was wrong, the `View` already had the header. I have added the assertions only for the `target_view` because in my vision data should only be written to the `target_view` (as the name suggests). However, currently nothing prevents the user from also writing things to the `source_view`. This could be solved by introducing an optional parameter `read_only` to the function `get_view`. Do you think this is necessary?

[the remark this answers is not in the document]
philippotto: I think it would be better to be on the safe side. A `read_only` flag sounds like a very nice solution for this. If it's easy to implement, I'd vote for that :)

philippotto (approved):
I think, this is ready to be merged, right? :)

PR #263, opened 2021-01-11
Make mypy a dev dependency

rschwanhold:
fixes #263

### DIFF
--- a/poetry.lock
+++ b/poetry.lock
@@ -1,59 +1,62 @@
 [[package]]
-name = "appdirs"
-version = "1.4.4"
-description = "A small Python module for determining appropriate platform-specific dirs, e.g. a \"user data dir\"."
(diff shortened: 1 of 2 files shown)


rschwanhold:
I dont't understand why the order of many lines changes. All I did was to run `poetry update mypy`

jstriebel:
> I dont't understand why the order of many lines changes. All I did was to run `poetry update mypy`

True, that's weird. Did you maybe use a different mypy version?

jstriebel (approved):
LGTM :+1:

rschwanhold:
> > I dont't understand why the order of many lines changes. All I did was to run `poetry update mypy`
> 
> True, that's weird. Did you maybe use a different mypy version?

No, I used `0.770`

PR #265, opened 2021-01-22
Downgrade scikit-learn to 0.21.3

philippotto:
@leowe I saw that you integrated scikit-learn for https://github.com/scalableminds/webknossos-cuber/pull/261. However, this caused some downstream issues in voxelytics. Since voxelytics uses a lower version of scikit-learn, I suggest to downgrade the package in wkcuber. I assume that this shouldn't be a problem, right?

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       max-parallel: 4
       matrix:
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.6, 3.7]
 
     steps:
     - uses: actions/checkout@v1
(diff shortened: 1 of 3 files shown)


philippotto:
> I assume that this shouldn't be a problem, right?

Hm, I assume it's a problem that the dependencies cannot be installed for python 3.8 -.-

philippotto:
Follow up: https://github.com/scalableminds/webknossos-cuber/issues/267

philippotto:
Nevermind. We can workaround the original issue by using `spawn` or `forkserver` when creating subprocesses.

philippotto:
Good point! I closed it now.

PR #266, opened 2021-01-22
Update mypy version

rschwanhold:
This PR updates the mypy version to `0.800`. This release is better in detecting source files. The option `-v` now also logs which files were checked.

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -2,6 +2,8 @@ import filecmp
 import json
 import os
 from os.path import dirname, join
+from typing import Tuple, cast
+
 import pytest
 
 import numpy as np
@@ -9,12 +11,19 @@ from shutil import rmtree, copytree
 
 from wkw.wkw import WKWException
 
-from wkcuber.api.Dataset import WKDataset, TiffDataset, TiledTiffDataset
+from wkcuber.api.Dataset import (
+    WKDataset,
+    TiffDataset,
+    TiledTiffDataset,
+    AbstractDataset,
+)
 from os import path, makedirs
 
 from wkcuber.api.Layer import Layer
+from wkcuber.api.MagDataset import TiledTiffMagDataset
 from wkcuber.api.Properties.DatasetProperties import TiffProperties, WKProperties
 from wkcuber.api.TiffData.TiffMag import TiffReader
+from wkcuber.api.View import View
 from wkcuber.api.bounding_box import BoundingBox
 from wkcuber.compress import compress_mag_inplace
 from wkcuber.mag import Mag
@@ -23,12 +32,12 @@ from wkcuber.utils import get_executor_for_args
 expected_error_msg = "The test did not throw an exception even though it should. "
 
 
-def delete_dir(relative_path):
+def delete_dir(relative_path: str) -> None:
     if path.exists(relative_path) and path.isdir(relative_path):
         rmtree(relative_path)
 
 
-def chunk_job(args):
+def chunk_job(args: Tuple[View, tuple]) -> None:
     view, additional_args = args
 
     # increment the color value of each voxel
@@ -39,7 +48,7 @@ def chunk_job(args):
     view.write(data)
 
 
-def advanced_chunk_job(args):
+def advanced_chunk_job(args: Tuple[View, tuple]) -> None:
     view, additional_args = args
 
     # write different data for each chunk (depending on the global_offset of the chunk)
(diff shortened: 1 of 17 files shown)


wkcuber/api/Layer.py
```diff
@@ -107,12 +118,12 @@ class WKLayer(Layer):
...
-        mag: Union[str, Mag],
-        block_len: int = None,
-        file_len: int = None,
-        block_type: int = None,
+        self, mag: Union[int, str, list, tuple, np.ndarray, Mag], **kwargs: Any
     ) -> MagDataset:
+        block_len: int = kwargs.get("block_len", None)
+        file_len: int = kwargs.get("file_len", None)
+        block_type: int = kwargs.get("block_type", None)
+
```
rschwanhold, line 126: Unfortunately, I had to change the signature of the method to add this method to the base class. This was not an issue in #249 because this method was only called in the tests, which were not covered by `./typecheck`

tests/test_dataset.py
```diff
@@ -717,7 +726,10 @@ def test_tiled_tiff_read_and_write():
         pattern="{xxx}_{yyy}_{zzz}.tif",
     )
 
-    mag = tiled_tiff_ds.add_layer("color", Layer.COLOR_TYPE).add_mag("1")
+    mag = cast(
+        TiledTiffMagDataset,
+        tiled_tiff_ds.add_layer("color", Layer.COLOR_TYPE).add_mag("1"),
```
jstriebel, line 731: What would be the actual return type of `add_mag`? I'm just wondering why we need to cast here, but probably fine :+1:

wkcuber/api/Layer.py
jstriebel: Probably fine for now :+1: That mostly means that the classes are not interchangeable anymore, so this is mostly a question if we should change something about the API design. Not sure though, let's have a quick call about this :calling:

jstriebel (commented):
Nice, thanks for following up on this :heart: 
Mostly LGTM, just two minor nitpicks:

tests/test_dataset.py
rschwanhold: There are three different types of datasets (`WKDataset`, `TiffDataset`, `TiledTiffDataset`), each of these datasets has a corresponding "MagDataset" (`WKMagDataset`, `TiffMagDataset`, `TiledTiffMagDataset`) (with `MagDataset` being the base class). 
`TiledTiffMagDataset` has an extra method, to access the tiles by an index. This is not part of `MagDataset` and the only case where a subclass defines an additional function. 
The function `add_mag` of `Layer` returns the corresponding MagDataset, but the typing hint is only `MagDataset` (not e.g. `TiledTiffMagDataset`) . Therefore, the typechecker would complain about `assert mag.get_tile(1, 1, 6).shape == (1, 32, 64, 1)` if the type is not casted.

jstriebel: Ok, this is fine for now :+1: If we need specialized corresponding classes more often, we can use generics for this. E.g. defining
```
MagT = TypeVar('MagT')
class Layer(Generic[MagT]):
    add_layer(…) -> MagT:
        …

class TiledTiffLayer(Layer[TiledTiffMagDataset]):
    …
```

rschwanhold:
@philippotto we would like to get a third opinion from you. @jstriebel and I made the typing for the dataset api very generic. Therefore, it is no longer necessary to cast the type of the dataset (as pointed out in the review of @jstriebel ).

jstriebel:
@philippotto I think you can focus on the changes in the latest commit, those are the ones we authored together, and I'm afraid I might be to involved to review them.

wkcuber/api/MagDataset.py
```diff
@@ -165,11 +171,16 @@ def __init__(
         block_len: int,
         file_len: int,
         block_type: int,
+        exists: bool = True,
```
philippotto, line 174: Personally, I'd rename this (and invert the semantics) to `create` to signal the action this will have. But I don't know where this is coming from. Maybe there is a stronger reason for `exists`?

philippotto (approved):
Cool types :100: The generics are especially nice. Didn't know that these could help so much for this class architecture. Learned something new :)

I only have one not-type-related naming comment. Otherwise :100: 

(I only reviewed the latest commit as suggested by @jstriebel)

jstriebel (approved):
:+1:

PR #268, opened 2021-01-28
Auto-detect the right conversion 

### LINKED ISSUE #177: Cubing 16-bit tiffs without providing --dtype simply crops that data
I suggest that the dtype should simply be inferred from the src image. So, if the input data is in 16 bit, than the dtype defaults to 16 bit.

Do you agree, @normanrz?

### LINKED ISSUE #251: Auto-detect what conversion to run

youri-k:
fixes #251 
fixes #177

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,10 +11,10 @@ The tools are modular components to allow easy integration into existing pipelin
 ## Features
 
 * `wkcuber`: Convert supported input files to fully ready WKW datasets (includes type detection, downsampling, compressing and metadata generation)
+* `wkcuber.convert_image_stack_to_wkw`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
 * `wkcuber.export_wkw_as_tiff`: Convert WKW datasets to a tiff stack (writing as tiles to a `z/y/x.tiff` folder structure is also supported)
 * `wkcuber.cubing`: Convert image stacks (e.g., `tiff`, `jpg`, `png`, `dm3`, `dm4`) to WKW cubes
 * `wkcuber.tile_cubing`: Convert tiled image stacks (e.g. in `z/y/x.ext` folder structure) to WKW cubes
-* `wkcuber.convert_image_stack_to_wkw`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
 * `wkcuber.convert_knossos`: Convert KNOSSOS cubes to WKW cubes
 * `wkcuber.convert_nifti`: Convert NIFTI files to WKW files (Currently without applying transformations).
 * `wkcuber.downsampling`: Create downsampled magnifications (with `median`, `mode` and linear interpolation modes). Downsampling compresses the new magnifications by default (disable via `--no-compress`).
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ The tools are modular components to allow easy integration into existing pipelin
 
 * `wkcuber`: Convert supported input files to fully ready WKW datasets (includes type detection, downsampling, compressing and metadata generation)
 * `wkcuber.export_wkw_as_tiff`: Convert WKW datasets to a tiff stack (writing as tiles to a `z/y/x.tiff` folder structure is also supported)
-* `wkcuber.cubing`: Convert image stacks (e.g., `tiff`, `jpg`, `png`, `dm3`) to WKW cubes
+* `wkcuber.cubing`: Convert image stacks (e.g., `tiff`, `jpg`, `png`, `dm3`, `dm4`) to WKW cubes
 * `wkcuber.tile_cubing`: Convert tiled image stacks (e.g. in `z/y/x.ext` folder structure) to WKW cubes
 * `wkcuber.convert_image_stack_to_wkw`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
 * `wkcuber.convert_knossos`: Convert KNOSSOS cubes to WKW cubes
--- a/README.md
+++ b/README.md
@@ -10,10 +10,11 @@ The tools are modular components to allow easy integration into existing pipelin
 
 ## Features
 
-* `wkcuber`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
+* `wkcuber`: Convert supported input files to fully ready WKW datasets (includes type detection, downsampling, compressing and metadata generation)
 * `wkcuber.export_wkw_as_tiff`: Convert WKW datasets to a tiff stack (writing as tiles to a `z/y/x.tiff` folder structure is also supported)
 * `wkcuber.cubing`: Convert image stacks (e.g., `tiff`, `jpg`, `png`, `dm3`) to WKW cubes
 * `wkcuber.tile_cubing`: Convert tiled image stacks (e.g. in `z/y/x.ext` folder structure) to WKW cubes
+* `wkcuber.convert_image_stack_to_wkw`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
 * `wkcuber.convert_knossos`: Convert KNOSSOS cubes to WKW cubes
 * `wkcuber.convert_nifti`: Convert NIFTI files to WKW files (Currently without applying transformations).
 * `wkcuber.downsampling`: Create downsampled magnifications (with `median`, `mode` and linear interpolation modes). Downsampling compresses the new magnifications by default (disable via `--no-compress`).
--- a/tests/test_auto_detection.py
+++ b/tests/test_auto_detection.py
@@ -145,7 +145,7 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         assert dataset_name == "knossos"
         assert len(layer_paths) == 1
         assert list(layer_paths.keys())[0] == prefix + "knossos/color"
-        assert list(layer_paths.values())[0] == "1"
+        assert list(layer_paths.values())[0] == {"1"}
 
         # test if in subfolder
         converter = KnossosConverter()
@@ -160,7 +160,7 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
             list(layer_paths.keys())[0]
             == prefix + "superfolder/superfolder/knossos/color"
         )
-        assert list(layer_paths.values())[0] == "1"
+        assert list(layer_paths.values())[0] == {"1"}
 
         # test for multiple layer
         converter = KnossosConverter()
@@ -175,7 +175,7 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         assert len(layer_paths) == 2
         assert prefix + "knossos/color" in layer_paths.keys()
         assert prefix + "knossos/segmentation" in layer_paths.keys()
-        assert all(map(lambda m: m == "1", layer_paths.values()))
+        assert all(map(lambda m: m == {"1"}, layer_paths.values()))
 
         # test if only layer folder given
         converter = KnossosConverter()
@@ -186,7 +186,7 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         assert dataset_name == "dataset"
         assert len(layer_paths) == 1
         assert list(layer_paths.keys())[0] == prefix + "color"
-        assert list(layer_paths.values())[0] == "1"
+        assert list(layer_paths.values())[0] == {"1"}
 
         # test if only mag folder given
         converter = KnossosConverter()
@@ -197,7 +197,7 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         assert dataset_name == "dataset"
         assert len(layer_paths) == 1
         assert list(layer_paths.keys())[0] == prefix
-        assert list(layer_paths.values())[0] == "1"
+        assert list(layer_paths.values())[0] == {"1"}
 
         # test if already in mag folder
         converter = KnossosConverter()
@@ -208,7 +208,7 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         assert dataset_name == "dataset"
         assert len(layer_paths) == 1
         assert list(layer_paths.keys())[0] == prefix
-        assert list(layer_paths.values())[0] == ""
+        assert list(layer_paths.values())[0] == {""}
 
         # test if too short path gets detected
         converter = KnossosConverter()
@@ -234,4 +234,4 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         assert dataset_name == "knossos"
         assert len(layer_paths) == 1
         assert list(layer_paths.keys())[0] == prefix + "knossos/color"
-        assert list(layer_paths.values())[0] == "2"
+        assert list(layer_paths.values())[0] == {"2", "4"}
--- a/tests/test_auto_detection.py
+++ b/tests/test_auto_detection.py
@@ -1,0 +1,84 @@
+from argparse import Namespace
+
+from wkcuber.converter import ImageStackConverter
+
+TEST_NAMESPACE = Namespace()
+TEST_NAMESPACE.target_path = "a/random/path"
+
+
+def test_tiff_dataset_name_and_layer_name_detection():
+    converter = ImageStackConverter()
+
+    # test if ds name and layer name are correctly detected
+    converter.source_files = [
+        "test/color/001.tif",
+        "test/color/002.tif",
+        "test/color/003.tif",
+    ]
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "test"
+    assert len(layer_names) == 1
+    assert layer_names[0] == "color"
+
+    # test if in subfolder
+    converter.source_files = [
+        "superfolder/test/color/001.tif",
+        "superfolder/test/color/002.tif",
+        "superfolder/test/color/003.tif",
+    ]
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "test"
+    assert len(layer_names) == 1
+    assert layer_names[0] == "color"
+
+    # test for multiple layers
+    converter.source_files = [
+        "test/color/001.tif",
+        "test/color/002.tif",
+        "test/color/003.tif",
+        "test/segmentation/001.tif",
+        "test/segmentation/002.tif",
+        "test/segmentation/003.tif",
+    ]
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "test"
+    assert len(layer_names) == 2
+    assert "color" in layer_names
+    assert "segmentation" in layer_names
+
+    # test if in single folder and folder name is layer name
+    converter.source_files = [
+        "color/001.tif",
+        "color/002.tif",
+        "color/003.tif",
+    ]
+    converter.args = TEST_NAMESPACE
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "path"
+    assert len(layer_names) == 1
+    assert layer_names[0] == "color"
+
+    # test if in single folder and folder name is ds name
+    converter.source_files = [
+        "dataset/001.tif",
+        "dataset/002.tif",
+        "dataset/003.tif",
+    ]
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "dataset"
+    assert len(layer_names) == 1
+    assert layer_names[0] == "color"
+
+    # test if single file in folder
+    converter.source_files = ["dataset/brain.tif"]
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "dataset"
+    assert len(layer_names) == 1
+    assert layer_names[0] == "brain"
+
+    # test if single file
+    converter.source_files = ["brain.tif"]
+    dataset_name, layer_names = converter.detect_dataset_and_layer_names()
+    assert dataset_name == "brain"
+    assert len(layer_names) == 1
+    assert layer_names[0] == "color"
--- a/tests/test_auto_detection.py
+++ b/tests/test_auto_detection.py
@@ -100,7 +100,7 @@ def test_tiff_dataset_name_and_layer_name_detection() -> None:
         ) = converter.detect_dataset_name_and_layer_path_to_layer_name()
         assert dataset_name == "test_dataset"
         assert len(layer_path_to_layer_name) == 1
-        assert list(layer_path_to_layer_name)[0] == prefix + "test_dataset"
+        assert list(layer_path_to_layer_name)[0] == prefix + "test_dataset/brain.tif"
         assert list(layer_path_to_layer_name.values())[0] == "brain"
 
         # test if single file
@@ -112,7 +112,7 @@ def test_tiff_dataset_name_and_layer_name_detection() -> None:
         ) = converter.detect_dataset_name_and_layer_path_to_layer_name()
         assert dataset_name == "brain"
         assert len(layer_path_to_layer_name) == 1
-        assert list(layer_path_to_layer_name)[0] == prefix
+        assert list(layer_path_to_layer_name)[0] == prefix + "brain.tif"
         assert list(layer_path_to_layer_name.values())[0] == "color"
 
         # test for multiple files with no parent directory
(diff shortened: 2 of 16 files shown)


youri-k:
I think you can already review the general structure. I noticed that KNOSSOS cubes were missing in the original issue description, so I'm currently adding support for those.
I'm especially interested in how you like the argument parsing mechanism. I just read that argument parser could also parse known args only, which would could lead to a clearer architecture.

philippotto (commented):
Thanks for working on this! The current version looks promising!

On a high-level perspective I have some remarks:
- A very common case for the input datasets are multiple layers. For example, I could pass `.../dataset_X` which contains `.../dataset_X/layer_1` and `.../dataset_X/layer_2`. It would be very cool, if this was supported out of the box. (I assume that was also meant with "recursively check the folders" in the original issue by @normanrz). In that scenario, the layer names would be detected automatically.
- overall, we want to be quite robust regarding different input formats (e.g., if the input is wrapped in another folder, it should still be recognized correctly)
- tiled image stacks are not supported, yet, right? these are usually expected to be in a `z/y/x.ext` hierarchy, which would need to be detected so that the tiled cubing code-path can be invoked.
- implementation-wise it would make sense to have a `WkwConverter` parent class with several subclasses, such as: `NiftiConverter`, `ImageStackConverter`, `TiledImageStackConverter` etc. These converter classes could implement an `acceptsInput(input_path: str)` method. then, the main-code could simply ask the converter-classes whether they are able to convert the input. that would be cleaner than deciding from the outside which conversion should be picked. not sure, whether the converter classes should know about the "layers" concept.
- The use cases "nested input folder with multiple layers" and "nested/tiled images in z/y/x" might interfere with each other. We should think this through before implementing it.
- tests would be cool :)

A bit unrelated, but still interesting: should we also support archives as inputs? the cuber could unzip/untar it to /tmp and then convert it. I imagine this to be quite useful for webKnossos' auto-convert feature.

Let me know if you want to discuss this in person!

normanrz:
I agree with @philippotto that we need tests for this. There are many different ways how people store their image files. With the tests we can make sure that we support them all and don't break them, when we implement a new pattern.

I also think that the OO approach that @philippotto could help to solve the complexity that this will have. 

I don't think tiled images need to be support right now. Also, I think layers can be supported selectively (e.g. for knossos datasets or single tiff files that have them).

youri-k:
@philippotto could you have a look at the image stack converter? This should be ready for review except the multi layer case.

wkcuber/tile_cubing.py
```diff
@@ -291,6 +289,10 @@ def tile_cubing(
             file_count, tile_size[0], tile_size[1]
         )
     )
+    if args is None or not hasattr(args, "dtype") or args.dtype is None:
+        dtype = image_reader.read_dtype(arbitrary_file)
+        if args is not None:
```
philippotto, line 294: If args is None, the dtype isn't passed anywhere if I'm seeing this correctly.

wkcuber/image_readers.py
```diff
@@ -28,6 +28,9 @@ def read_z_slices_per_file(
     ) -> int:
         return 1
 
+    def read_dtype(self, file_name: str) -> str:  # pylint: disable=unused-argument
+        return "uint8"
```
philippotto, line 32: why does this default to uint8?

wkcuber/convert_nifti.py
```diff
@@ -13,47 +13,24 @@
...
-        "-l",
-        help="Name of the cubed layer (color or segmentation).",
-        default="color",
-    )
-
+def add_nifti_arguments(parser: ArgumentParser) -> ArgumentParser:
     parser.add_argument(
         "--is_segmentation_layer",
-        "-s",
+        "-sl",
```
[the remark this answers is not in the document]
philippotto: does it interfere with the scale flag?

wkcuber/cubing.py
```diff
@@ -95,17 +96,25 @@ def create_parser() -> ArgumentParser:
...
-    # Find all files in a folder that have a matching file extension
-    source_files = list(
-        find_files(path.join(source_path, "*"), image_reader.readers.keys())
-    )
+    # Find all source files that have a matching file extension
+
+    input_path = Path(source_path)
+
+    if input_path.is_dir():
+        joined_path = path.join(source_path, "**")
```
philippotto, line 104: the old code only had one `*`. is this intended for cubing nested structures? what happens if I pass a folder which contains two layers? within this pattern matching logic, the layer name extraction would need to happen. otherwise, all input files (of both layers) are thrown together, arent't they? :thinking: or is the `converter` class responsible for dissecting this? then, I'm not sure whether `**` is necessary here..

wkcuber/cubing.py
```diff
@@ -251,17 +259,24 @@ def cubing(
...
         layer_name,
         target_mag,
         wkw.Header(
-            convert_element_class_to_dtype(dtype),
+            convert_element_class_to_dtype(args.dtype),
             num_channels,
             file_len=args.wkw_file_len,
         ),
     )
+    # TODO I think this uses a different layer category guessing method
```
philippotto, line 279: todo

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        put_default_if_not_present(args, "isotropic", False)
+        put_default_if_not_present(args, "verbose", True)
+
+        # detect layer and ds name
+        dataset_name, layer_names = self.detect_dataset_and_layer_names()
+        put_default_if_not_present(args, "name", dataset_name)
+
+        self.args = args
+
+        # TODO we can detect the layer names correctly, but we currently do not split the source files accordingly
```
philippotto, line 209: todo

philippotto (commented):
Cool stuff! Looks already very good. I'm not completely done with the `converter.py` review, but already left some comments.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+            len(dataset_names) == 1
+        ), "More than one dataset found. Stopping conversion..."
+        assert len(layers) > 0, "No layers found. Stopping conversion..."
+
+        layer_names_and_mags = dict()
+
+        for layer, mags in layers.items():
+            layer_names_and_mags[layer] = natsorted(list(mags))[0]
+
+        return dataset_names.pop(), layer_names_and_mags
```
philippotto, line 162:
```suggestion
        return dataset_names[0], layer_names_and_mags
```

pop mutates the list, which I would avoid as a habit when possible (even though, it doesn't make a difference here)

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+    elif len(fitting_converters) > 1:
+        print(
+            "Multiple converters found. Check if your source path contains multiple datasets."
+        )
+        exit(1)
+
+    converter = fitting_converters[0]
+    print("Choosing the", converter.__class__.__name__)
+
+    converter.convert_input(args)
```
philippotto, line 283: I would put this in a `main` function and call that here. This avoids "polluting" the global state.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+    def __init__(self) -> None:
+        self.converter: List[Converter] = [
+            WkwConverter(),
+            NiftiConverter(),
+            KnossosConverter(),
+            ImageStackConverter(),
+        ]
+
+
+converter_manager = ConverterManager()
```
philippotto, line 257: This should probably only happen in `__main__` / `main()`

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+
+if __name__ == "__main__":
+    args = create_parser().parse_args()
+
+    fitting_converters = list(
```
philippotto, line 262: nit: `matching_converters` sounds better to me

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        filter(
+            lambda c: c.accepts_input(args.source_path),
+            converter_manager.converter,
+        )
+    )
+
+    if len(fitting_converters) == 0:
+        print(
+            "No converter found. Please specify which converter you want to use or check the source path."
```
philippotto, line 271: How would one specify the converter?

philippotto (changes requested):
Awesome stuff! The tests look promising :rocket:  Left some feedback and questions :)

wkcuber/convert_nifti.py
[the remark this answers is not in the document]
youri-k: Yes

wkcuber/tile_cubing.py
youri-k: Yes, you're correct. :slightly_smiling_face:

wkcuber/image_readers.py
youri-k: This was my workaround because I didn't know how to read the dtype from other libraries. I added a comment that this needs to be done.

wkcuber/cubing.py
youri-k: Yes, you're correct again. Initially, I wanted to add the layer config into the cubing model, but now it is handled by the `converter` class, so I can roll back to the old version.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+    dictionary = vars(args)
+    if name not in dictionary:
+        dictionary[name] = default
+
+
+def add_or_update_if_present(dictionary: dict, name: str, value: Any) -> None:
+    if name not in dictionary:
+        dictionary[name] = {value}
+    else:
+        dictionary[name].add(value)
```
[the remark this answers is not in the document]
youri-k: Ah, thanks for the suggestion.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+
+
+def get_source_files(
+    input_path: str, extensions: Iterable[str], allows_single_file_input: bool
+) -> List[str]:
+    source_path = Path(input_path)
```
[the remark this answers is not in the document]
youri-k: Went with the second suggestion because the utils method expects a string

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        pass
+
+
+class WkwConverter(Converter):
+    def accepts_input(self, source_path: str) -> bool:
+        source_files = get_source_files(source_path, {".wkw"}, False)
+        return len(source_files) > 0
+
+    def convert_input(self, args: Namespace) -> None:
+        print("Already a WKW dataset. No conversion necessary...")
```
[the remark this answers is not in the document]
youri-k: This is a more general question for me: Do we want to have so much console output? I created these `print` calls as validation for me, but maybe they are interesting for the user as well.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+            args.layer_name = "color" if layer_path == "" else path.basename(layer_path)
+            args.name = dataset_name
+
+            convert_knossos(args)
+
+    def detect_dataset_and_layer_paths_with_mag(self) -> Tuple[str, Dict[str, str]]:
+        # Path structure for knossos is .../(dataset_name)/(layer_name)/(mag)folder/x0000/y0000/z0000/filename.raw
+        layers: Dict[str, set] = dict()
+        dataset_names = set()
+        for f in self.source_files:
```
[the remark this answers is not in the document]
youri-k: This is an interesting question, yes it does iterate over all files. Since I already extracted all source files, I didn't even think about other traversal methods.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+                dataset_names.add(split_path[-7])
+
+        assert (
+            len(dataset_names) == 1
+        ), "More than one dataset found. Stopping conversion..."
+        assert len(layers) > 0, "No layers found. Stopping conversion..."
+
+        layer_names_and_mags = dict()
+
+        for layer, mags in layers.items():
```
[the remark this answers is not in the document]
youri-k: > `add_or_update_if_present(layers, "", split_path[0])`
`layers` is dict from layer path to a set of existing mags. Every time we discover a new mag it is inserted into that set, but for the first mag the set needs to be created. That's why I created the helper method, but I see that the method name might be confusing, so I'm happy if you have a suggestion for a better name.

We end up with a set of all mags, which is then converted to a list to be sorted and we extract the "smallest" mag to use the best available quality for the data conversion.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        put_default_if_not_present(args, "interpolation_mode", "default")
+        put_default_if_not_present(args, "start_z", 0)
+        put_default_if_not_present(args, "jobs", cpu_count())
+        put_default_if_not_present(args, "distribution_strategy", "multiprocessing")
+        put_default_if_not_present(args, "job_resources", None)
+        put_default_if_not_present(args, "pad", False)
+        put_default_if_not_present(args, "max_mag", 32)
+        put_default_if_not_present(args, "no_compress", False)
+        put_default_if_not_present(args, "isotropic", False)
+        put_default_if_not_present(args, "verbose", True)
```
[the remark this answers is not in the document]
youri-k: Since they are only defined inside the argument parser, I am not sure how to handle this in a more DRY fashion. The first idea that comes to my mind, would be to set the default of every value to `None` in the argument parser and have a similar validator check this after argument parsing. Then we could have a unified default args dict inside each class. Do you think this is a good idea?

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        source_files = get_source_files(source_path, image_reader.readers.keys(), True)
+
+        if len(source_files) == 0:
+            return False
+
+        _, ext = path.splitext(source_files[0])
+
+        assert all(
+            map(lambda p: path.splitext(p)[1] == ext, source_files)
+        ), "Not all image files are of the same type"
```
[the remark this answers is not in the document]
youri-k: There was no similar assertion to that previously, but I was anxious to just allow this, since this order of processing depends on the file names, so there could be some inconsistencies. However, if you think this is not a problem, I'll gladly remove that check again.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+
+        # TODO we can detect the layer names correctly, but we currently do not split the source files accordingly
+        args.layer_name = layer_names[0]
+        convert_image_stack(args)
+
+    def detect_dataset_and_layer_names(self) -> Tuple[str, List[str]]:
+        paths = set()
+        for f in self.source_files:
+            p = path.dirname(f)
+            if p != "":
```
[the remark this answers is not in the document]
youri-k: p is `""` when the user has only a single input file which is located in the current working directory.

wkcuber/converter.py
youri-k: dataset_names is a `set`, so it is not subscriptable and I check that the set only contains one element. Therefore, I would rather call `pop` than converting the set to a list and then taking the first element of that list.

wkcuber/converter.py
youri-k: This was an initial idea of me that you could overwrite the converter checking. However, this should probably be removed or do you think there is a use case for that?

wkcuber/image_readers.py
```diff
@@ -28,6 +28,9 @@ def read_z_slices_per_file(
     ) -> int:
         return 1
 
+    def read_dtype(self, file_name: str) -> str:  # pylint: disable=unused-argument
+        return "uint8"  # TODO: Implement dtype method for all readers
```
philippotto, line 32: todo

wkcuber/converter.py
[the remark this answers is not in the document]
philippotto: I think, high-level logging which shows which "insights" the cuber has gathered and which actions it intends to make is useful. I think one needs to decide case-by-case if something is useful. Usually, it's best to validate this by running the cuber and inspecting whether the output seems useful (instead of going through the logging code).

If something is too detailed, one can also switch from `logger.info` to `logger.debug`. Then, that information is only printed when `--verbose` is used :)

wkcuber/converter.py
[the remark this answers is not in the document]
philippotto: I think, the code would be easier to read when an actual tree traversal is done, but let's discuss this in person.

wkcuber/converter.py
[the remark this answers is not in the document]
philippotto: > There was no similar assertion to that previously,

Ok, then let's just keep it :)

wkcuber/converter.py
philippotto: I think, there is no strong need for this, since we also have the other high-level entry points for conversion. In the long run, it might make sense to unify all of this, so there is simply `python -m cuber convert ...` with optional `--source_type tiff`. But that's not important for now imo.

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+    return parser
+
+
+def put_default_if_not_present(args: Namespace, name: str, default: Any) -> None:
+    dictionary = vars(args)
+    if name not in dictionary:
+        dictionary[name] = default
+
+
+def add_or_update_if_present(dictionary: dict, name: str, value: Any) -> None:
```
philippotto, line 36: `add_to_set_in_dictionary`

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        for layer_path, mag in layer_paths_and_mag.items():
+            args.source_path = path.join(layer_path, mag)
+            args.layer_name = "color" if layer_path == "" else path.basename(layer_path)
+            args.name = dataset_name
+
+            convert_knossos(args)
+
+    def detect_dataset_and_layer_paths_with_mag(self) -> Tuple[str, Dict[str, str]]:
+        # Path structure for knossos is .../(dataset_name)/(layer_name)/(mag)folder/x0000/y0000/z0000/filename.raw
+        layers: Dict[str, set] = dict()
```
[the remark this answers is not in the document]
philippotto: `layer_path_to_mag_set`

tests/test_auto_detection.py
```diff
@@ -0,0 +1,184 @@
...
+    ]
+    dataset_name, layer_paths = converter.detect_dataset_and_layer_paths_with_mag()
+    assert dataset_name == "knossos"
+    assert len(layer_paths) == 1
+    assert list(layer_paths.keys())[0] == "/knossos/color"
+    assert list(layer_paths.values())[0] == "1"
+
+    # test if too short absolute path gets detected
+    converter.source_files = [
+        "/y0000/z0000/test_mag1_x0000_y0000_z0000.raw",
```
philippotto, line 166: maybe do all test cases with these three scenarios:
- without leading slash
- with leading "../"
- with leading "/"

wkcuber/converter.py
```diff
@@ -0,0 +1,283 @@
...
+        put_default_if_not_present(args, "name", dataset_name)
+
+        self.args = args
+
+        # TODO we can detect the layer names correctly, but we currently do not split the source files accordingly
+        args.layer_name = layer_names[0]
+        convert_image_stack(args)
+
+    def detect_dataset_and_layer_names(self) -> Tuple[str, List[str]]:
+        paths = set()
```
philippotto, line 208: `parent_paths_of_images`

youri-k:
@philippotto I think this is ready for another round of review :slightly_smiling_face: I hope I addressed all of your suggestions and I'm happy to hear more feedback.

wkcuber/converter.py
```diff
@@ -0,0 +1,461 @@
...
+        self.source_files: List[str] = []
+        self.prefix: str = ""
+
+    def accepts_input(self, source_path: str) -> bool:
+        pass
+
+    def convert_input(self, args: Namespace) -> None:
+        pass
+
+    # returns [prefix, traversal_depth]
```
philippotto, line 78: seems outdated? can probably be removed.
```suggestion
```

wkcuber/converter.py
```diff
@@ -0,0 +1,461 @@
...
+            self.prefix = "/"
+        elif first_split_path[0] == "..":
+            self.prefix = "../"
+
+        assert all(
+            map(
+                lambda p: len(p.split(sep)) == traversal_depth,
+                self.source_files,
+            )
+        )
```
philippotto, line 93: please add an assertion message which explains what went wrong (`assert ..., "msg"`)

wkcuber/converter.py
```diff
@@ -0,0 +1,461 @@
...
+
+        assert (
+            traversal_depth >= 4 if not starts_with_prefix else traversal_depth >= 5
+        ), "Input Format is unreadable. Make sure to pass the path which points at least to a KNOSSOS magnification (e.g., testdata/knossos/color/1)."
+        assert all(
+            map(
+                lambda p: len(p.split(sep)) == traversal_depth,
+                self.source_files,
+            )
+        )
```
philippotto, line 223: please add an assertion message

philippotto (approved):
Great work! Thanks for pulling this through the finish line. I left some cosmetic feedback, but nothing major.

Could you create a follow-up issue that the converter classes should also be used when not using `__main__` (but for example `wkcuber.cubing`). That way, we can further refactor the argparse code, I hope.

Also, please update the readme in the following section, so that it explains the auto-convert behavior.

> wkcuber: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)

youri-k:
@philippotto Can you have one last look? I hope this is good to go now :slightly_smiling_face:

README.md
```diff
@@ -10,10 +10,11 @@ The tools are modular components to allow easy integration into existing pipelin
 
 ## Features
 
-* `wkcuber`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
+* `wkcuber`: Convert supported input files to fully ready WKW datasets (includes type detection, downsampling, compressing and metadata generation)
 * `wkcuber.export_wkw_as_tiff`: Convert WKW datasets to a tiff stack (writing as tiles to a `z/y/x.tiff` folder structure is also supported)
-* `wkcuber.cubing`: Convert image stacks (e.g., `tiff`, `jpg`, `png`, `dm3`) to WKW cubes
+* `wkcuber.cubing`: Convert image stacks (e.g., `tiff`, `jpg`, `png`, `dm3`, `dm4`) to WKW cubes
 * `wkcuber.tile_cubing`: Convert tiled image stacks (e.g. in `z/y/x.ext` folder structure) to WKW cubes
+* `wkcuber.convert_image_stack_to_wkw`: Convert image stacks to fully ready WKW datasets (includes downsampling, compressing and metadata generation)
```
philippotto, line 17: I'd move that directly below the `wkcuber` part :)

wkcuber/__main__.py
```diff
@@ -36,52 +57,49 @@ def create_parser() -> ArgumentParser:
...
-            not args.no_compress,
-            args=args,
-        )
+        for (layer_path, mags) in layer_path_to_mags.items():
+            layer_name = layer_path.stem
+            mags.sort(reverse=True)
+            downsample_mags_anisotropic(
+                args.target_path,
+                layer_name,
+                mags[0],
```
philippotto, line 77: shouldn't it be this?
```suggestion
                mags[-1],
```

Otherwise, mag 1 would overwrite the existing mag 2 if both exist? Also: Is the mags list natsorted? This would be necessary for this to work.

philippotto: Ah, I missed line 73. Alrighty then. I'd prefer to not use `reverse=True` and instead access `[-1]`. But you can leave it as is.

However: Wouldn't the sorting need to be natsorted? For example, sorting `[1, 8, 16]` wouldn't work otherwise, I think.

philippotto (approved):
Awesome work! See my last few comments, but after that, feel free to merge :)

wkcuber/__main__.py
youri-k: As far as I tested it, it works. My guess would be that the `Mag` objects that we're using have an internal correct sorting method (or less-equal method). I went with your suggestion and accessed `[-1]`

philippotto: > My guess would be that the Mag objects that we're using have an internal correct sorting method (or less-equal method).

Ah, you are right! I thought of them as strings, but that was wrong :) The class indeed implements `__le__` and co :thinking:

PR #271, opened 2021-02-02
Improve knossos conversion

philippotto:
The file listing was able to find arbitrarily nested raw files, but the actual conversion code didn't find the files later and produced only black output. I adapted the code so that the files can only be found when the nesting is as expected (e.g., `testdata/knossos/color/1/x0001/y0001/z0001/file_name.raw`). Otherwise, an error message will be raised like this:

```
# python -m wkcuber.convert_knossos   --jobs 2   --dtype uint8   --layer_name color   --mag 1   testdata/knossos/color/ testoutput/knossos
2021-02-02 13:52:38,430 INFO Using pool of 2 workers.
2021-02-02 13:52:38,431 ERROR No input KNOSSOS cubes found. Make sure to pass the path which points to a KNOSSOS magnification (e.g., testdata/knossos/color/1).
```

### DIFF
--- a/wkcuber/convert_knossos.py
+++ b/wkcuber/convert_knossos.py
@@ -94,7 +94,9 @@ def convert_knossos(
         with get_executor_for_args(args) as executor:
             knossos_cubes = list(source_knossos.list_cubes())
             if len(knossos_cubes) == 0:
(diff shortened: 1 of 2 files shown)


fm3 (approved):
LGTM :)

Issue #157, opened 2019-11-18
Downsampling to an anisotropic target mag should output that mag properly


philippotto:
At the moment, defining something like 32-32-1 as a target mag will create all mags until 32 using the scale to aim for near-cubic voxels. The target mag is merely used as an abort criteria to avoid that mag 64 is created). Therefore, the output is likely to be something like 32-32-8.

For this case it might be necessary to differentiate between `32-32-32` and `32` (the first might enforce isotropic downsampling and the last "auto-anisotropy"-downsampling). Alternatively, two different interfaces might be more elegant here?

PR #279, opened 2021-03-12
Calculate max mag from dataset size

### LINKED ISSUE #152: Default max-mag in downsample task should depend on size of dataset
If the input dataset is only 1k**3 voxel big, there is no need to create all mags until 512. I think, we can find a sensible value for this automatically.

rschwanhold:
Previously, the default value for `downsampling` was always `512`, regardless of the size of the dataset. This new approach calculates the default value depending on the size of the dataset. The max mag will always be at least `Mag(4)`.

fixes #152 
fixes: #157 (This already worked before for `downsampling.py` but now it also works for `Layer.downsample`)

### DIFF
--- a/wkcuber/downsampling.py
+++ b/wkcuber/downsampling.py
@@ -34,22 +34,16 @@ def calculate_virtual_scale_for_target_mag(
 def calculate_default_max_mag_for_dataset(dataset_path: str, layer_name: str) -> Mag:
     ds = WKDataset(dataset_path)
     ds_size = ds.properties.data_layers[layer_name].get_bounding_box_size()
-    cube_length = (
-        ds.properties.data_layers[layer_name].wkw_magnifications[0].cube_length
-    )
-    return calculate_default_max_mag(ds_size, cube_length)
+    return calculate_default_max_mag(ds_size)
 
 
-def calculate_default_max_mag(
-    dataset_size: Tuple[int, int, int], cube_length: int
-) -> Mag:
-    cubes_per_dim = -(-np.array(dataset_size) // np.array(cube_length))
-    magnification = [
-        int(math.pow(2, int(math.log(s, 2)))) for s in cubes_per_dim
-    ]  # highest power of 2 smaller than cubes_per_dim
+def calculate_default_max_mag(dataset_size: Tuple[int, int, int]) -> Mag:
+    # The lowest mag should have a size of ~ 100vx**3
+    max_x_y = max(dataset_size[0], dataset_size[1])
+    # highest power of 2 larger (or equal) than max_x_y divided by 100
     return Mag(
-        [min(max(m, 16), 512) for m in magnification]
-    )  # at least 16 and at most 512
+        max(int(math.pow(2, math.ceil(math.log(max_x_y / 100, 2)))), 4)
+    )  # at least 4
 
 
 def create_parser() -> ArgumentParser:
@@ -81,7 +75,9 @@ def create_parser() -> ArgumentParser:
         help="Max resolution to be downsampled. In case of anisotropic downsampling, the process is considered "
         "done when max(current_mag) >= max(max_mag) where max takes the largest dimension of the mag tuple "
         "x, y, z. For example, a maximum mag value of 8 (or 8-8-8) will stop the downsampling as soon as a "
-        "magnification is produced for which one dimension is equal or larger than 8.",
+        "magnification is produced for which one dimension is equal or larger than 8. "
+        "The default value is calculated depending on the dataset size. In the lowest Mag, the size will be "
+        "smaller than 100vx per dimension",
         type=int,
         default=None,
     )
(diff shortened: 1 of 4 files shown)


wkcuber/downsampling.py
```diff
@@ -30,6 +31,27 @@ def calculate_virtual_scale_for_target_mag(
...
+def calculate_default_max_mag(
+    dataset_size: Tuple[int, int, int], cube_length: int
+) -> Mag:
+    cubes_per_dim = -(-np.array(dataset_size) // np.array(cube_length))
+    magnification = [
+        int(math.pow(2, int(math.log(s, 2)))) for s in cubes_per_dim
+    ]  # highest power of 2 smaller than cubes_per_dim
+    return Mag(
+        [min(max(m, 16), 512) for m in magnification]
+    )  # at least 16 and at most 512
```
rschwanhold, line 52: I have no good intuition on how large the typical datasets are and what magnifications are necessary. @philippotto is the order of magnitude correct?

philippotto: [just for the record, as we already talked about most of this in person]

I gave this a bit of thought and I think what is ultimately needed is a very small, lightweight representation of the entire dataset. For this, we can set an arbitrary threshold such as 100*100*100 vx. That would be 1 MB uncompressed (in total) and one slice of it would be 10 kB (ideal for a thumbnail). Making the dataset even smaller probably doesn't help much for most usecases.

I think, I would use such a logic as a general heuristic instead of considering the cube counts. To give an example for a "typical dataset", a rather big one could be around 100000 × 300000 × 8000. Going for 100 voxel per edge would result in a target mag 1024, 4096, 128.
However, I would add the following constraints:
- X and Y should be equivalent (that's what we see 99% for existing datasets). Therefore, I'd simply consider `max(100000, 300000)` for x and y (--> 4096, 4096, 128).
- Additionally, I'd follow the anisotropic downsampling approach we usually do (see the code for that logic). Therefore, the method could simply output `Mag(4096)` as the max mag and combined with https://github.com/scalableminds/webknossos-cuber/issues/157, the z dimension would be downsampled according to the scale.

wkcuber/downsampling.py
```diff
@@ -34,22 +34,16 @@ def calculate_virtual_scale_for_target_mag(
...
-        int(math.pow(2, int(math.log(s, 2)))) for s in cubes_per_dim
-    ]  # highest power of 2 smaller than cubes_per_dim
+def calculate_default_max_mag(dataset_size: Tuple[int, int, int]) -> Mag:
+    # The lowest mag should have a size of ~ 100vx**3
+    max_x_y = max(dataset_size[0], dataset_size[1])
+    # highest power of 2 larger (or equal) than max_x_y divided by 100
     return Mag(
-        [min(max(m, 16), 512) for m in magnification]
-    )  # at least 16 and at most 512
+        max(int(math.pow(2, math.ceil(math.log(max_x_y / 100, 2)))), 4)
```
philippotto, line 45: why is the outer `int` needed? shouldn't 2**integer always be an integer? Also, you could use `2**(math.ceil(...))` instead of `math.pow` if you want.

wkcuber/downsampling.py
```diff
@@ -81,7 +75,9 @@ def create_parser() -> ArgumentParser:
         help="Max resolution to be downsampled. In case of anisotropic downsampling, the process is considered "
         "done when max(current_mag) >= max(max_mag) where max takes the largest dimension of the mag tuple "
         "x, y, z. For example, a maximum mag value of 8 (or 8-8-8) will stop the downsampling as soon as a "
-        "magnification is produced for which one dimension is equal or larger than 8.",
+        "magnification is produced for which one dimension is equal or larger than 8. "
+        "The default value is calculated depending on the dataset size. In the lowest Mag, the size will be "
```
philippotto, line 79: :+1:

wkcuber/downsampling.py
```diff
@@ -34,22 +34,16 @@ def calculate_virtual_scale_for_target_mag(
...
 
-def calculate_default_max_mag(
-    dataset_size: Tuple[int, int, int], cube_length: int
-) -> Mag:
-    cubes_per_dim = -(-np.array(dataset_size) // np.array(cube_length))
-    magnification = [
-        int(math.pow(2, int(math.log(s, 2)))) for s in cubes_per_dim
-    ]  # highest power of 2 smaller than cubes_per_dim
+def calculate_default_max_mag(dataset_size: Tuple[int, int, int]) -> Mag:
+    # The lowest mag should have a size of ~ 100vx**3
```
philippotto, line 41:
```suggestion
    # The lowest mag should have a size of ~ 100vx**2 per slice
```

wkcuber/downsampling.py
```diff
@@ -34,22 +34,16 @@ def calculate_virtual_scale_for_target_mag(
...
+    max_x_y = max(dataset_size[0], dataset_size[1])
+    # highest power of 2 larger (or equal) than max_x_y divided by 100
     return Mag(
```
philippotto, line 44:
```suggestion
    // The calculated factor will be used for x, y and z here. If anisotropic downsampling takes place,
    // the dimensions can still be downsampled independently according to the scale.
    return Mag(
```

philippotto (approved):
Cool :100: Looks very good :)

philippotto (approved):
Cool beans :rocket: Looks good to me! :)

PR #286, opened 2021-04-01
CZI support

youri-k:


wkcuber/image_readers.py
```diff
@@ -199,11 +208,128 @@ def read_dtype(self, file_name: str) -> str:
...
+            if (
+                entry.start[z_index] - z_file_start == z_slice
+                and entry.start[channel_index] - channel_file_start == channel_slice
+            ):
+                data = entry.data_segment().data()
+                if x_index > y_index:
+                    data = data.reshape(data.shape[y_index], data.shape[x_index])
+                    data = data.swapaxes(0, 1)
+                else:
+                    data = data.reshape(data.shape[x_index], data.shape[y_index])
```
philippotto, line 235: Can you add a comment which explains the reasoning behind this a bit?

philippotto (commented):
Great stuff! I think, this just needs a few additional comments to make it easier to understand the hard bits of the code :)

wkcuber/image_readers.py
youri-k: I have added some comments. The reasoning is pretty similar to the all channel case, but is easier because we only need to handle two dimensions.

wkcuber/image_readers.py
```diff
@@ -199,11 +208,128 @@ def read_dtype(self, file_name: str) -> str:
...
+    # return format will be (X, Y, channel_count)
+    def _read_array_all_channels(
+        self, czi_file: CziFile, dtype: np.dtype, z_slice: int
+    ) -> np.ndarray:
+        x_index = czi_file.axes.find("X")
+        y_index = czi_file.axes.find("Y")
+        z_index = czi_file.axes.find("Z")
+        c_index = czi_file.axes.find("C")
+        indices = [(x_index, "X"), (y_index, "Y"), (c_index, "C")]
+        indices.sort()
```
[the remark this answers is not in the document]
youri-k: The example file I used had the following axes: `VIHRSCTZYX0` with the following dimensions `(1, 1, 1, 1, 1, 2, 1, 624, 1920, 1920, 1)`. Therefore, I am not sure, whether there is a fix ordering or if all axes are always present.
To make the code work for all cases, I use the sorted indices to reshape this multi-dimensional data into 3 dimensional data. In this example, this would lead to `reshape(2, 1920,1920)` for each tile, which represents a single Z layer. Afterwards, I can use the sorted indices to transpose the data into our desired format of `X,Y,C`. 

I hope my comments are enough to explain this in the code, too.

wkcuber/image_readers.py
```diff
@@ -146,6 +147,14 @@ def find_count_of_axis(tif_file: TiffFile, axis: str) -> int:
         return tif_series.shape[index]  # pylint: disable=unsubscriptable-object
 
 
+def find_count_of_czi_axis(czi_file: CziFile, axis: str) -> int:
```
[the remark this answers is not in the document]
youri-k: Thanks for the idea. I didn't know that was possible. I refactored the `TiffImageReader` as well

wkcuber/image_readers.py
philippotto: Great, thanks for dealing with this complexity :+1:

philippotto (approved):
Awesome :)

Issue #277, opened 2021-03-04
Add upsampling


normanrz:
In some cases it is useful to upsample a dataset. For example, if you have a segmentation in 2-2-1, but also want to show it in 1-1-1. An upsampling script just like the downsampling one, would be nice.

Here is a sample script that does upsampling for a dataset from 2-2-1 to 1:
```python
import wkw
import numpy as np

CUBE_LENGTH = 1024
DATASET_SIZE = (5445, 8380, 3285)

in_ds = wkw.Dataset.open("segmentation/2-2-1")
out_ds = wkw.Dataset.open("segmentation/1", wkw.Header(voxel_type=in_ds.header.voxel_type, block_type=in_ds.header.block_type))

def zoom(in_buf):
  out_buf = np.zeros((CUBE_LENGTH, CUBE_LENGTH, CUBE_LENGTH), in_buf.dtype)
  for dx in (0, 1):
    for dy in (0, 1):
      out_buf[dx:CUBE_LENGTH:2, dy:CUBE_LENGTH:2, :] = in_buf
  return out_buf

for x in range(0, DATASET_SIZE[0], CUBE_LENGTH):
  for y in range(0, DATASET_SIZE[1], CUBE_LENGTH):
    for z in range(0, DATASET_SIZE[2], CUBE_LENGTH):
      print(f"Start {(x,y,z)}")
      in_buf = in_ds.read((x // 2, y // 2, z), (CUBE_LENGTH // 2, CUBE_LENGTH // 2, CUBE_LENGTH))[0]
      print(f"Read {(x,y,z)}")
      up_buf = zoom(in_buf)
      print(up_buf.dtype, up_buf.shape)
      out_ds.write((x, y, z), up_buf)
      print(f"Done {(x,y,z)}")
```

PR #287, opened 2021-04-08
Implement upsampling

rschwanhold:
- closes: #277

### DIFF
--- a/wkcuber/api/Dataset.py
+++ b/wkcuber/api/Dataset.py
@@ -11,6 +11,7 @@ import numpy as np
 import os
 import re
 
+from wkcuber.api.bounding_box import BoundingBox
 from wkcuber.mag import Mag
 from wkcuber.utils import logger, get_executor_for_args, ceil_div_np
 
@@ -344,22 +345,21 @@ class AbstractDataset(Generic[LayerT]):
                     largest_segment_id=largest_segment_id,
                 )
 
+                bbox = self.properties.get_bounding_box_of_layer(layer_name)
+
                 for mag_name, mag in layer.mags.items():
                     target_mag = target_layer.add_mag(mag_name)
 
                     # The bounding box needs to be updated manually because chunked views do not have a reference to the dataset itself
-                    bbox = mag.layer.dataset.properties.get_bounding_box_of_layer(
-                        layer_name
-                    )
                     # The base view of a MagDataset always starts at (0, 0, 0)
                     target_mag.view.global_offset = (0, 0, 0)
                     target_mag.view.size = cast(
                         Tuple[int, int, int],
                         tuple(
-                            ceil_div_np(
-                                np.array(bbox[0]) + np.array(bbox[1]),
-                                Mag(mag_name).as_np(),
-                            )
+                            BoundingBox(topleft=bbox[0], size=bbox[1])
+                            .align_with_mag(Mag(mag_name), ceil=True)
+                            .in_mag(Mag(mag_name))
+                            .bottomright
                         ),
                     )
                     target_mag.layer.dataset.properties._set_bounding_box_of_layer(
--- a/wkcuber/api/Layer.py
+++ b/wkcuber/api/Layer.py
@@ -417,7 +417,7 @@ class WKLayer(Layer[WKMagDataset]):
 
         try:
             with wkw.Dataset.open(
-                    find_mag_path_on_disk(self.dataset.path, self.name, mag)
+                find_mag_path_on_disk(self.dataset.path, self.name, mag)
             ) as wkw_dataset:
                 wk_header = wkw_dataset.header
 
@@ -429,7 +429,9 @@ class WKLayer(Layer[WKMagDataset]):
                 self.name, mag, cube_length=wk_header.block_len * wk_header.file_len
             )
         except wkw.WKWException:
-            logging.error(f"Failed to setup magnification {str(mag)}, which is specified in the datasource-properties.json")
+            logging.error(
+                f"Failed to setup magnification {str(mag)}, which is specified in the datasource-properties.json"
+            )
 
     def _initialize_mag_from_other_mag(
         self, new_mag_name: Union[str, Mag], other_mag: MagDataset, compress: bool
--- a/wkcuber/api/Layer.py
+++ b/wkcuber/api/Layer.py
@@ -48,6 +48,7 @@ from wkcuber.downsampling_utils import (
     downsample_cube_job,
     determine_buffer_edge_len,
 )
+from wkcuber.upsampling_utils import upsample_cube_job
 from wkcuber.mag import Mag
 from wkcuber.utils import (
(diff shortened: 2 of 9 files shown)


wkcuber/api/Dataset.py
```diff
@@ -344,22 +345,21 @@ def copy_dataset(
...
                         Tuple[int, int, int],
                         tuple(
-                            ceil_div_np(
-                                np.array(bbox[0]) + np.array(bbox[1]),
-                                Mag(mag_name).as_np(),
-                            )
+                            BoundingBox(topleft=bbox[0], size=bbox[1])
+                            .align_with_mag(Mag(mag_name), ceil=True)
+                            .in_mag(Mag(mag_name))
+                            .bottomright
```
rschwanhold, line 362: This does the same as before, but is better understandable.

wkcuber/api/Layer.py
```diff
@@ -167,41 +164,27 @@ def _pad_existing_mags_for_downsampling(
...
-        )
-        # translate alignment into mag 1
-        aligned_offset_in_mag1 = lowest_mag * offset_in_lowest_mag
-        aligned_size_in_mag1 = lowest_mag * (
-            end_offset_in_lowest_mag - offset_in_lowest_mag
+        bb_mag1 = BoundingBox(
+            topleft=self.dataset.properties.data_layers[
+                self.name
+            ].get_bounding_box_offset(),
+            size=self.dataset.properties.data_layers[self.name].get_bounding_box_size(),
```
rschwanhold, line 171: I realized during the hackathon that this can be optimized

wkcuber/api/Layer.py
```diff
@@ -293,6 +276,85 @@ def downsample(
...
+                job_count_per_log = math.ceil(
+                    1024 ** 3 / voxel_count_per_cube
+                )  # log every gigavoxel of processed data
+
+                if buffer_edge_len is None:
+                    buffer_edge_len = determine_buffer_edge_len(
+                        prev_mag_ds.view
+                    )  # DEFAULT_EDGE_LEN
+                func = named_partial(
+                    downsample_cube_job,
```
rschwanhold, line 335: The name of the method is not good. It was previously just used for downsampling but I reused it for upsampling. Do you have a better idea than `up_or_downsample_cube_job`?

wkcuber/api/Layer.py
```diff
@@ -353,17 +415,23 @@ def setup_mag(self, mag: Union[str, Mag]) -> None:
...
-        self.dataset.properties._add_mag(
-            self.name, mag, cube_length=wk_header.block_len * wk_header.file_len
-        )
+            self.mags[mag] = WKMagDataset(
+                self, mag, wk_header.block_len, wk_header.file_len, wk_header.block_type
+            )
+
+            self.dataset.properties._add_mag(
+                self.name, mag, cube_length=wk_header.block_len * wk_header.file_len
+            )
```
rschwanhold, line 430: Leaving this inside the `try`-block means that it will eventually remove the entry from the `datasource-properties.json`. Do you think this is okay?

wkcuber/downsampling_utils.py
```diff
@@ -263,23 +288,49 @@ def downsample_cube(
...
+        for dy in (0, factors[1] - 1):
+            for dz in (0, factors[2] - 1):
+                out_buf[
+                    dx : out_buf.shape[0] : factors[0],
+                    dy : out_buf.shape[1] : factors[1],
+                    dz : out_buf.shape[2] : factors[2],
+                ] = cube_buffer
+    return out_buf
+
+
```
rschwanhold, line 304: I implemented this like in your description of the issue (https://github.com/scalableminds/webknossos-cuber/issues/277). However, I also came across a branch where `from scipy.ndimage.interpolation import zoom` is used (https://github.com/scalableminds/webknossos-cuber/compare/upsampling). There is no need for an `InterpolationMode` because we only upsample by powers of 2, right?

wkcuber/api/Layer.py
```diff
@@ -293,6 +276,85 @@ def downsample(
     def setup_mag(self, mag: Union[str, Mag]) -> None:
         pass
 
+    def upsample(
+        self,
+        from_mag: Mag,
+        min_mag: Optional[Mag],
+        compress: bool,
+        anisotropic: Optional[bool] = None,
+        scale: Optional[Tuple[float, float, float]] = None,
```
normanrz, line 285: Why is `scale` an argument? Shouldn't that be part of the dataset properties?

wkcuber/api/Layer.py
normanrz: I don't understand yet. In what cases would the entry be removed?

wkcuber/api/Layer.py
normanrz: I think `up_or_downsample_cube_job` would be fine. However, I am not sure if it would be better to just make 2 functions, which would get rid of the `upsample` flag and some if/else branches.

wkcuber/downsampling_utils.py
normanrz: Yeah. We are only upsampling by powers of 2 (including 1). I experimented with `zoom`, but it was way slower than implementing it with numpy directly.

wkcuber/api/Layer.py
rschwanhold: When opening an existing dataset, the metadata is stored in an `Properties` object. These properties are used to write the `datasource-properties.json` to disk if necessary. 
If an `WKDataset` is opened, which has e.g. an entry for `Mag(2)` but this mag does not exist on disk (this might be the case if the directory is deleted, but the properties are not updated), this implementation  would fail in the `try` block. Therefore, the entry would not be added to the properties object in RAM. The next time the properties are dumped to disk (e.g. because the bounding box changed), the file would be overwritten (this would "remove" the previously existing entry of `Mag(2)` from the properties on disk)

wkcuber/api/Layer.py
rschwanhold: Yes your right. I have updated it.
The code was initially copied from `downsampling` which still has the parameter 'scale' (optionally) for backwards compatibility

wkcuber/api/Layer.py
normanrz: Yeah. I think that is ok.

wkcuber/api/Layer.py
normanrz: @philippotto do you have a take on that?

philippotto: regarding the name, "resample_cube_job" comes to mind. didn't look into whether it would be better to have two distinct functions, though.

philippotto: Or was the question directed towards the latter? I think @rschwanhold can decide this better, since he worked with the code.

normanrz: Ok. fine with me.

normanrz (approved):

Issue #283, opened 2021-03-19
Support for RGB TIFF files


youri-k:
The `TiffFile` library seems to differentiate between standard channel count (denoted as `C`) and a different channel count for rgb files (denoted as `S`). We should support both formats and possibly use the `C` count as a hint for different layers for wK.

fm3:
Turns out the old version of the cuber didn’t support this case either (though it had a slightly prettier error message). So It’s not a regression. It would still be a useful feature, so let’s keep the Issue

PR #292, opened 2021-04-15
Support RGB TIFF

youri-k:
I have added a fix to support the `S` channel count. Currently, I have not implemented supported for mixed `C` and `S` channel count. Let me know if this is necessary.

fixes: #283


wkcuber/image_readers.py
```diff
@@ -150,34 +154,54 @@ def find_count_of_axis(tif_file: TiffFile, axis: str) -> int:
...
+                        map(
+                            lambda x: x.asarray(), tif_file.pages[0 : self.num_channels]
+                        )
+                    ),
                     dtype,
                 )
+
+            # if the pages are multi-channel, then we'll have 4 dimensions here because of [x:x+1] notation, so we reshape the data
+            if self.is_page_multi_channel:
+                data = data.reshape((data.shape[1], data.shape[2], data.shape[3]))
```
rschwanhold, line 193: Is this the same as `data = data[0]`?

wkcuber/image_readers.py
```diff
@@ -150,34 +154,54 @@ def find_count_of_axis(tif_file: TiffFile, axis: str) -> int:
...
+                x_index = tif_file.pages[0].axes.find("X")
+                y_index = tif_file.pages[0].axes.find("Y")
+                c_index = tif_file.pages[0].axes.find("S")
```
rschwanhold, line 196: Can the axis also be `C` (instead of `S`)? 
Consider renaming the variable to `s_index`.

wkcuber/image_readers.py
```diff
@@ -150,34 +154,54 @@ def find_count_of_axis(tif_file: TiffFile, axis: str) -> int:
...
-            num_channels = self.read_channel_count(file_name)
-            if len(tif_file.pages) > num_channels:
+            if self.num_channels is None:
+                self.num_channels = self.read_channel_count(file_name)
+            if self.is_page_multi_channel is None:
+                # we assume all tif pages have the same dimensions
+                self.is_page_multi_channel = tif_file.pages[0].ndim == 3
+
+            if self.is_page_multi_channel:
+                self.num_channels = 1
```
rschwanhold, line 164: Why are `num_channels` and `is_page_multi_channel` stored as an instance variable?

rschwanhold (commented):
Thanks for the quick PR. I have left some comments

wkcuber/image_readers.py
youri-k: Yes, you are right ^^

wkcuber/image_readers.py
youri-k: All example files always used the `S` index if a page was multi-channel. That's why I assume it here. To use the same variable in both if-branches, I would leave as `c_index`, if that's fine for you

wkcuber/image_readers.py
youri-k: I wanted to "cache" these values for future calls.

wkcuber/image_readers.py
rschwanhold: Ok, maybe I misunderstood it. If the axis are `XYC`, this code still works even though you are not checking for `C`, right?

youri-k: Right because as far as I know if `self.is_page_multi_channel` is `true`, there is no `C`

wkcuber/image_readers.py
rschwanhold: I am not as deep in the `ImageReader` as you are, but my understanding is that there is a single instance for managing the image readers (`image_reader = ImageReaderManager()`).  The manager has a dict with an instance for each different type (e.g. `TiffImageReader`). If I now want to read two completely different images (both .tiff), wouldn't this be a problem because the `num_channels` could differ?

youri-k: Yes, but for the cubing case we assume that all files have similar dimensions.
If we want to use the cubing for two different datasets, the cuber is executed twice, so that shouldn't be a problem. 

Are there places, where only parts of the cuber code are executed?

rschwanhold: I only found the `image_reader` in connection with `cubing`. Then it should be fine.

rschwanhold (approved):
Thanks a lot. 
If you want you can add a test case for the new functionality.

Issue #260, opened 2021-01-06
Support pathlib Path in api calls


fm3:
A lot of api calls are typed to expect str where Path could also be valid. We should either expect Path entirely or allow their union

I had to make my Paths str for
- BufferedSliceWriter
- write_webknossos_metadata
- refresh_metadata
- downsample_mags_anisotropic

PR #293, opened 2021-04-15
Support pathlib Path in api calls

rschwanhold:
## Description
- use `pathlib.Path` for api calls

## Things I have noticed
- even though the cuber is typed, there are cases where the typing does not catch the mistake (e.g. if a function that now requires a `Path` is called like `refresh_metadata(args.target_path)` instead of `refresh_metadata(Path(args.target_path))`)
- I left the type of a few parameters as `str`. For example `find_files(source_path: str, extensions: Iterable[str]) -> Generator[str, Any, None]` because `iglob` requires a `str` and a path with `*` is more like a pattern than an actual `Path`
- `wkw.Dataset.open(...)` requires a `str`. For consistency, I also left `KnossosDataset.open(root: str, ...)` as it is

fixes: #260


rschwanhold:
I changed the types to `Path` instead of `Union[str, Path]`. Do you think this is the right thing to do?

tests/test_dataset.py
```diff
@@ -48,7 +48,7 @@
 expected_error_msg = "The test did not throw an exception even though it should. "
 
 
-def delete_dir(relative_path: str) -> None:
+def delete_dir(relative_path: Union[str, Path]) -> None:
```
fm3, line 51: is str still needed here?

tests/test_dataset.py
```diff
@@ -167,7 +167,7 @@ def get_multichanneled_data(dtype: type) -> np.ndarray:
 def test_create_wk_dataset_with_layer_and_mag() -> None:
     delete_dir("./testoutput/wk_dataset")
 
-    ds = WKDataset.create("./testoutput/wk_dataset", scale=(1, 1, 1))
+    ds = WKDataset.create(Path("./testoutput/wk_dataset"), scale=(1, 1, 1))
```
fm3, line 170: The `./` is probably redundant (?)

fm3: Also, it might make sense to store "testoutput" and "testdata" in constants in this file, rather than include it in so many strings. e.g. `TEST_OUTPUT_DIR = Path("testoutput")` and then down here `create(TEST_OUTPUT_DIR / "wk_dataset", scale=…)`

tests/test_dataset.py
```diff
@@ -1356,24 +1353,20 @@ def test_writing_subset_of_compressed_data_multi_channel() -> None:
...
-        write_data1
-    )
+    WKDataset.create(Path("./testoutput/compressed_data"), scale=(1, 1, 1)).add_layer(
+        "color", Layer.COLOR_TYPE, num_channels=3
+    ).add_mag("1", block_len=8, file_len=8).write(write_data1)
 
     # compress data
     compress_mag_inplace(
-        os.path.abspath("./testoutput/compressed_data/"),
+        Path(os.path.abspath("./testoutput/compressed_data/")),
```
fm3, line 1362: We could get rid of the os.path here by changing this to `(TEST_OUTPUT_DIR / "compressed_data").resolve()` (assuming my suggestion from above is implemented).
Assuming that the specific symlink behavior of abspath is not needed here? Compare my notes on abpath in the description here https://github.com/scalableminds/voxelytics/pull/2175

wkcuber/compress.py
```diff
@@ -164,4 +167,6 @@ def compress_mags(
 if __name__ == "__main__":
     args = create_parser().parse_args()
     setup_logging(args)
-    compress_mags(args.source_path, args.layer_name, args.target_path, args.mag, args)
+    compress_mags(
+        Path(args.source_path), args.layer_name, Path(args.target_path), args.mag, args
```
fm3, line 171: I believe argparse supports `type=Path` so we may be able to go without these conversions. Not sure if that creates incompabitilies with other parts of the wkcuber, though

tests/test_dataset.py
fm3: Or more radically, @jstriebel do you think we should try to avoid all `/` occurrences in strings and rather always use pathlibs `/` operator?

fm3 (changes requested):
Great stuff, thanks for taking the time to implement all these changes!

I would like to ask for a second opinion by @jstriebel on a few of your questions:

> I changed the types to Path instead of Union[str, Path]. Do you think this is the right thing to do?

I personally do think that Path only is right, but I might be persuaded otherwise. What do you think @jstriebel @normanrz ?

> even though the cuber is typed, there are cases where the typing does not catch the mistake (e.g. if a function that now requires a Path is called like refresh_metadata(args.target_path) instead of refresh_metadata(Path(args.target_path)))

Does one of you have an idea why mypy doesn’t catch the error here? Could it be that argparse does not report its type properly?

> I left the type of a few parameters as str. For example find_files(source_path: str, extensions: Iterable[str]) -> Generator[str, Any, None] because iglob requires a str and a path with * is more like a pattern than an actual Path.

Could https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob be an option there?

> wkw.Dataset.open(...) requires a str. For consistency, I also left KnossosDataset.open(root: str, ...) as it is

Seems fair, assuming that this is about the same level of abstraction?

tests/test_dataset.py
jstriebel: I think so, yes, to make this (more) Windows-compatible. This could simply be
```suggestion
    ds = WKDataset.create(Path("testoutput", "wk_dataset"), scale=(1, 1, 1))
```

or (my personal preference)
```suggestion
    ds = WKDataset.create(Path("testoutput") / "wk_dataset", scale=(1, 1, 1))
```

jstriebel:
> > I changed the types to Path instead of Union[str, Path]. Do you think this is the right thing to do?
> 
> I personally do think that Path only is right, but I might be persuaded otherwise. What do you think @jstriebel @normanrz ?

I'd use `Union[str, Path]` for paths iff the API (not the CLI interface) is used outside of scm often.

> > even though the cuber is typed, there are cases where the typing does not catch the mistake (e.g. if a function that now requires a Path is called like refresh_metadata(args.target_path) instead of refresh_metadata(Path(args.target_path)))
> 
> Does one of you have an idea why mypy doesn’t catch the error here? Could it be that argparse does not report its type properly?

Exactly, `args` is basically a `Dict[str, Any]`, that allows to access everything via attributes. I'd suggest to use [typer](https://typer.tiangolo.com/) instead (in a seperate PR), I found it quite useful in vx already.

> > I left the type of a few parameters as str. For example find_files(source_path: str, extensions: Iterable[str]) -> Generator[str, Any, None] because iglob requires a str and a path with * is more like a pattern than an actual Path.
> 
> Could https://docs.python.org/3/library/pathlib.html#pathlib.Path.glob be an option there?

Only if we represent the pattern as a `Path` + pattern-`str`, so we could do `path.glob(pattern)`. I agree that a pattern is not really a `Path` object, since most methods would be useless on it. Accepting a string and converting the glob results as early as possible to Paths would be my current choice, but I'm not really confident about this solution.

> > wkw.Dataset.open(...) requires a str. For consistency, I also left KnossosDataset.open(root: str, ...) as it is
> 
> Seems fair, assuming that this is about the same level of abstraction?

I'd go for `Union[Path, str]` here, but no strong opinion.

normanrz:
> I'd use `Union[str, Path]` for paths iff the API (not the CLI interface) is used outside of scm often.

I don't think this is the case atm

tests/test_dataset.py
rschwanhold: No

wkcuber/compress.py
rschwanhold: Yes that works

fm3 (approved):
Excellent :tada: 

We should make sure that the release is not a minor version only, since it contains breaking changes in the API

normanrz:
We are still at `0.x.x` where breaking changes are allowed in minor versions. I would not go to `1.x.x` for this change.

PR #294, opened 2021-04-16
[wkcuber] Export wkw as nifti

leowe:
For the stroke-vision project, it Is necessary to convert the predictions and segmentations to nifti files.
- add export_wkw_as_nifti script which converts a wkw dataset to nifti files
- add unit test for export script

For now, it doesn't have to be robust in every way, it only needs to serve this use case for now.


wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,132 @@
...
+    mag: Mag,
+    destination_path: Path,
+    name: str,
+) -> None:
+    wk_ds = WKDataset(wkw_file_path)
+    layers = wk_ds.layers
+
+    for layer_name in layers:
+        if bbox is None:
+            _, _, bbox_dim, origin = read_metadata_for_layer(wkw_file_path, layer_name)
```
philippotto, line 95: I think these properties should be available via `wk_ds.layers`. @rschwanhold might be able to give more precise hints :)

wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,132 @@
...
+    layer_name: str,
+    destination_path: Path,
+    name: str,
+) -> None:
+    wk_ds = WKDataset(wkw_file_path)
+    layer = wk_ds.get_layer(layer_name)
+    mag_layer = layer.get_mag(mag)
+    data = mag_layer.read(bbox["topleft"], bbox["size"])
+    data = data.transpose(1, 2, 3, 0)
+    # Problem: different dtypes (uint8 and uint32), implicit cast
```
philippotto, line 70: what's up with this comment?

philippotto (commented):
Looks cool :+1: Only added some nitpicks.

wkcuber/export_wkw_as_nifti.py
rschwanhold: You can use `wk_ds.properties.get_bounding_box_of_layer(...)`

hotzenklotz:
@leowe PING

wkcuber/export_wkw_as_nifti.py
rschwanhold: FYI: The Dataset API was changed a lot in the recent past. You can now get the bbox of a layer using `Layer.bounding_box`

leowe: @rschwanhold is this still up-to-date? I can't find this function in the `properties` source code. If it is not, what is the right way to do it?

leowe:
I added two flags that allow for padding and later cropping and incorporated the comments from the previous review.

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+    parse_bounding_box,
+    setup_logging,
+)
+
+
+def create_parser() -> ArgumentParser:
+    parser = ArgumentParser()
+
+    parser.add_argument(
+        "--source_path", "-s", help="Directory containing the wkw file.", required=True
```
philippotto, line 24:
```suggestion
        "--source_path", "-s", help="Directory containing the wkw file(s).", required=True
```

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+    )
+
+    parser.add_argument(
+        "--destination_path",
+        "-d",
+        help="Output directory for the generated nifti files.",
```
philippotto, line 30:
```suggestion
        help="Output directory for the generated nifti file.",
```

since the script only outputs one nifti file, right?

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+        required=True,
+    )
+
+    parser.add_argument("--name", "-n", help="Name of the nifti", default="")
+
+    parser.add_argument(
+        "--bbox",
```
philippotto, line 37: maybe rename to `--source_bbox`?

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+    parser.add_argument(
+        "--mag", "-m", help="The magnification that should be read", default=1
+    )
+
+    parser.add_argument(
+        "--downsample", help="Downsample each nifti image", default=1, type=int
+    )
+
+    parser.add_argument(
+        "--original_bbox",
```
philippotto, line 54:
```suggestion
        "--padded_bbox",
```

would fit better imo, if I understand its influence correctly.

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+                padding_per_axis.append((0, 0))
+            else:
+                left_pad = offset_into_orginal_bbox[i]
+                right_pad = (
+                    original_bbox_size[i] - offset_into_orginal_bbox[i] - data.shape[i]
+                )
+                padding_per_axis.append((left_pad, right_pad))
+
+        padding_per_axis.append((0, 0))
+        print(padding_per_axis)
```
philippotto, line 117:
```suggestion
```

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+        "-d",
+        help="Output directory for the generated nifti files.",
+        required=True,
+    )
+
+    parser.add_argument("--name", "-n", help="Name of the nifti", default="")
+
+    parser.add_argument(
+        "--bbox",
+        help="The bounding box of which the nifti file should be generated."
```
philippotto, line 38:
```suggestion
        help="The bounding box in the nifti file from which data is read."
```

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+        "--mag", "-m", help="The magnification that should be read", default=1
+    )
+
+    parser.add_argument(
+        "--downsample", help="Downsample each nifti image", default=1, type=int
+    )
+
+    parser.add_argument(
+        "--original_bbox",
+        help="Easy way to add padding to the generated file. Format is: x,y,z,width,height,depth"
```
philippotto, line 55:
```suggestion
        help="After reading from --source_bbox in the input file, the data is zero-padded to fit --padded_bbox. --source_bbox must be contained in --padded_bbox Format is: x,y,z,width,height,depth"
```

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+        "x,y,z is the offset of the wkw layer into the final bounding box; "
+        "width,height,depth corresponds to final dimensions",
+        default=None,
+        type=parse_bounding_box,
+    )
+
+    parser.add_argument(
+        "--bounding_box_crop",
+        help="Easy way to crop the final file. Is applied AFTER everything else (including padding)."
+        "Format is: x,y,z,width,height,depth",
```
philippotto, line 65:
```suggestion
        "--cropping_bbox",
        help="After reading (and potentially padding) the input data, the output is cropped according to --cropping_bbox. --cropping_bbox must be contained in --padded_bbox"
        "Format is: x,y,z,width,height,depth",
```

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+    if original_bbox_size and offset_into_orginal_bbox:
+        padding_per_axis = []
+
+        for i in range(3):
+            if original_bbox_size[i] == data.shape[i]:
+                padding_per_axis.append((0, 0))
+            else:
+                left_pad = offset_into_orginal_bbox[i]
+                right_pad = (
+                    original_bbox_size[i] - offset_into_orginal_bbox[i] - data.shape[i]
```
philippotto, line 112: the containment relationships I added to the argparse help should also be validated here. you can use `BoundingBox.contains_bbox` for that

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+    bbox: Dict,
+    mag: Mag,
+    layer_name: str,
+    destination_path: Path,
+    name: str,
+    original_bbox_size: Optional[List[int]] = None,
+    offset_into_orginal_bbox: Optional[List[int]] = None,
+    bounding_box_crop: Optional[List[int]] = None,
+) -> None:
+    wk_ds = Dataset(wkw_file_path)
```
philippotto, line 87:
```suggestion
    dataset = Dataset(wkw_file_path)
```

fits better to our existing naming scheme

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,206 @@
...
+
+    for layer_name, layer in wk_ds.layers.items():
+        if bbox is None:
+            bbox_dict = {
+                "topleft": layer.bounding_box.topleft,
+                "size": layer.bounding_box.size,
+            }
+        else:
+            bbox_dict = {"topleft": list(bbox.topleft), "size": list(bbox.size)}
```
philippotto, line 161: why do you convert the BoundingBox instance to a dict? Passing around the object seems better to me. It also provides all the necessary information plus helpful methods, such as `contains_bbox`.

philippotto (changes requested):
Great stuff and features :+1: Code looks really tidy, too :1st_place_medal:
I mostly left feedback for the bounding box handling (usability-wise and implementation-wise).

wkcuber/wkcuber/export_wkw_as_nifti.py
leowe: No, it actually writes one file per layer. I updated the description to reflect that.

wkcuber/wkcuber/export_wkw_as_nifti.py
leowe: It probably made sense a long time ago. Changed it now.

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,201 @@
...
+                cropping_bbox
+            ), "padded_bbox should contain cropping_bbox"
+
+            logging.info(f"Using Bounding Box {cropping_bbox}")
+
+            data = data[
+                cropping_bbox.topleft[0] : cropping_bbox.bottomright[0],
+                cropping_bbox.topleft[1] : cropping_bbox.bottomright[1],
+                cropping_bbox.topleft[2] : cropping_bbox.bottomright[2],
+            ]
```
philippotto, line 142:
```suggestion
            data = cropping_bbox.slice_array(data)
```

wkcuber/wkcuber/export_wkw_as_nifti.py
```diff
@@ -0,0 +1,201 @@
...
+        for i in range(3):
+            if padded_bbox.size[i] == data.shape[i]:
+                padding_per_axis.append((0, 0))
+            else:
+                left_pad = data_bbox_relative_to_pad_bbox.topleft[i]
+                right_pad = (
+                    pad_bbox_as_origin.size[i]
+                    - data_bbox_relative_to_pad_bbox.bottomright[i]
+                )
+                padding_per_axis.append((left_pad, right_pad))
```
philippotto, line 126: To be honest, I'm a bit lost with the semantics of the `--padded_bbox` parameter :see_no_evil: I thought, the padding would simply be the difference between `--source_bbox` and the `--padded_bbox`. My mental model is as follows:

image


However, your code does something different and I can't really figure out what the semantics are. The topleft of the `padded_bbox` seems to be simply used for left padding which could also be a way of designing the parameter it, but (1) then I would simply call the parameter `--padding` and make its format `(leftpadx, leftpady, leftpadz, rightpadx, rightpady, rightpadz)` and (2) right_pad doesn't fit to that semantics.

Maybe I'm missing something here, but I think it's easier to talk about a concrete example:

```
- source_bbox goes from 25, 25, 25   to   55, 55, 55
- padded_bbox goes from 10, 10, 10   to   70, 70, 70

- then, leftpad would currently be 10, but I'd expect 15
- rightpad would be 60 - (55 - (25 - 10)) = 20, but I'd expect 15
```

wkcuber/wkcuber/export_wkw_as_nifti.py
philippotto: This means that the cropping bbox is relative to the padded data. I would have thought it would be easier to use if all three bounding boxes are specified in the same global coordinate system by the user (see my mental model picture in the other comment). But maybe I'm wrong and it's easier to use the way you designed it. If you stick to that semantics, please update the argparse help to reflect the relativity. Also, the comments about the containment-relationships are then incorrect (and not useful) probably.

philippotto (changes requested):
Sorry to ping-pong here, but I think it's important to get the CLI interface right. Otherwise, we will have some headaches in the future. Something seems off with the bbox parameters currently. Either the implementation needs to be adapted or the argparse comments :see_no_evil:

wkcuber/wkcuber/export_wkw_as_nifti.py
leowe: Yes you are right, we need to get that right. I also stumbled over the semantics and what the best approach would be. I think I designed this interface in march to just throw in the bounding boxes that I used to transform the nifti to wkw. Simply using `--padding` would probably make this interface much easier. When I need to use it again later, I just add the necessary logic on the client side to calculate the necessary padding, this will also be easier if the semantics here are crystal clear.

leowe:
I just removed all flags and only added an `--padding` option to simplify the interface.

philippotto (approved):
Awesome :)

PR #295, opened 2021-04-22
Downsample single Mag

rschwanhold:
The PR #259 changed the implementation of `downsampling`. The old implementation had a function to downsample a singe mag (additionally to downsampling until `max_mag`). This function was not used in `wkcuber` and therefore removed/restructured in the new implementation. However `voxelytics` uses this method. Therefore, this PR restructures the new implementation a little bit to also support downsampling of a single mag.


philippotto (approved):
LGTM! :rocket:

Issue #288, opened 2021-04-08
Add compress function to high-level API


normanrz:
I think one of the last missing core functions of the high-level API is the ability to compress a mag-layer, and maybe even to write to compressed layers.

rschwanhold:
Writing to compressed layers is already supported for `WKDataset` (`https://github.com/scalableminds/webknossos-cuber/blob/master/wkcuber/api/MagDataset.py`). The `write` function has the parameter `allow_compressed_write`. In that case the necessary data is read to pad the data. This is not necessary if the data is already aligned with the boundingboxes of the files on disk.

normanrz:
That sounds great!

I think what I was looking for is a method to compress an existing mag-layer in-place similar to https://github.com/scalableminds/webknossos-cuber/blob/master/wkcuber/compress.py with an API like
```python
mag1 = ds.get_layer("color").get_mag("1")
# mag1 is uncompressed

mag1.compress()
# mag1 is compressed now

mag1.read(...)
```

PR #296, opened 2021-04-22
Add compression to Dataset API

rschwanhold:
The dataset API can now compress `MagDataset` (this is currently only implemented for `WKMagDataset`), by calling `.compress()`. This compresses the data in-place. Optionally a `target_path` can be specified.

fixes: #288

### DIFF
--- a/wkcuber/api/MagDataset.py
+++ b/wkcuber/api/MagDataset.py
@@ -233,7 +233,9 @@ class MagDataset:
     def _extract_file_index(self, file_path: Path) -> Tuple[int, int, int]:
         raise NotImplementedError
 
-    def compress(self, target_path: Path = None, args: Namespace = None) -> None:
+    def compress(
+        self, target_path: Union[str, Path] = None, args: Namespace = None
+    ) -> None:
         pass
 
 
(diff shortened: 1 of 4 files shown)


wkcuber/api/MagDataset.py
```diff
@@ -267,6 +280,64 @@ def _extract_file_index(self, file_path: str) -> Tuple[int, int, int]:
...
+
+        logging.info("Mag {0} successfully compressed".format(self.name))
+
+        if not was_opened:
+            self.close()
+
+        if target_path is None:
+            shutil.rmtree(uncompressed_full_path)
+            shutil.move(str(compressed_full_path), uncompressed_full_path)
+            shutil.rmtree(compressed_path)
```
normanrz, line 336: Can't you do this with a single move?

rschwanhold: `shutil.move` can only replace the target, if the target is a file. `uncompressed_full_path` is a directory, therefore, `shutil.rmtree(uncompressed_full_path)` is necessary. Deleting `compressed_path` is also necessary because otherwise this would leave an empty folder behind

normanrz: alright! thanks for clarifying

normanrz (approved):

Issue #298, opened 2021-04-27
Default args for Layer#downsample


normanrz:
The `Layer#downsample` method should work without the requirement of any args, because they can be either inferred or have sensible defaults.

```
    def downsample(
        self,
        from_mag: Mag, # Highest currently available mag
        max_mag: Optional[Mag], # nice!
        interpolation_mode: str, # See https://github.com/scalableminds/webknossos-cuber/blob/517e1e9f00c6c4f977f507efe7006b232175971c/wkcuber/downsampling_utils.py#L114
        compress: bool = True, # True
        anisotropic: Optional[bool] = None,
        scale: Optional[Tuple[float, float, float]] = None,
        buffer_edge_len: Optional[int] = None,
        args: Optional[Namespace] = None,
    ) -> None:
```

Issue #297, opened 2021-04-27
Detect max downsampling


normanrz:
For the default cubing, a downsampling until mag32 is preset. I would recommend that this value is set dynamically. For example, a dataset that only is 128x128x128vx doesn't need to be downsampled to 32 (i.e. 4x4x4vx). I would propose to downsample a dataset until the highest mag is ≤128x128x128vx in size.

normanrz:
See https://github.com/scalableminds/webknossos-cuber/blob/517e1e9f00c6c4f977f507efe7006b232175971c/wkcuber/downsampling_utils.py#L105

PR #304, opened 2021-04-29
Use default values for downsampling in Dataset API

rschwanhold:
### Description:
- Previously `from_mag` and `interpolation_mode` had no default values.
- The interface of the `downsample`-functions was redesigned to be more intuitive. 
- A changelog was added + the PR template was updated

### Issues:
fixes: #298 
fixes: #305 
fixes: #314 
fixes: #297

### Todos:
 - [x] Updated Changelog


wkcuber/api/Layer.py
```diff
@@ -199,30 +199,36 @@ def _pad_existing_mags_for_downsampling(
...
+        compress: bool = True,
         anisotropic: Optional[bool] = None,
         scale: Optional[Tuple[float, float, float]] = None,
         buffer_edge_len: Optional[int] = None,
         args: Optional[Namespace] = None,
     ) -> None:
+        if from_mag is None:
+            assert (
+                len(self.mags.keys()) > 0
+            ), "Failed to downsample data. The mag exists that could be used as 'from_mag'"
```
philippotto, line 214:
```suggestion
            ), "Failed to downsample data, because no existing mag was found."
```

wkcuber/api/Layer.py
```diff
@@ -199,30 +199,36 @@ def _pad_existing_mags_for_downsampling(
...
+            from_mag = Mag(min(self.mags.keys()))
```
philippotto, line 215: Why `min`? Shouldn't it be `max` to get the `Highest currently available mag` mentioned in #298.

wkcuber/api/Layer.py
```diff
@@ -199,30 +199,36 @@ def _pad_existing_mags_for_downsampling(
 
     def downsample(
         self,
-        from_mag: Mag,
-        max_mag: Optional[Mag],
-        interpolation_mode: str,
-        compress: bool,
+        from_mag: Optional[Mag] = None,
+        max_mag: Optional[Mag] = None,
```
philippotto, line 203: I'm still a bit confused about the semantics of the following parameters:

- max_mag
- anisotropic
- scale

I think the major use case for `max_mag` is to tell the cuber to stop after magnification X. However, it can also be used to determine the exact target mag (i.e., it prescribes the exact downsampling strategy) *if* `anisotropic` is set to True, correct? In that case a provided `scale` is overridden (I think, there should be a warning or maybe even a raised assertion for that).
I'm not sure the latter behavior of `max_mag` is intuitive. Maybe this should be a separate parameter? (e.g., `target_mag`).

What happens if I explicitly set `anisotropic` to `False` ? If I'm not mistaken, this will still use the DS scale and create anisotropic magnifications which is probably not what one would expect when passing `False`.

Regarding the scale: What is the usecase for providing this if the dataset's metadata scale exists? If one wants to mess with the downsampling steps, it's probably easier to pass a target mag.

My suggestion would be:
- let's make sure the parameters all have a helpful and easy-to-understand description which also mentions interdependencies between the parameters
- if it's possible to pass incompatible parameters (e.g., pass a scale, but also pass a target mag?), throw an error
- make sure the interface as a whole makes sense. maybe we can simplify some things. you probably have a better picture of this in your head than me :)

philippotto (changes requested):
Nice! Looking forward to a streamlined downsample-interface :rocket: However, I think that the interface still can use some work. At least, the parameters should have good comments to explain what they do. And if I'm not mistaken, there are still some unintuitive things happening if one passes certain parameter combinations.

wkcuber/api/Layer.py
rschwanhold: You are right. Here is an updated version:
```
      if anisotropic is not None:
          if anisotropic:
              assert scale is not None  # The user has to specify an anisotropic scale
          else:
              scale = (2, 2, 2)
      else:
          if scale is None:
              scale = self.dataset.properties.scale
``` 

However, this raises the question if `anisotropic` as a parameter is really necessary. What if we remove `anisotripic` completely?  If the dataset has an anisotropic scale, but the user want to downsample isotropic, then he has to specify `scale=(2,2,2)`
```
if scale is None:
    scale = self.dataset.properties.scale
# otherwise the parameter is used
```

philippotto (changes requested):
Great stuff! Left some suggestions, though :)

wkcuber/api/Layer.py
```diff
@@ -313,6 +329,46 @@ def downsample_mag(
...
+        interpolation_mode: str = "default",
+        compress: bool = True,
+        buffer_edge_len: Optional[int] = None,
+        args: Optional[Namespace] = None,
+    ) -> None:
+        assert (
+            from_mag.to_layer_name() in self.mags.keys()
+        ), f"Failed to downsample data. The from_mag ({from_mag}) does not exist."
+
+        target_mags = sorted(target_mags, key=lambda m: m.to_array())
```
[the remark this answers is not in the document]
rschwanhold: The `__lt__` operator only compares the largest values of the `Mags`:

```
>>> [m.as_np() for m in sorted([Mag(1), Mag(2), Mag([1, 1, 2])])]
[array([1, 1, 1]), array([2, 2, 2]), array([1, 1, 2])]

>>> [m.as_np() for m in sorted([Mag(1), Mag(2), Mag([1, 1, 2])], key=lambda m: m.to_array())]
array([1, 1, 1]), array([1, 1, 2]), array([2, 2, 2])]
```

wkcuber/downsampling_utils.py
```diff
@@ -14,6 +14,12 @@
 from wkcuber.utils import time_start, time_stop
 
 
+class SamplingModes:
+    AUTOMATIC = "automatic"
+    ISOTROPIC = "isotropic"
+    FIX_Z = "fix_z"
```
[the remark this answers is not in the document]
philippotto: according to the internal discussion, I'd say we should switch to "auto" and "constant_z".

philippotto (approved):
Excellent! Looking forward to use this myself :)

PR #307, opened 2021-04-29
Allow path to be of type str for public facing api

### LINKED ISSUE #301: Convenience for wkcuber.api
To open/create a dataset with the cool new high-level API the following code is required:

```python
from wkcuber.api.Dataset import WKDataset
from pathlib import Path

ds1 = WKDataset.create(Path("path") / "to" / "dataset1", scale=(128,128,128))
ds2 = WKDataset.open(Path("path") / "to" / "dataset2")

```

For one-off scripts, I think that could be a bit more convenient, if we had an API like this

```python
from wkcuber import WKDataset

ds1 = WKDataset.create("path/to/dataset1", scale=(128, 128, 128))
ds2 = WKDataset.open("path/to/dataset2")
```

Any thoughts? @rschwanhold @jstriebel @philippotto

rschwanhold:
fixes #301

### DIFF
--- a/wkcuber/api/Dataset.py
+++ b/wkcuber/api/Dataset.py
@@ -89,7 +89,8 @@ LayerT = TypeVar("LayerT", bound=Layer)
 
 class AbstractDataset(Generic[LayerT]):
     @abstractmethod
-    def __init__(self, dataset_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


philippotto (approved):
Awesome! This should also make the dependency update within vx easier :)

Issue #299, opened 2021-04-27
float/double instead of float32/float64 in WKDataset


normanrz:
webknossos expects the datatypes to be called `float` and `double` instead of `float32` or `float64` in the datasource-properties.json.

See https://github.com/scalableminds/webknossos-cuber/issues/100

rschwanhold:
If a color layer has 3 channel and `uint8` as the dtype (`dtype_per_channel`) then `uint24` is stored in the `datasource-properties.json`. How does this translate to floating points? 
Lets say I have a layer with 1 channel and I specify the dtype to be `float`: So then the `datasource-properties.json` should contain `float`, right? But what if the layer has 3 channel (should it then be `float96` or just `float`)?

normanrz:
wk only supports a subset of the what wkw could support. Multi-channel floats are not supported. See https://github.com/scalableminds/webknossos/blob/5777d8334386bd3716db8024beee9da36c861962/webknossos-datastore/app/com/scalableminds/webknossos/datastore/models/datasource/DataLayer.scala#L30-L42

PR #316, opened 2021-05-09
Webknossos compatible dtype in properties

rschwanhold:
Webknossos interprets `float` as `float32` and `double` as `float64`. Since `double` does not exist in python, `float` is interpreted as `float64`. This PR interprets `float` as `float32` and exports the dtype in a way that is compatible with webknossos.

- fixes: #299


tests/test_dataset.py
```diff
@@ -1396,6 +1409,68 @@ def test_adding_layer_with_valid_dtype_per_layer() -> None:
...
+        Layer.SEGMENTATION_TYPE,
+        dtype_per_channel="float",
+        num_channels=3,
+        largest_segment_id=100000,
+    )
+
+    with open(TESTOUTPUT_DIR / "valid_dtype" / "datasource-properties.json", "r") as f:
+        data = json.load(f)
+        # The order of the layers in the properties equals the order of creation
+        assert data["dataLayers"][0]["elementClass"] == "uint24"
```
normanrz, line 1451: @fm3 Should we deprecate `uint24` in favor of `uint8x3`? Same for `float96` -> `float32x3` (not sure if wk supports this).

normanrz (approved):

tests/test_dataset.py
fm3: wK does not currently support the `x3` syntax. I agree that it would be easier to read, but I am not sure it’s worth the effort, since legacy properties jsons would still have to be supported also, creating more complexity.

PR #318, opened 2021-05-12
PEP 561 compatibility to support typechecks in other projects

rschwanhold:
Even though the cuber is typed, `voxelytics` does not check if function calls to the cuber are correct.
This PR aims to make this project compatible with PEP 561 which tells `mypy` to typecheck this project if it is used as a library in other projects.

As https://www.python.org/dev/peps/pep-0561/ points out, the requires a `py.typed` file


fm3 (approved):
Let’s give it a try :)

PR #320, opened 2021-05-14
Save bounding box correctly after downsampling

### LINKED ISSUE #310: Bounding box in JSON after conversion
Starting from v0.5.0 the bounding box specified in the JSON is no longer the exact dimensions of the source TIFF image stack. If convert using just `python -m wkcuber`, then the xyz dimensions of the bounding box are all set to `1024`. If convert using `python -m wkcuber.convert_image_stack_to_wkw`, then the xyz dimensions of the bounding box are slightly larger than the source stack dimensions. Converting using v0.4.0 gives the correct bounding box dimensions.

youri-k:
### Description:
- after downsampling, padded bounding box should not be persisted to disk

### Issues:
- fixes #310 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/api/Layer.py
+++ b/wkcuber/api/Layer.py
@@ -132,24 +132,30 @@ class Layer(Generic[MagT]):
     def set_bounding_box(
         self, offset: Tuple[int, int, int], size: Tuple[int, int, int]
     ) -> None:
-        self.set_bounding_box_offset(offset)
-        self.set_bounding_box_size(size)
+        self.dataset.properties._set_bounding_box_of_layer(self.name, offset, size)
+        bounding_box = BoundingBox(offset, size)
+
+        for mag_name, mag in self.mags.items():
+            mag.view.size = cast(
+                Tuple[int, int, int],
+                tuple(
+                    bounding_box.align_with_mag(Mag(mag_name), ceil=True)
+                    .in_mag(Mag(mag_name))
+                    .bottomright
+                ),
+            )
 
     def set_bounding_box_offset(self, offset: Tuple[int, int, int]) -> None:
-        size: Tuple[int, int, int] = self.dataset.properties.data_layers[
-            "color"
-        ].get_bounding_box_size()
-        self.dataset.properties._set_bounding_box_of_layer(self.name, offset, size)
-        for _, mag in self.mags.items():
-            mag.view.global_offset = offset
+        size: Tuple[int, int, int] = self.dataset.properties.get_bounding_box_of_layer(
+            self.name
+        )[1]
+        self.set_bounding_box(offset, size)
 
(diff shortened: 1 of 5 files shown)


wkcuber/api/MagDataset.py
```diff
@@ -73,7 +73,7 @@ def read(
         offset: Tuple[int, int, int] = (0, 0, 0),
         size: Tuple[int, int, int] = None,
     ) -> np.array:
-        return self.view.read(offset, size)
+        return self.get_view().read(offset, size)
```
youri-k, line 76: @rschwanhold I was a bit confused by the test failure, but I was able to fix this with this change. I don't know the impact, but I think we need the correct offset => size calculation for the test to work.

youri-k: Oops, this seems to break more than it fixes. Can you have a look what the desired change should be?

rschwanhold: I will take a look at it

rschwanhold: The previous behavior was correct, but the test case was wrong.

wkcuber/api/Layer.py
```diff
@@ -258,6 +267,8 @@ def downsample(
             prev_mag = target_mag
             target_mag = get_next_mag(target_mag, scale)
 
+        self.set_bounding_box(original_bounding_box[0], original_bounding_box[1])
```
rschwanhold, line 270: You also have to fix it for `downsample_mag` and `downsample_mag_list`. (Everywhere where `_pad_existing_mags_for_downsampling` is used)

I have given this another thought and I think it is important to make sure that in case of an error, the bounding box does not get overwritten. 
I have come up with three possibilities:

1. One way would be to avoid exporting the properties to disk for this function call. However, this does not work because the properties are exported automatically occasionally (e.g. when a new mag is added (which is the case for downsampling)).
2. Not updating the properties in the first place (just the views themselves). If I am not mistaken, the only time during downsampling when the bounding box from the properties are used, is when `mag.get_view()` is called (in `downsample_mag`). The best approach would be to remove `_pad_existing_mags_for_downsampling` completely and also the parameter `pad_data` in `downsample_mag_list`. Then in `downsample_mag` you would have to pass the correct offset and size to `get_view()`.
3. Or you wrap the code in a `try` block and then reset the bb in the `except` before you propagate the error anyways. However, this still causes problems if the process is killed completely.

I would encourage you to go for Idea 2. I hope it works the way I imagine it (if you need help, just call me). This would also make the code much nicer because this "padding" might be confusing if someone encounters it for the first time.

rschwanhold (changes requested):
Looks very good and you also spotted a few errors in the old code :+1:, but there is one remark to make the code more resilient.

rschwanhold (approved):
LGTM

PR #326, opened 2021-05-27
Fix compressed write

rschwanhold:
### Description:
- Downsampling of compressed datasets failed. The cause was a bug in the the way compressed data was written
- `_handle_compressed_write` passed an absolute offset to `read`, which expected a relative offset. This is now fixed.

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/tests/test_dataset.py
+++ b/tests/test_dataset.py
@@ -1482,7 +1482,7 @@ def test_writing_subset_of_compressed_data_multi_channel() -> None:
     delete_dir(TESTOUTPUT_DIR / "compressed_data")
 
     # create uncompressed dataset
-    write_data1 = (np.random.rand(3, 20, 40, 60) * 255).astype(np.uint8)
+    write_data1 = (np.random.rand(3, 100, 120, 140) * 255).astype(np.uint8)
     WKDataset.create(TESTOUTPUT_DIR / "compressed_data", scale=(1, 1, 1)).add_layer(
         "color", Layer.COLOR_TYPE, num_channels=3
     ).add_mag("1", block_len=8, file_len=8).write(write_data1)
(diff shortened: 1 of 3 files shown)


tests/test_dataset.py
```diff
@@ -2050,11 +2050,16 @@ def test_compression() -> None:
 
     # writing unaligned data to an uncompressed dataset
     write_data = (np.random.rand(3, 10, 20, 30) * 255).astype(np.uint8)
-    mag1.write(write_data)
+
+    # Writing compressed data directly to "mag1" also works, but using a View here covers an additional edge case
```
philippotto, line 2054: Does this code "write compressed data directly"? Since `compress()` happens after the `write`, I would have thought that this is not the case.. If I'm wrong, please correct me, but otherwise, I would prefer if the test case was extended so that the compressed-write is also covered.

philippotto (changes requested):
Awesome that you found this bug :+1: I'm a bit confused about the test case. See my one comment. Other than that, please update the changelog, too :rocket:

tests/test_dataset.py
rschwanhold: Good catch, I adjusted the wrong test cases. 

I updated the `Changelog` in a way that we can create a new release `0.6.7` once this is merged.

tests/test_dataset.py
```diff
@@ -1500,24 +1500,25 @@ def test_writing_subset_of_compressed_data_multi_channel() -> None:
...
-    compressed_mag.write(
-        offset=(10, 20, 30), data=write_data2, allow_compressed_write=True
+    # Writing compressed data directly to "compressed_mag" also works, but using a View here covers an additional edge case
+    compressed_mag.get_view(offset=(50, 60, 70), is_bounded=False).write(
+        relative_offset=(10, 20, 30), data=write_data2, allow_compressed_write=True
     )
 
-    np.array_equal(
-        write_data2, compressed_mag.read(offset=(10, 20, 30), size=(10, 10, 10))
+    assert np.array_equal(
```
philippotto, line 1508: whoops, good catch with the asserts :+1:

philippotto (approved):
Perfect!

PR #333, opened 2021-06-17
Fix BoundinBox.copy()

georgwiese:
### Description:
Fixes a bug in `BoundingBox.copy()`.

Before:
```
>>> BoundingBox(topleft=(3072, 3072, 512), size=(1024, 1024, 1024)).copy()
BoundingBox(topleft=(3072, 3072, 512), size=(4096, 4096, 1536))
```


rschwanhold (approved):
LGTM

PR #339, opened 2021-07-15
Small improvements of the dataset API

rschwanhold:
### Description:
- allow to set the name of a dataset explicitly 
- use `Mag` as keys for the dictionary `Layer.mags`
- implement methods to get the only color/segmentation layer
- expose the `largest_segment_id` of a `Layer`
- improve the interface of `Layer.add_mag()`

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/api/dataset.py
+++ b/wkcuber/api/dataset.py
@@ -4,7 +4,7 @@ from shutil import rmtree
 from os import makedirs
 from os.path import join, normpath, basename
 from pathlib import Path
-from typing import Tuple, Union, Dict, Any, Optional, cast, List
+from typing import Tuple, Union, Dict, Any, Optional, cast
 
 import numpy as np
 import os
--- a/wkcuber/api/dataset.py
+++ b/wkcuber/api/dataset.py
@@ -4,18 +4,20 @@ from shutil import rmtree
 from os import makedirs
 from os.path import join, normpath, basename
 from pathlib import Path
-from typing import Tuple, Union, Dict, Any, Optional, cast
+from typing import Tuple, Union, Dict, Any, Optional, cast, List
 
 import numpy as np
 import os
 import re
 
+import wkw
+
 from wkcuber.api.properties.layer_properties import (
     properties_floating_type_to_python_type,
     SegmentationLayerProperties,
+    LayerProperties,
 )
 from wkcuber.api.bounding_box import BoundingBox
-from wkcuber.mag import Mag
 from wkcuber.utils import get_executor_for_args
 
 from wkcuber.api.properties.dataset_properties import Properties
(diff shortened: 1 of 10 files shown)


wkcuber/api/layer.py
```diff
@@ -97,38 +98,43 @@ def get_mag(self, mag: Union[int, str, list, tuple, np.ndarray, Mag]) -> MagView
...
         The return type is `wkcuber.api.mag_view.MagView`.
 
         Raises an IndexError if the specified `mag` already exists.
         """
         # normalize the name of the mag
-        mag = Mag(mag).to_layer_name()
+        mag = Mag(mag)
+        block_type = (
+            wkw.Header.BLOCK_TYPE_LZ4HC if compress else wkw.Header.BLOCK_TYPE_RAW
+        )
```
rschwanhold, line 128: I was using `LZ4HC` somewhere else in the API. Therefore, I decided to also use `LZ4HC` here. Do you think 'LZ4' would be better?

jstriebel: `LZ4HC` seems fine as a default for now. This improves the compression ratio and slows down the writing therefore (compared to `LZ4`), but reading does not seem to get worse, according to [this SO post](https://stackoverflow.com/questions/28635496/difference-lz77-vs-lz4-vs-lz4hc-compression-algorithms).
We're using `LZ4` so far in vx, but that was never conscious decision against the HC version.

wkcuber/api/layer.py
```diff
@@ -282,11 +290,11 @@ def downsample(
             assert (
                 len(self.mags.keys()) > 0
             ), "Failed to downsample data because no existing mag was found."
-            from_mag = Mag(max(self.mags.keys()))
+            from_mag = max(self.mags.keys())
```
jstriebel, line 293: :+1:

wkcuber/api/layer.py
```diff
@@ -560,43 +568,61 @@ def _setup_mag(self, mag: Union[str, Mag]) -> None:
...
+    def largest_segment_id(self) -> int:
+        layer = self.dataset.properties.data_layers[self.name]
+        assert isinstance(layer, SegmentationLayerProperties)
+        return layer.largest_segment_id
+
+    @largest_segment_id.setter
+    def largest_segment_id(self, largest_segment_id: int) -> None:
+        layer = self.dataset.properties._data_layers[self.name]
+        assert isinstance(layer, SegmentationLayerProperties)
+        layer._largest_segment_id = largest_segment_id
```
jstriebel, line 622: I'd rename `layer` to `layer_properties`, and we should re-export the datasource-properties after setting largest_segment_id: 

```suggestion
    @property
    def largest_segment_id(self) -> int:
        layer_properties = self.dataset.properties.data_layers[self.name]
        assert isinstance(layer_properties, SegmentationLayerProperties)
        return layer_properties.largest_segment_id

    @largest_segment_id.setter
    def largest_segment_id(self, largest_segment_id: int) -> None:
        layer_properties = self.dataset.properties._data_layers[self.name]
        assert isinstance(layer_properties, SegmentationLayerProperties)
        layer_properties._largest_segment_id = largest_segment_id
        self.dataset.properties._export_as_json()
```

jstriebel (commented):
Awesome stuff :heart: I added some suggestions here and there, probably @valentin-pinkau and I cross-add those and will merge it later.

valentin-pinkau (approved):
LGTM!

wkcuber/api/dataset.py
```diff
@@ -479,16 +507,16 @@ def copy_dataset(
...
+                            .in_mag(mag)
                             .bottomright
                         ),
                     )
                     target_mag.layer.set_bounding_box(offset=bbox[0], size=bbox[1])
 
                     # The data gets written to the target_mag.
                     # Therefore, the chunk size is determined by the target_mag to prevent concurrent writes
-                    mag.get_view().for_zipped_chunks(
+                    mag_view.get_view().for_zipped_chunks(
```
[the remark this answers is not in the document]
rschwanhold: I think you are right. The same applies to `target_view=target_mag.get_view()`. The idea behind this is that `mag_view.get_view()` returns the view that is limited to the bounding box from the properties (this works because the default parameter of `get_view()` are used). In general, this makes a difference if the bounding box does not start at `(0, 0, 0)`. However, the function `for_each_chunks` already does this internally for the very same reason. Therefore, this is redundant.

jstriebel:
@normanrz @hotzenklotz It seems that the python 3.8 CI check is not enforced to be green, could you also add this somewhere in the gh settings?

normanrz:
done

PR #344, opened 2021-07-20
Integrate view configurations into dataset API

### LINKED ISSUE #338: Add struct for view configuration
It would be great if there was a typed model for the view-configuration of a layer. 

```python
ds = Dataset("...")
l = ds.layers["color"]
l.set_view_configuration(ViewConfiguration(color=(255,255,0), alpha=1.0, intensity_range=(-12.3e+1,123), is_inverted=True))
assert l.get_view_configuration().color == (255, 255, 0)
```

rschwanhold:
### Description:
- Add methods to get and set the view configuration of a layer

### Issues:
- fixes #338

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/api/layer.py
+++ b/wkcuber/api/layer.py
@@ -45,21 +45,8 @@ from wkcuber.utils import (
     DEFAULT_WKW_FILE_LEN,
     get_executor_for_args,
     named_partial,
+    _to_camel_case,
 )
-
-
-class ViewConfiguration:
-    def __init__(
-        self,
-        color: Tuple[int, int, int],
-        alpha: Optional[float] = None,
-        intensity_range: Optional[Tuple[float, float]] = None,
-        is_inverted: Optional[bool] = None,
-    ):
-        self.color = color
-        self.alpha = alpha
-        self.intensity_range = intensity_range
-        self.is_inverted = is_inverted
 
 
 class Layer:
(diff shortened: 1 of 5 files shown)


wkcuber/api/layer.py
```diff
@@ -48,6 +48,20 @@
...
 
 
+class ViewConfiguration:
+    def __init__(
+        self,
+        color: Tuple[int, int, int],
+        alpha: Optional[float] = None,
+        intensity_range: Optional[Tuple[float, float]] = None,
+        is_inverted: Optional[bool] = None,
+    ):
```
rschwanhold, line 58: Is it correct that only the `color` is required?

wkcuber/api/layer.py
```diff
@@ -607,6 +621,29 @@ def _initialize_mag_from_other_mag(
             compress=compress,
         )
 
+    def set_view_configuration(self, view_configuration: ViewConfiguration) -> None:
+        self.dataset.properties._data_layers[
+            self.name
+        ]._default_view_configuration = vars(view_configuration)
```
rschwanhold, line 627: I am not familiar with the use case of this issue. I have assumed that this refers to the entry `defaultViewConfiguration` in the properties. Is this correct or is there an additional view configuration besides `defaultViewConfiguration`?

rschwanhold:
I just realized that the `datasource-properties.json` at https://github.com/scalableminds/webknossos-cuber/blob/master/testdata/simple_wk_dataset/datasource-properties.json also contains a `defaultViewConfiguration` for the whole dataset. Currently I only added the functions for individual layers.

Its only attribute is `position`. Are there any other attributes that are not considered yet?

jstriebel:
Btw, there is a json-schema definition of the datasource-properties.json in webknossos, this might help to find relevant fields and also validate the files:
https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/types/schemas/datasource.schema.js

philippotto (approved):
looks very good to me :)

PR #345, opened 2021-07-22
Add method to Dataset API to copy an existing layer from a different dataset

### LINKED ISSUE #343: Add method “add_copy_layer” analog to add_symlink_layer
add_symlink_layer takes the properties of the source layer from another dataset.
It would be useful to have an analog that copies instead of symlinking

rschwanhold:
### Description:
- Add method to Dataset API to copy an existing layer from a different dataset

### Issues:
- fixes #343 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/api/dataset.py
+++ b/wkcuber/api/dataset.py
@@ -1,4 +1,5 @@
 import operator
+import shutil
 from argparse import Namespace
 from shutil import rmtree
 from os import makedirs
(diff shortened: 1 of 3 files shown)


wkcuber/api/dataset.py
```diff
@@ -442,6 +443,40 @@ def add_symlink_layer(self, foreign_layer_path: Union[str, Path]) -> Layer:
...
+        """
+
+        foreign_layer_path = Path(os.path.abspath(foreign_layer_path))
+        layer_name = foreign_layer_path.name
+        if layer_name in self.layers.keys():
+            raise IndexError(
+                f"Cannot copy {foreign_layer_path}. This dataset already has a layer called {layer_name}."
+            )
+
+        shutil.copytree(foreign_layer_path, join(self.path, layer_name))
```
rschwanhold, line 459: Is it enough to copy the data like this, or should it be distributed like https://github.com/scalableminds/voxelytics/blob/master/voxelytics/distributed_copy.py?

fm3: For my current use case, copytree will do fine :) It may make sense to add parallelization later.

fm3 (approved):
Wow, that was quick! Looks good to me

PR #347, opened 2021-07-22
Dataset api read at a specific bounding box

rschwanhold:
### Description:
- `View.read()` has the parameters `offset` and `size`. However, often it is easier to pass a bounding box. Therefore, the method `read_bbox()` was added.

### Todos:
 - [ ] Updated Changelog


jstriebel (approved):
:heart:

PR #356, opened 2021-07-22
Remove the parameter allow_compressed_write by always allowing it

### LINKED ISSUE #354: dataset api: always allow_compressed_write and warn in this case
To simplify the dataset api writes on compressed datasets, we want to allow unaligned compressed writes, which need `allow_compressed_write=True` so far, to guard against read-whole-chunk-write-a-few-vx patterns. Instead, we could create a [warning](https://docs.python.org/3/library/warnings.html), which has the nice properties of showing up only once by default, and can be converted to an error if wished (see [warning filters](https://docs.python.org/3/library/warnings.html#the-warnings-filter)).

rschwanhold:
### Description:
- Remove the parameter `allow_compressed_write` from `View.write()`. This parameter existed to avoid that the performance was reduced by making many small write-requests. However, we decided to use warnings for this case.

### Issues:
- fixes #354 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/api/layer.py
+++ b/wkcuber/api/layer.py
@@ -419,7 +419,6 @@ class Layer:
                 mag_factors=mag_factors,
                 interpolation_mode=parsed_interpolation_mode,
                 buffer_edge_len=buffer_edge_len,
-                compress=compress,
                 job_count_per_log=job_count_per_log,
             )
 
@@ -549,7 +548,6 @@ class Layer:
                     upsample_cube_job,
                     mag_factors=mag_factors,
                     buffer_edge_len=buffer_edge_len,
-                    compress=compress,
                     job_count_per_log=job_count_per_log,
                 )
                 prev_mag_view.get_view().for_zipped_chunks(
--- a/wkcuber/api/layer.py
+++ b/wkcuber/api/layer.py
@@ -118,7 +118,7 @@ class Layer:
         The parameter `block_len`, `file_len` and `compress` can be
         specified to adjust how the data is stored on disk.
         Note that writing compressed data which is not aligned with the blocks on disk may result in
-        diminished performance, as full blocks will automatically read to pad the write actions. Alternatively,
+        diminished performance, as full blocks will automatically be read to pad the write actions. Alternatively,
         you can call mag.compress() after all the data was written
 
         The return type is `wkcuber.api.mag_view.MagView`.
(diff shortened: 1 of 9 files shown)


jstriebel:
@rschwanhold Is this ready for review?

rschwanhold:
> @rschwanhold Is this ready for review?

Sorry, yes it is ready

wkcuber/api/layer.py
```diff
@@ -116,6 +116,9 @@ def add_mag(
         Creates a new mag called and adds it to the layer.
         The parameter `block_len`, `file_len` and `compress` can be
         specified to adjust how the data is stored on disk.
+        Note that writing compressed data which is not aligned with the blocks on disk may result in
+        diminished performance, as full blocks will automatically read to pad the write actions. Alternatively,
```
philippotto, line 120:
```suggestion
        diminished performance, as full blocks will automatically be read to pad the write actions. Alternatively,
```

wkcuber/api/mag_view.py
```diff
@@ -99,24 +99,18 @@ def __init__(
...
+    def write(self, data: np.ndarray, offset: Tuple[int, int, int] = (0, 0, 0)) -> None:
         """
         Writes the `data` at the specified `offset` to disk (like `wkcuber.api.view.View.write()`).
 
         The `offset` refers to the absolute position, regardless of the offset in the properties (because the global_offset is set to (0, 0, 0)).
         If the data exceeds the original bounding box, the properties are updated.
 
-        If the data on disk is compressed, the passed `data` either has to be aligned with the files on disk
-        or `allow_compressed_write` has to be `True`. If `allow_compressed_write` is `True`, `data` is padded by
+        If the data on disk is compressed and the passed `data` is not aligned with the files on disk, it is padded by
```
philippotto, line 109: should the performance implication be mentioned here, too?

philippotto (commented):
looks good to me! maybe @jstriebel can also have a look at the warning-related code, since I don't have any experience with that. what happens if some code uses the cuber-api and another 3rd party library Y which both emit warnings? can one configure to escalate the cuber-warning to an error, but not the Y warnings?

jstriebel:
I had a rough look at the code already, seems fine to me :+1: 

> Can one configure to escalate the cuber-warning to an error, but not the Y warnings?

Yes, it's possible to filter warnings by the module where it's coming from using a regex. From the `warnings.filterwarnings` docs:

> `module` is a string containing a regular expression that the module name must match. The expression is compiled to be case-sensitive.

philippotto (approved):

PR #358, opened 2021-07-23
[ds-api] Modification of dataset with float-type layer results in JSON dump error

erjel:
### Description:
`LayerProperties._from_json()` and `LayerProperties._to_json()` are not idempotent. I.e.
```
from wkcuber.api.layer import LayerCategories
from wkcuber.api.properties.layer_properties import LayerProperties
from wkcuber.api.properties.resolution_properties import Resolution

# Initialize valid test_layer
test_layer = LayerProperties(
    "float_valued_color_channel",
    LayerCategories.COLOR_TYPE,
    'float',
    "wkw",
    1    
)

test_layer_2 = LayerProperties._from_json(
    test_layer._to_json(), Resolution, None
)


assert(test_layer._to_json() == test_layer_2._to_json())
```
Fails, since `test_layer[elementClass'] = 'float'` and `test_layer_2 ['elementClass'] = dtype('float32')`. The reason for the behavior is the missing conversion from `numpy.float` -> `str` by[ `python_floating_type_to_properties_type`](https://github.com/scalableminds/webknossos-cuber/blob/c89f3267d25dc5f7dbc5142f578278cc3fcd8d90/wkcuber/api/properties/layer_properties.py#L54).

This has practical relevance: The modification of an existing `Dataset` with a float-type `Layer` (i.e. by adding another layer) results in a `TypeError` during the dump of the updated data set properties. JSON can not serialize objects of type `dtype[float32]`.

This pull request adds a basic test case and provides the necessary fix to prevent a test case failure.

### Todos


rschwanhold (changes requested):
Thank you very much for your PR.
While your solution works, it does not fix the underlying mistake in the code.
`LayerProperties.element_class` is supposed to be a `str` (not a `np.dtype(...)`). However, this is currently not the case when loading a JSON (as your test points out).
I would suggest to change the two dicts to:
```
properties_floating_type_to_python_type = {
    "float": "float32",
    np.float: "float32",
    float: "float32",
    "double": "float64",
}

python_floating_type_to_properties_type = {
    "float32": "float",
    "float64": "double",
}
```
(note that `properties_floating_type_to_python_type` returns strings)

Your test case is great :+1:

rschwanhold (approved):
LGTM

rschwanhold:
We are ready to merge this, once you have signed the CLA.

erjel:
I asked my supervisor for permission to sign it (the CLA directly conflicts with other agreements between me and my employer). Currently I do not see a major road block, but I would like to have a formal permission ...

rschwanhold:
I have now opened a new PR (https://github.com/scalableminds/webknossos-cuber/pull/375) that fixes this bug because this mistake is quite critical and could block other people. Merging this PR is still beneficial to add your testcase. 
Thank you again for spotting this mistake.

erjel:
ping @rschwanhold

PR #360, opened 2021-07-26
Add optional make_relative: bool to add_symlink_layer

fm3:
### Description:
- If the dataset directory is mounted via docker, absolute symlink paths can break outside of the docker container
- For such cases it is helpful to make the symlink relative instead
- This PR adds the optional parameter `make_relative: bool` to `wkcuber.api.dataset.Dataset.add_symlink_layer`

### Todos:
 - [x] Updated Changelog


jstriebel (approved):
LGTM

PR #362, opened 2021-07-26
default to -1 for largestSegmentId

jstriebel:
### Description:
Sometimes the `largestSegmentId` is not set in a datasource-properties.json. A default of `-1` is used then, together with a warning about the missing field. When writing the datasource-properties the next time, the field is persisted.

### Todos:
 - [x] Updated Changelog


georgwiese (approved):
Code LGTM!

PR #365, opened 2021-07-28
Fix add_symlink_layer properties lookup

fm3:
Fixed a bug where Dataset.add_symlink_layer(make_relative=True) failed to look up dataset properties.

### Todos:
- [x] Updated Changelog


jstriebel (approved):
:+1:

PR #372, opened 2021-08-05
Bump Dependencies for scipy and scikit-image

leowe:
### Description:
- for building the project on a M1 Mac newer packages are necessary, because those are available over `condo-forge`, so bumping `scikit-image` and `scipy`
- Fixing `numpy` version to `1.19.5` since `vx` needs a version `<1.20.0`

### Todos:
 - [x] Updated Changelog


philippotto (approved):
LGTM! :+1:

PR #374, opened 2021-08-08
WIP: Dataset api properties refactoring

### LINKED ISSUE #348: Dataset API: remove parallel properties-classes structure
The …Properties classes are (mostly) mirroring the same structure as the Dataset/Layer/Mag-structure. On the other hand, those classes interact with the properties, often via simple setter/getter proxies. Instead, we should incorporate the property-information into the Dataset/Layer/Mag classes and add to-json/from-json converters to them. (As discussed in person, [cattrs](https://cattrs.readthedocs.io) might be a good candidate for those.)

rschwanhold:
### Description:
- abc

### Issues:
- fixes #348 

### Todos:
 - [ ] Updated Changelog


tests/test_dataset.py
```diff
@@ -517,14 +490,14 @@ def test_open_dataset_without_num_channels_in_properties() -> None:
...
-    ds.properties._export_as_json()
+    assert ds.get_layer("color").num_channels == 1
+    ds._export_as_json()
 
     with open(
         TESTOUTPUT_DIR / "old_wk_dataset" / "datasource-properties.json"
     ) as datasource_properties:
         data = json.load(datasource_properties)
-        assert data["dataLayers"][0].get("num_channels") == 1
+        assert data["dataLayers"][0].get("numChannels") == 1
```
rschwanhold, line 500: The entry `num_channels` in the properties is now renamed to `numChannels` to be consistent with the other parameter. This should not be a problem for existing datasets because the parameter `num_channels` was only used by the dataset API (to my knowledge (see: https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/types/schemas/datasource.schema.js)) and it is optional.

testdata/simple_wk_dataset/datasource-properties.json
```diff
@@ -4,7 +4,7 @@
         "team": ""
     },
     "scale": [
-        1
+        1, 1, 1
```
rschwanhold, line 7: According to https://github.com/scalableminds/webknossos/blob/master/frontend/javascripts/types/schemas/datasource.schema.js, the scale always consists of 3 numbers.

wkcuber/api/dataset.py
```diff
@@ -647,81 +586,33 @@ def get_or_create(
...
     def __repr__(self) -> str:
         return repr("Dataset(%s)" % self.path)
 
+    def _export_as_json(self) -> None:
+        with open(self.path / PROPERTIES_FILE_NAME, "w") as outfile:
+            json.dump(
+                dataset_converter.unstructure(self._properties),
+                outfile,
+                indent=4,
+                separators=(",", ": "),
```
jstriebel, line 607: Those are the default seperators with indent anyways:
https://docs.python.org/3/library/json.html#json.dump
```suggestion
```

wkcuber/api/layer.py
```diff
@@ -68,25 +168,65 @@ class Layer:
...
+    def _properties(self) -> LayerProperties:
+        return next(
+            layer_property
+            for layer_property in self.dataset._properties.data_layers
+            if layer_property.name == self.name
+        )
+
+    @property
+    def name(self) -> str:
+        return self._name
```
jstriebel, line 204: This makes the name read-only. I think we could also add a setter ([using `@name.setter`](https://docs.python.org/3/library/functions.html#property)), since we have the reference to the dataset, and can change the dict-entry as well as the json-file.

wkcuber/api/layer.py
```diff
@@ -193,7 +339,6 @@ def delete_mag(self, mag: Union[int, str, list, tuple, np.ndarray, Mag]) -> None
             )
 
         del self._mags[mag]
-        self.dataset.properties._delete_mag(self.name, mag.to_layer_name())
```
jstriebel, line 196: Don't we also need to re-export the datasource-properties.json? Please double-check all places where properties have been changed before, that we don't forget any re-exports.

wkcuber/api/layer.py
```diff
@@ -223,37 +368,33 @@ def set_bounding_box(
...
         """
-        offset: Tuple[
-            int, int, int
-        ] = self.dataset.properties.get_bounding_box_of_layer(self.name)[0]
-        self.set_bounding_box(offset, size)
+        self.set_bounding_box(self._properties.bounding_box.topleft, size)
 
     def get_bounding_box(self) -> BoundingBox:
-        return self.dataset.properties.data_layers[self.name].get_bounding_box()
+        return self._properties.bounding_box.copy()
```
jstriebel, line 397: It's weird that `get_bounding_box` and `set_bounding_box` have different signatures, but probably fine for now. Let's rework this another time.

wkcuber/api/layer.py
```diff
@@ -223,37 +368,33 @@ def set_bounding_box(
...
 
     def rename(self, layer_name: str) -> None:
```
jstriebel, line 399: Ah, this might actually just become the setter I mentioned above.

wkcuber/api/layer.py
```diff
@@ -635,58 +761,39 @@ def _initialize_mag_from_other_mag(
             compress=compress,
         )
 
-    def set_view_configuration(
```
jstriebel, line 638: :+1: for converting this into a getter & setter :-)

wkcuber/api/layer.py
```diff
@@ -68,25 +168,65 @@ class Layer:
...
+        for resolution in properties.wkw_resolutions:
+            self._setup_mag(Mag(resolution.resolution).to_layer_name())
+
+    @property
+    def _properties(self) -> LayerProperties:
+        return next(
+            layer_property
+            for layer_property in self.dataset._properties.data_layers
+            if layer_property.name == self.name
+        )
```
jstriebel, line 200: This is a quite elegant solution :+1:

wkcuber/api/mag_view.py
```diff
@@ -91,14 +92,30 @@ def __init__(
             False,
         )
 
-        self.layer = layer
-        self.name = name
+        self._layer = layer
+        self._name = name
```
jstriebel, line 96: I'd rather use the `mag` itself as the main identifier and also pass it in the init instead of the name. The name could still be available as a property using `self._mag.to_layer_name()`.

jstriebel: Then also the `Mag(self.name)` could be simple `self.mag`/`self._mag` calls.

wkcuber/api/properties.py
```diff
@@ -0,0 +1,220 @@
...
+from wkcuber.api import BoundingBox
+
+
+def _extract_num_channels(
+    num_channels_in_properties: Optional[int],
+    path: Path,
+    layer: str,
+    mag: Optional[Union[int, Mag]],
+) -> int:
+    # if a wk dataset is not created with this API, then it most likely doesn't have the attribute 'num_channels' in the
```
jstriebel, line 21: Please use the new `numChannels` name consistently when referring to the json key:

```suggestion
    # if a wk dataset is not created with this API, then it most likely doesn't have the attribute 'numChannels' in the
```

wkcuber/api/properties.py
```diff
@@ -0,0 +1,220 @@
...
+    # datasource-properties.json. In this case we need to extract the 'num_channels' from the 'header.wkw'.
+    if num_channels_in_properties is not None:
+        return num_channels_in_properties
+
+    mag = Mag(mag)
+    if mag is None:
```
jstriebel, line 27: This seems weird, `mag` cannot be None after the previous line anymore, probably must be switched?

wkcuber/api/properties.py
```diff
@@ -0,0 +1,220 @@
...
+    )
+    dataset_converter.register_structure_hook(
+        cls,
+        make_dict_structure_fn(
+            cls,
+            dataset_converter,
+            **{
+                a.name: override(rename=_snake_to_camel_case(a.name))
+                for a in attr.fields(cls)  # pylint: disable=not-an-iterable
+            },
```
jstriebel, line 181: Wow, this is quite complex, but also super elegant :heart: Please add a comment though (maybe directly in the new function, as explained in the other comment)

wkcuber/api/properties.py
```diff
@@ -0,0 +1,220 @@
...
+
+# register (un-)structure hooks for attr-classes
+for cls in [
+    DatasetProperties,
+    LayerProperties,
+    SegmentationLayerProperties,
+    MagViewProperties,
+    DatasetViewConfiguration,
+    LayerViewConfiguration,
+]:
```
jstriebel, line 159: Instead of this loop, you could register (un-)structuring factories:
```
def snake_to_camel_case_overrides(attrs_type):
    return {
        a.name: override(omit_if_default=True, rename=_snake_to_camel_case(a.name))
        for a in attr.fields(attrs_type)
    }

register_unstructure_hook_factory(attrs.has, lambda attrs_type: make_dict_unstructure_fn(cls, dataset_converter, **snake_to_camel_case_overrides(attrs_type)))
# same for structuring
```

wkcuber/api/properties.py
```diff
@@ -0,0 +1,220 @@
...
+def disambiguate_mag(obj: dict, _: Any) -> Mag:
+    # This function is necessary because cattrs does not support unions of non-attrs objects out of the box
+    return Mag(obj)
+
+
+dataset_converter.register_structure_hook(Union[int, Mag], disambiguate_mag)
+
+# Separate converter to unstructure LayerProperties
+# This is used to initialize SegmentationLayerProperties from LayerProperties
+# The important difference to the dataset_converter is that the names of the attributes stay the same while defaults are omitted.
```
jstriebel, line 213:
```suggestion
# The important difference to the dataset_converter is that the names of the attributes stay the same while defaults are also omitted.
```

jstriebel (commented):
I have to say I am a huge fan of this refactoring :heart_eyes:



I added some comments, mostly nitpicks and minor suggestions, overall this looks great to me :+1:

wkcuber/api/layer.py
rschwanhold: Do you mean we should rename the `rename()`-method to `name` (the setter)?

wkcuber/api/layer.py
rschwanhold: Yes

wkcuber/api/layer.py
jstriebel: Yes, exactly.

wkcuber/api/dataset.py
```diff
@@ -294,17 +214,42 @@ def add_layer(
...
-            self._data_format,
-            num_channels,
-            **kwargs,
-        )
-        self._layers[layer_name] = self._create_layer(
-            layer_name, dtype_per_channel, num_channels, category
+        layer_properties = LayerProperties(
+            name=layer_name,
+            category=category,
+            bounding_box=BoundingBox((-1, -1, -1), (-1, -1, -1)),
```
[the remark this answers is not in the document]
jstriebel: Both fine for me, but if you keep the -1s, please add an explaining comment.

wkcuber/api/properties.py
rschwanhold: I agree that is looks cleaner, but `cattr.Converter.register_unstructure_hook_factory` seems to be added in `1.9.0` which is not released yet.

pyproject.toml
```diff
@@ -22,6 +22,8 @@ scikit-learn = "^0.24.0"
 tifffile = "^2020.11.26"
 imagecodecs = "^2020.5.30"
 czifile = "^2019.7.2"
+attrs = "^21.1.0"
+cattrs = "1.7.1"
```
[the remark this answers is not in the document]
rschwanhold: I don't know why, but with `1.8.0` I get the following error in many test cases:
`cattr.errors.StructureHandlerNotFoundError: Unsupported type: typing.Union[int, wkcuber.mag.Mag]. Register a structure hook for it.`
Even though I have a structure-hook.

testdata/simple_wk_dataset/datasource-properties.json
rschwanhold: I don't know why this file contained such a scale. We already used to write `Tuple[float, float, float]` before this PR.
This should not have worked before because the scale would have been simply `(1,)`. I assume that the test cases which used this dataset never needed the scale.
The current implementation does not support this. Do you think this is necessary? I am not sure how to define a structuring hook for an individual attribute (instead of a class).

wkcuber/api/layer.py
rschwanhold: @valentin-pinkau also noticed this. We could use `@property` for getter and setter. The methods `set_bounding_box_offset` and `set_bounding_box_size` could be removed. They are only used in the tests in cuber.

PR #375, opened 2021-08-08
Fix serialization of existing properties with floats as dtype

rschwanhold:
### Description:
- Fixes a bug where modifications to an existing dataset with floats as dtype failed.

### Todos:
 - [x] Updated Changelog


wkcuber/api/properties/layer_properties.py
```diff
@@ -46,10 +46,10 @@ def _extract_num_channels(
 
 
 properties_floating_type_to_python_type = {
-    "float": np.dtype("float32"),
-    np.float: np.dtype("float32"),
-    float: np.dtype("float32"),
-    "double": np.dtype("float64"),
+    "float": "float32",
+    np.float: "float32",
+    float: "float32",
```
jstriebel, line 51: Not sure if those are actually needed, since they can't be part of the json, can they? Doesn't matter too much, though.

jstriebel (approved):
LGTM :+1: 

PS: Usually mypy should catch this. In this case numpy is untyped and `np.dtype(…)` tranlates to `Any`. We might want to upgrade to numpy 1.20+, since this is typed, for this reason (but not depend on the high version, just use it for CI etc…).

PR #378, opened 2021-08-12
[ds-api] Don't warn about unaligned/compressed writes if these are at dataset boundary

### LINKED ISSUE #377: Don't warn about unaligned/compressed writes if these are at dataset boundary
For example, during downsampling there are warnings about unaligned/compressed writes (see cuber CI). I think, these happen when the dataset boundaries are not block-aligned. In that case, the unaligned/compressed writes should not be problematic (since they cannot be avoided except for intentionally writing outside of the dataset's bounding box).

@rschwanhold Could you have a look into this? Let me know if we should have a call about it. I assume that `_handle_compressed_write` needs to be adapted so that the mag's bounding box is taken into account, too.

rschwanhold:
### Description:
- do not raise a warning for compressed datasets while writing if the data is at the boundary of the bounding box

### Issues:
- fixes #377 

### Todos:
 - [x] Updated Changelog


wkcuber/api/view.py
```diff
@@ -484,6 +509,11 @@ def __repr__(self) -> str:
             % (self.path, self.global_offset, self.size)
         )
 
+    @property
+    def _mag_view_bbox(self) -> BoundingBox:
```
philippotto, line 513: I think, the getter should also have the `_at_creation` suffix.

wkcuber/api/view.py
```diff
@@ -433,17 +436,39 @@ def _handle_compressed_write(
...
-            # That is the case if the data is compressed but the view does not include the whole file on disk.
-            # In this case we avoid checking the bounds because the aligned_offset and aligned_shape are calculated internally.
-            warnings.warn(
-                "Warning: write() was called on a compressed mag without block alignment. Performance will be degraded as the data has to be padded first.",
-                RuntimeWarning,
+            mag_view_bbox_at_creation = self._mag_view_bbox
+
+            # Calculate in which dimensions the data is aligned and in which dimensions it matches the bbox of the mag.
+            aligned_top_left = aligned_offset == np.array(absolute_offset)
+            aligned_bottom_right = aligned_bottom_right == bottom_right
```
philippotto, line 443:
```suggestion
            is_top_left_aligned = aligned_offset == np.array(absolute_offset)
            is_bottom_right_aligned = aligned_bottom_right == bottom_right
```

would be better imo (e.g., `aligned_offset` has currently the same naming scheme and isn't a boolean).

wkcuber/api/view.py
```diff
@@ -433,17 +436,39 @@ def _handle_compressed_write(
...
+            at_border_top_left = np.array(
+                mag_view_bbox_at_creation.topleft
+            ) == np.array(absolute_offset)
+            at_border_bottom_right = (
```
philippotto, line 447:
```suggestion
            is_at_border_top_left = np.array(
                mag_view_bbox_at_creation.topleft
            ) == np.array(absolute_offset)
            is_at_border_bottom_right = (
```

wkcuber/api/view.py
```diff
@@ -433,17 +436,39 @@ def _handle_compressed_write(
...
+                np.array(mag_view_bbox_at_creation.bottomright) == bottom_right
             )
+
+            if not (
+                np.array_equal(
+                    np.logical_or(aligned_top_left, at_border_top_left),
+                    np.array([True, True, True]),
+                )
```
philippotto, line 455:
```suggestion
                np.all(
                    np.logical_or(aligned_top_left, at_border_top_left),
                )
```

wkcuber/api/view.py
```diff
@@ -433,17 +436,39 @@ def _handle_compressed_write(
...
+                and np.array_equal(
+                    np.logical_or(aligned_bottom_right, at_border_bottom_right),
+                    np.array([True, True, True]),
+                )
```
philippotto, line 459:
```suggestion
                and np.all(
                    np.logical_or(aligned_bottom_right, at_border_bottom_right),
                )
```

wkcuber/api/view.py
```diff
@@ -41,6 +42,7 @@ def __init__(
         self._is_bounded = is_bounded
         self.read_only = read_only
         self._is_opened = False
+        self._mag_view_bbox_at_creation = mag_view_bbox_at_creation
```
philippotto, line 45: Please add a comment which explains the reasoning we had about this (we don't want to store a reference to the mag and therefore pass a snapshot of the bbox which can be used for the alignment checks, since it's not super critical when the warning will be (not) emitted incorrectly if the bounding box changes in between).

philippotto (changes requested):
Great stuff! Only have some cosmetic feedback :)

rschwanhold:
There are 2 occurrences in the cuber where warnings are raised:
- this scenario
- if the user does not specify `largest_segment_id` for segmentation layers

It is possible to specify the type of the warning (instead of just `Warning`). Do you think we should raise a `RuntimeWarning` in this scenario and a `UserWarning` for the `largest_segment_id`?

philippotto (approved):
LGTM :+1:

> if the user does not specify largest_segment_id for segmentation layers

> It is possible to specify the type of the warning (instead of just Warning). Do you think we should raise a RuntimeWarning in this scenario and a UserWarning for the largest_segment_id?

What's the motivation for this? In voxelytics, we will only raise exceptions for warnings regarding the unaligned/compressed stuff (the filtering can be done via the message). So, I don't see the need to change the `largest_segment_id` thing to a UserWarning. However, it might still be a good idea (?). Let's have a call if you think we should discuss that option further. I'm also fine with merging the PR as is, since I don't see any problems (but maybe I'm missing something?).

PR #387, opened 2021-08-18
Suggestions for memory optimization

elhuhdron:
### Description:

As discussed with @normanrz, I've run into memory usage issues when utilizing the cuber for a relatively large dataset, total size about 32T and per-slice size of 32G.

### Issues:

I identified a couple locations that allowed the cuber to continue on our cluster (nodes with 375G each) using --batch_size 8. Before these changes, the jobs were killed because of memory usage unless I used --batch_size 1.

The issues were identified in [cubing.py](https://github.com/scalableminds/webknossos-cuber/blob/master/wkcuber/cubing.py), particularly in regards to the function [prepare_slices_for_wkw](https://github.com/scalableminds/webknossos-cuber/blob/eb34831485b933f58ce96f49520dfcd067918491/wkcuber/cubing.py#L138).
(1) concatenation of the slices array into the buffer, this duplicates the memory usage since the slices are then in the buffer and still resident in the slices array
(2) the transpose of the buffer creates a view, which means the ndarray is no longer contiguous. This means that when it is passed into webknossos-wrap, it is reallocated again so that it is F-order contiguous [here](https://github.com/scalableminds/webknossos-wrap/blob/6937ec22d5f647c34be7b5d9a42c3a8a535afcf4/python/wkw/wkw.py#L204).

### Todos:
 - I also modified [image_readers.py](https://github.com/scalableminds/webknossos-cuber/blob/master/wkcuber/image_readers.py) so that it can read the format of the slices coming out of our aligner. The format is just a single slice stored per hdf5 file, in a dataset named 'image'. This is somewhat specific use case, so you may not want to incorporate this as is. Additionally I use BLOSC compression in hdf5, which is the reason for the pytables import.


rschwanhold:
@elhuhdron thank you again for your contribution. I have now implemented your performance improvements in PR #480, which is now merged. For that reason, I am closing this PR.
If you feel that your image reader for hdf5 files should also be integrated, feel free to open a separate PR for this.

PR #390, opened 2021-08-19
Rollback to older scikit-image version

rschwanhold:
### Description:
- Rollback `scikit-image` version from `0.18.0` to `0.16.2` because the newer version cause problems in voxelytics.

### Todos:
 - [x] Updated Changelog


valentin-pinkau (approved):
LGTM

PR #391, opened 2021-08-20
Fix downsampling scheme

### LINKED ISSUE #337: Anisotropic downsampling doesn't work as expected
I was trying  to downsample a dataset with scale 11.24,11.24,28. By default it will create 1, 2-2-1, 4-4-1, 8-8-2, ...
However, I want something like 1, 2-2-1, 4-4-2, 8-8-4, ...

Earlier, I was able to do that with `--anisotropic_target_mag 512-512-256`. That flag seems to be deprecated. The error message says, I should be using  `target_mag` which is not a valid arg. https://github.com/scalableminds/webknossos-cuber/blob/master/wkcuber/downsampling.py#L135-L138

Btw, my understanding of "deprecated" is that a functionality is not encouraged anymore, but still works now, and will be removed later.

rschwanhold:
### Description:
- Fix downsampling scheme to always try to minimize the scaled difference between the different dimensions of a mag.
  - Previously, the scale `(11.24,11.24,28)` resulted in the following mags: `1, 2-2-1, 4-4-1, 8-8-2, ...`
  - However, `1, 2-2-1, 4-4-2, 8-8-4, ...` was expected. 
- Renamed the sampling mode `auto` to `anisotropic`.
- Fixed an error messages that recommended to use a non existing parameter.

### Issues:
- fixes #337

### Todos:
 - [x] Updated Changelog


wkcuber/api/layer.py
```diff
@@ -415,27 +415,36 @@ def downsample(
...
-        prev_mag = from_mag
-        target_mag = get_next_mag(prev_mag, scale)
+        mags_to_downsample = calculate_mags_to_downsample(from_mag, max_mag, scale)
+
+        if len(set([max(m.to_array()) for m in mags_to_downsample])) != len(
+            mags_to_downsample
+        ):
+            warnings.warn(
+                "The downsampling scheme contains multiple magnifications with the same maximum value. This is not supported by webknossos."
+            )
```
philippotto, line 443: When would this happen? Shouldn't we avoid this at all costs, since this will be very annoying for users.

philippotto (changes requested):
Looks cool! I just left some comments and questions for now.

wkcuber/api/layer.py
rschwanhold: For example if the user uses `from_mag=(1, 1, 4)` with `CONSTANT_Z` (and some `max_mag`). This would lead to e.g. `1-1-4, 2-2-4, 4-4-4, 8-8-4, ...`.

rschwanhold: Another example (that does not require `CONSTANT_Z`):
`mode=ANISOTROPIC`, `from_mag=(2, 2, 1)`, `max_mag=(8, 8, 16)` and `scale=(2, 2, 1)`. This would result in `(4, 4, 2), (8, 8, 4), (8, 8, 8), (8, 8, 16)`.

wkcuber/downsampling_utils.py
```diff
@@ -36,73 +36,49 @@ def determine_buffer_edge_len(dataset: wkw.Dataset) -> int:
...
-    target_mag: Mag,
-) -> Tuple[float, float, float]:
-    """
-    This scale is not the actual scale of the dataset
-    The virtual scale is used for downsample_mags_anisotropic.
-    """
-    max_target_value = max(list(target_mag.to_array()))
-    scale_array = max_target_value / np.array(target_mag.to_array())
-    return cast(Tuple[float, float, float], tuple(scale_array))
+            current_size = current_mag.as_np() * np.array(scale)
```
philippotto, line 49: please add a comment that this else branch is only for the anisotropic case. also that isotropic and constant-z are handled in the if-branch above.

wkcuber/downsampling_utils.py
```diff
@@ -36,73 +36,49 @@ def determine_buffer_edge_len(dataset: wkw.Dataset) -> int:
...
+            min_value = np.min(current_size)
+            min_value_bitmask = np.array(current_size == min_value)
+            factor = min_value_bitmask + 1
+
+            # Calculate the two potential magnifications
```
philippotto, line 54:
```suggestion
            # Calculate the two potential magnifications.
            # Either, double all components or only double the smallest component.
```

philippotto (approved):
:tada:

PR #392, opened 2021-08-20
Publish wkcuber docs to S3

normanrz:
### Description:
- Publishes the wkcuber docs to a bucket in S3. Example: https://static.webknossos.org/lib-docs/docs_s3/wkcuber/api.html
- Refactors CI to just run the docker builds once

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -119,6 +119,7 @@ jobs:
       run: ./publish.sh
 
   docker:
+    needs: build
     runs-on: ubuntu-latest
 
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -164,7 +164,7 @@ jobs:
         CI_BRANCH=${GITHUB_REF##*/}
         NORMALIZED_CI_BRANCH=${CI_BRANCH//[\/-]/_}
         docker tag \
-          scalableminds/webknossos-cuber:$GITHUB_SHA \s
+          scalableminds/webknossos-cuber:$GITHUB_SHA \
           scalableminds/webknossos-cuber:$NORMALIZED_CI_BRANCH
         docker push scalableminds/webknossos-cuber:$NORMALIZED_CI_BRANCH
         if [ "${CI_BRANCH}" == "master" ]; then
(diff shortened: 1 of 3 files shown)


.github/workflows/main.yml
```diff
@@ -95,6 +80,65 @@ jobs:
     - name: Remove reference magnification data
       run: rm -r testdata/tiff_mag_2_reference/
 
+    - name: Generate Docs
+      if: matrix.python-version == '3.7'
+      run: |
+        docs/api.sh --persist
+    
+    - name: Push docs (for branch)
+      if: matrix.python-version == '3.7'
```
philippotto, line 89: should this job check whether only a tag was pushed? or does it simply work in that case, too?

.github/workflows/main.yml
```diff
@@ -95,6 +80,65 @@ jobs:
     - name: Remove reference magnification data
       run: rm -r testdata/tiff_mag_2_reference/
 
+    - name: Generate Docs
+      if: matrix.python-version == '3.7'
```
philippotto, line 84: is there a specific reason for using 3.7 here? I think I would prefer to build the docs with 3.8 if there's no counter argument I'm missing.

philippotto (approved):
sweet stuff :candy:

.github/workflows/main.yml
normanrz: good point!

.github/workflows/main.yml
normanrz: i switched to 3.8

README.md
```diff
@@ -128,6 +128,9 @@ tests/scripts/all_tests.sh
 
 PyPi releases are automatically pushed when creating a new Git tag/Github release. 
 
+## API documentation
+Check out the [latest version of the API documentation](https://static.webknossos.org/lib-docs/master/wkcuber/api.html).
```
philippotto, line 132: :+1: just wanted to suggest that

hotzenklotz: Wrong link? (at least does not work for me) Did you mean `https://static.webknossos.org/lib-docs/docs_s3/wkcuber/api.html`?

normanrz: The link will work, once it is deployed on `master` branch

hotzenklotz: Never mind

PR #394, opened 2021-08-23
Fixes for Windows

### LINKED ISSUE #393: wkcuber fails on Windows
This regex fails to compile on Windows, because of the backslash in `os.path.sep`: https://github.com/scalableminds/webknossos-libs/blob/master/wkcuber/knossos.py#L14-L16

Because the regex is defined in the global scope of the module, other modules, e.g. `wkcuber.__main__` can't start.

normanrz:
### Description:
- Fixes a bug with path-regexes #393 
- Fixes hard-coded path seperators in the tests
- Add windows and mac builds to CI
- Tweaked the tests to be more conservative with peak storage consumption, because the Windows CI runner seems to have less storage available

### Issues:
- fixes #393 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/tests/test_auto_detection.py
+++ b/tests/test_auto_detection.py
@@ -1,7 +1,11 @@
 from wkcuber.converter import ImageStackConverter, KnossosConverter
 from os.path import sep
 
-TEST_PREFIXES = ["", sep, f"..{sep}"]
+TEST_PREFIXES = ["", "/", f"../"]
+
+
+def fix_sep(path: str) -> str:
+    return path.replace("/", sep)
 
 
 def test_tiff_dataset_name_and_layer_name_detection() -> None:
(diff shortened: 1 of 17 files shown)


tests/test_auto_detection.py
```diff
@@ -134,64 +140,58 @@ def test_knossos_dataset_name_and_layer_path_detection() -> None:
         # test if dataset name and layer name and mag are correct
         converter = KnossosConverter()
         converter.source_files = [
-            prefix
-            + "knossos/color/1/x0000/y0000/z0000/test_mag1_x0000_y0000_z0000.raw",
-            prefix
-            + "knossos/color/1/x0000/y0000/z0001/test_mag1_x0000_y0000_z0001.raw",
-            prefix
-            + "knossos/color/1/x0000/y0001/z0000/test_mag1_x0000_y0001_z0000.raw",
+            f"{prefix}knossos{sep}color{sep}1{sep}x0000{sep}y0000{sep}z0000{sep}test_mag1_x0000_y0000_z0000.raw",
```
philippotto, line 143: I find the many `{sep}`s quite noisy. Maybe we should implement a small helper, like this?

```suggestion
            adapt_separators(f"{prefix}knossos/color/1/x0000/y0000/z0000/test_mag1_x0000_y0000_z0000.raw"),
```

and

```
def adapt_separators(s):
  return s.replace('/', os.sep)
```

But I can understand if you want to stick to the current solution.

tests/test_utils.py
```diff
@@ -44,9 +46,9 @@ def test_get_regular_chunks_max_inclusive() -> None:
 def test_buffered_slice_writer() -> None:
     test_img = np.arange(24 * 24).reshape(24, 24).astype(np.uint16) + 1
     dtype = test_img.dtype
-    bbox = {"topleft": (0, 0, 0), "size": (24, 24, 35)}
+    # bbox = {"topleft": (0, 0, 0), "size": (24, 24, 35)}
```
philippotto, line 49: can it be removed?

philippotto (approved):
great work :+1:

PR #403, opened 2021-09-01
Publish docs for webknossos and wkcuber

normanrz:
See https://static.webknossos.org/lib-docs/docs_s3/index.html


jstriebel (approved):
:+1:

PR #404, opened 2021-09-02
Make parameter `executor` optional for `View.for_each_chunk`

### LINKED ISSUE #400: executor arg should be optional
I was just using the `MagView.for_each_chunk` function, which  is a great utility. However, I would like that the `executor` arg is optional. If it is `None`, a simple for loop should be used.

Also, a few comments about the code  sample in the docs:

    The view is chunked into multiple sub-views of size `chunk_size`.
    Then, `work_on_chunk` is performed on each sub-view.
    Besides the view, the counter 'i' is passed to the 'work_on_chunk',
    which can be used for logging. Additional parameter for 'work_on_chunk' can be specified.
    The computation of each chunk has to be independent of each other.
    Therefore, the work can be parallelized with `executor`.

    If the `View` is of type `MagView`, only the bounding box from the properties is chunked.

    Example:
    ```python
    from wkcuber.utils import get_executor_for_args, named_partial

    def some_work(args: Tuple[View, int], some_parameter: int) -> None:
        view_of_single_chunk, i = args
        # perform operations on the view
        ...

    # ...
    # let 'mag1' be a `MagView`
    view = mag1.get_view()
    with get_executor_for_args(None) as executor:
        func = named_partial(advanced_chunk_job, some_parameter=42)
        view.for_each_chunk(
            func,
            chunk_size=(100, 100, 100),  # Use mag1._get_file_dimensions() if the size of the chunks should match the size of the files on disk
            executor=executor,
        )
    ```

* `get_executor_for_args`  shouldn't be used in a sample, because it is hard to construct for end-users in a script
* the argument is called `work_on_chunk`. the sample function is  called `some_work`.  it is invoked as `advanced_chunk_job` (bug). maybe this can be unified

rschwanhold:
### Description:
- If the parameter `executor` is not specified for `View.for_each_chunk` or `View.for_zipped_chunks`, the chunks are executed sequentially. 

### Issues:
- fixes #400

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/view.py
+++ b/webknossos/webknossos/dataset/view.py
@@ -296,7 +296,9 @@ class View:
         self,
         work_on_chunk: Callable[[Tuple["View", int]], None],
         chunk_size: Tuple[int, int, int],
-        executor: Optional[Union[ClusterExecutor, cluster_tools.WrappedProcessPoolExecutor]] = None,
+        executor: Optional[
+            Union[ClusterExecutor, cluster_tools.WrappedProcessPoolExecutor]
+        ] = None,
     ) -> None:
         """
         The view is chunked into multiple sub-views of size `chunk_size`.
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/dataset/view.py
```diff
@@ -439,7 +445,11 @@ def for_zipped_chunks(
             job_args.append((source_chunk_view, target_chunk_view, i))
 
         # execute the work for each pair of chunks
-        wait_and_ensure_success(executor.map_to_futures(work_on_chunk, job_args))
+        if executor is None:
+            for args in job_args:
+                work_on_chunk(args)
```
rschwanhold, line 450: I added this as a simple for-loop (as stated in the issue). An alternative would be to use `get_executor_for_args(None)` per default.

normanrz (approved):

PR #389, opened 2021-08-19
dataset.add_symlink_layer handles Layer argument

valentin-pinkau:
### Description:
- This PR adds the option to pass a Layer argument to the dataset.add_symlink_layer method

### Issues:
- fixes https://github.com/scalableminds/webknossos-cuber/issues/386

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [x] Updated Changelog

PR #406, opened 2021-09-03
Add option to not resolve symlink when adding a foreign layer

rschwanhold:
### Description:
- When adding a foreign layer, the path to the layer was previously resolved. This caused problems if the foreign layer was only a symlink and the original dataset did not contain a datasource-properties.json
- The parameter `resolve_symlinks` defaults to `True`, which is the previous behavior

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -385,7 +385,10 @@ class Dataset:
         self._export_as_json()
 
     def add_symlink_layer(
-        self, foreign_layer: Union[str, Path, Layer], make_relative: bool = False, resolve_symlinks: bool = True
+        self,
+        foreign_layer: Union[str, Path, Layer],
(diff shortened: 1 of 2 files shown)


daniel-wer (commented):
@rschwanhold As far as I can see, the symlink resolving was only introduced in https://github.com/scalableminds/webknossos-libs/pull/389 (since abspath which was used before doesn't resolve symlinks). Not sure whether that was intentional, but if not we could simply remove it again. If it was intentional I would argue the `resolve_symlinks` flag should be `False` by default to mirror the behavior before the breaking change. /cc @valentin-pinkau  @jstriebel

daniel-wer:
I talked to @jstriebel  and @valentin-pinkau and we agreed to restore the functionality from before #389. So, use `os.path.abspath(foreign_layer_path)` and get rid of the symlink resolving. I'll create a new PR for that.

PR #408, opened 2021-09-06
Revert symlink resolving in add_{symlink,copy}_layer

daniel-wer:
### Description:
- See discussion in https://github.com/scalableminds/webknossos-libs/pull/406
- I successfully tested this with the msem spine head regression test which, due to the regression, failed before.

### Todos:
 - [x] Updated Changelog


valentin-pinkau (approved):
LGTM!

PR #410, opened 2021-09-07
wkcuber: use venv by default (except for Dockerfile & win build)

jstriebel:
### Description:
wkcuber CI was failing with errors such as `FileNotFoundError: [Errno 2] No such file or directory: '…/python3.8/site-packages/packaging-20.9.dist-info/METADATA`. This PR changes the poetry configuration to use a venv by default (except for docker), which worked fine on mac & linux. Only the windows build fails when using a venv with weird errors, so not using it there.

If you want to use your default Python env or another managed conda/venv/… env, simply run to configure the local repo `poetry config virtualenvs.create false --local` or without `--local` to have it in your user-wide settings.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -131,6 +131,7 @@ jobs:
     - name: Install dependencies
       shell: bash
       run: |
+        pip install -U pip
         pip install poetry
         poetry config virtualenvs.create false --local
         poetry install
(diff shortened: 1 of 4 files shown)


rschwanhold (approved):
LGTM.

Are the changes only relevant for the CI or should everyone now use `poetry run ./typecheck.sh` instead of `./typecheck.sh`?

jstriebel:
@rschwanhold Good point! If you want to use poetry venv (and not have your personal `poetry.toml`), you will need to use `poetry run ./typecheck.sh`.

Issue #381, opened 2021-08-13
[wk-lib] add webknossos-REST client


jstriebel:
We want to have a REST-client to down- and upload typical datastructures from webknossos, such as datasets and annotations to start with.
If https://github.com/scalableminds/webknossos/issues/5643 is possible, we can auto-generate a client for the endpoints, e.g. using  https://github.com/openapi-generators/openapi-python-client. This generates clients using attrs and httpx, which seems like an excellent choice, e.g. see [this example](https://github.com/openapi-generators/openapi-python-client/tree/main/end_to_end_tests/golden-record).
The client should then be made available via simple top-level functions, see [the DesignDoc exmples](https://www.notion.so/scalableminds/Design-Doc-wk-tooling-API-48345819ef3a42cfae19fe49a6a6c26e).

PR #411, opened 2021-09-07
Generated client v1

jstriebel:
### Description:
Adds the basic structure for the webknossos client code generation, such as:
- The `__generate_client.py` script.
- Small helpers to get a client and tokens (will be extended later)
- Two small test cases for the generated client.

### Future Work (in extra PR)
- Rename wk paths (PR is WIP) to get better names
- Add more response schemas
- Add user-facing wrapper functions (the generated client purely serves as a helper)

### Issues:
- part of #381

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,6 +41,7 @@ jobs:
 
     - name: Install dependencies
       run: |
+        pip install -U pip
         pip install poetry
         poetry install
 
@@ -84,6 +85,7 @@ jobs:
 
     - name: Install dependencies
       run: |
+        pip install -U pip
         pip install poetry
         poetry install
 
@@ -134,6 +136,7 @@ jobs:
     - name: Install dependencies
       shell: bash
       run: |
+        pip install -U pip
         pip install poetry
         poetry install
 
@@ -168,6 +171,7 @@ jobs:
 
     - name: Install dependencies
       run: |
+        pip install -U pip
         pip install poetry
         poetry install
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -41,7 +41,6 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install -U pip
         pip install poetry
         poetry install
 
@@ -85,7 +84,6 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install -U pip
         pip install poetry
         poetry install
 
@@ -136,7 +134,6 @@ jobs:
     - name: Install dependencies
       shell: bash
       run: |
-        pip install -U pip
         pip install poetry
         poetry install
 
@@ -171,7 +168,6 @@ jobs:
 
     - name: Install dependencies
       run: |
-        pip install -U pip
         pip install poetry
         poetry install
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -55,6 +55,8 @@ jobs:
         poetry run ./typecheck.sh
 
     - name: Python tests
+      env:
+        WK_TOKEN: ${{ secrets.WK_TOKEN }}
       run: poetry run pytest tests
 
     - name: Check if git is dirty
--- a/webknossos/__generate_client.py
+++ b/webknossos/__generate_client.py
@@ -38,7 +38,10 @@ def generate_client(openapi_schema: Dict):
 
 
 def add_api_prefix_for_non_data_paths(openapi_schema: Dict):
-    """TODO"""
+    """The current webKnossos backend does not include the
+    /api prefix into the different backend paths.
+    Howevery, the /data prefix for datastore paths is included.
+    This adds the missing /api prefixes in the openapi_schema (in-place)."""
     assert_valid_schema(openapi_schema)
     paths = openapi_schema["paths"]
     for path in list(paths.keys()):
@@ -99,7 +102,8 @@ def set_response_schema_by_example(
 
 
 def bootstrap_response_schemas(openapi_schema: Dict) -> None:
-    """TODO"""
+    """Inserts the response schemas into openapi_schema (in-place),
+    as recorded by example requests."""
     assert_valid_schema(openapi_schema)
     for operation_id, response in iterate_request_ids_with_responses():
         set_response_schema_by_example(
--- a/webknossos/__generate_client.py
+++ b/webknossos/__generate_client.py
@@ -1,0 +1,117 @@
+import calendar
+import json
+from datetime import datetime
+from pathlib import Path
+from tempfile import NamedTemporaryFile
+from typing import Dict, Iterable, Tuple
+
+import httpx
+from inducoapi import build_openapi
+from openapi_python_client import Config, MetaType, _get_project_for_url_or_path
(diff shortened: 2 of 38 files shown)


webknossos/webknossos/client/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+
+load_dotenv()
+
+
+@lru_cache(maxsize=None)
+def _get_token(webknossos_url: str) -> str:
+    if "WK_TOKEN" in os.environ:
+        return os.environ["WK_TOKEN"]
+    else:
+        # TODO  pylint: disable=fixme
```
philippotto, line 18: todo

webknossos/webknossos/client/__init__.py
```diff
@@ -0,0 +1,37 @@
...
+        # -ask if the token should be saved in some .env file
+        return Prompt.ask(
+            f"\nPlease enter your webknossos token as shown on {webknossos_url}/auth/token ",
+            password=True,
+        )
+
+
+@lru_cache(maxsize=None)
+def _get_generated_client(
+    webknossos_url: str = DEFAULT_WEBKNOSSOS_URL, with_token: bool = False
```
philippotto, line 29: What do you think about using the token by default if it's in the ENV? The parameter could be `enforce_token: bool = False` with the semantic that the user prompt is only shown if `enforce_token` is true and there is no WK_TOKEN in the env.. No strong opinion, though.

.github/workflows/ci.yml
```diff
@@ -55,6 +55,8 @@ jobs:
         poetry run ./typecheck.sh
 
     - name: Python tests
+      env:
+        WK_TOKEN: ${{ secrets.WK_TOKEN }}
```
philippotto, line 59: what kind of token is used here? does it belong to a separate dummy organization on webknossos.org? Alternatively, the tests could be hooked up with https://master.webknossos.xyz/

webknossos/tests/test_generated_client.py
```diff
@@ -0,0 +1,48 @@
...
+    )
+    assert info_object is not None
+    assert info_object.id == id
+    assert info_object.typ == typ
+    assert not isinstance(info_object.data_store, Unset)
+    assert info_object.data_store.url == client.base_url
+
+
+def test_list_datastores() -> None:
+    datastores = list_.sync(client=auth_client)
```
philippotto, line 38: `list` is subject for change and meant with your statement here, right?

> Rename wk paths (PR is WIP) to get better names

webknossos/tests/test_generated_client.py
```diff
@@ -0,0 +1,48 @@
...
+from webknossos.client.generated.types import Unset
+
+
+def time_since_epoch_in_ms() -> int:
+    d = datetime.utcnow()
+    unixtime = calendar.timegm(d.utctimetuple())
+    return unixtime * 1000
+
+
+WK_URL = "https://webknossos.org"
```
philippotto, line 18: see my other comment. maybe we don't want to run the tests against webknossos.org? but maybe it's alright.

webknossos/__generate_client.py
```diff
@@ -0,0 +1,121 @@
...
+            package_name_override=".",
+        )
+        generator_project = _get_project_for_url_or_path(
+            url=None,
+            path=Path(schema_file.name),
+            meta=MetaType.POETRY,
+            config=generator_config,
+        )
+        errors = generator_project.update()
+        # handle_errors(errors)  # prints warnings
```
philippotto, line 37: remove comment?

philippotto (approved):
cool stuff! looks good to me :+1:

webknossos/webknossos/client/__init__.py
jstriebel: I'd leave this for a future PR

philippotto: ok

webknossos/__generate_client.py
jstriebel: This might help if the generation fails somehow, therefore I'd leave it in.

webknossos/webknossos/client/__init__.py
jstriebel: I think that's exactly what `with_token` is doing, do you agree? But `enforce_token` or `needs_token` is a much better name, I'll adapt it. I'd be happy to discuss this further in a broader sense anyway (e.g. should we do user-wide caching?)

.github/workflows/ci.yml
jstriebel: Yes, this is for a separate dummy user on webknossos.org, with it's own Test orga. I've no strong opinion if we rather should use https://master.webknossos.xyz, as long as they are in sync. If we hold back a version from deployment using wk.org might be the better option. Also, wk.org has extra datastores, which is helpful for demos and testing. /cc @fm3 Do you have an opinion on this?

webknossos/tests/test_generated_client.py
jstriebel: Exactly :+1:

.github/workflows/ci.yml
fm3: I think both instances should only be used for things that have no side effects, though errors in the tests (or webknossos’ handling of them) may be more damaging on webknossos.org. I’d tend to prefer the non-production instance for tests, in order to reduce load and risk, but I did not read the code here so it may be safe enough.

Note that master.webknossos.xyz is updated nightly to the current master branch, while wk.org is *usually* updated directly after merge, but is sometimes rolled back if there is a problem with the master, or sometimes the deployment is held back a few days if we want to test on other instances first.

webknossos/webknossos/client/__init__.py
philippotto: > I think that's exactly what with_token is doing, do you agree?

I might be missing something, but currently `_get_generated_client()` will not use any tokens from ENV even if these exist. Instead, I have to explicitly pass `with_token=True`. My suggestion is to always use an existing token. If no token exists **and** `enforce_token=True`, the prompt will come up. Does this make sense?

> I'd be happy to discuss this further in a broader sense anyway (e.g. should we do user-wide caching?)

User-wide caching sounds like a nice-to-have. However, probably not high-pri for now and also we should communicate this clearly with the user so that tokens are not written to disk without their knowledge.

.github/workflows/ci.yml
philippotto: Since the test suite is likely to grow in the future, a strict separation from production instances sounds like a good strategy to me (with respect to load, risk, resources/garbage etc.).

> as long as they are in sync.

We could probably setup a dev instance which always has the same version as wkorg.

> Also, wk.org has extra datastores, which is helpful for demos and testing. 

Ideally, we would extend the dev instance so that it also has these things. However, I can understand that some effort is needed for this.

I'd be fine with leaving it as is for now, but in the long-term picture I'd prefer the separation explained above.

jstriebel: Sound's good, I'll adapt this in one of the next PRs

webknossos/webknossos/client/__init__.py
jstriebel: Good point, I adapted it in the newest commit

PR #412, opened 2021-09-07
Fix assert bounds in Dataset API

rschwanhold:
### Description:
- use the correct bounding box to assert bounds


rschwanhold:
This is actually not a bug. I was just confused because the offset was relative to the view instead of being the global offset.

jstriebel:
> This is actually not a bug. I was just confused because the offset was relative to the view instead of being the global offset.

Thanks, that was exactly what I was just wondering about. Please try to make the difference clearer in the code which variables are the views global offset, offsets within the view and the addition of both (by names and or comments).

PR #413, opened 2021-09-08
Windows CI: don't use venv

jstriebel:
### Description:
The Windows CI run was unreliable with similar errors as the linux and mac runs before, so we're also using the default poetry venv here now.


philippotto (approved):
lgtm

Issue #380, opened 2021-08-13
[wk-lib] add new Skeleton implementation


jstriebel:
Add a new skeleton implementation which complies with [the Design Doc](https://www.notion.so/scalableminds/Design-Doc-wk-tooling-API-48345819ef3a42cfae19fe49a6a6c26e?d=c833454c-8b43-451c-a6ea-da7e2d4301ba#beeacbfcd83e40ec85fc3109dbfbbce4). This might either be built on top of [the current wknml-lib](https://github.com/scalableminds/wknml) using loxun, or using [the std-lib ElementTree](https://docs.python.org/3/library/xml.etree.elementtree.html), or [lxml](https://lxml.de/).

jstriebel:
Further improvements after the initial implementation (#416) that we should look into:
* cleanups in the NML import / export
* should skeleton.Graph inherit from nx.Graph?
* discourage use of Graph/Group/Node inits (e.g. via warnings?)
* add warning if exported graph is not a tree, add conversion methods
* check that _enforced_id is unused

PR #416, opened 2021-09-08
Skeleton/NML API

philippotto:
### Description:
- Adds new Skeleton/NML API which internally uses networkx graphs. See `test_skeleton.py` for the example usage.
- The old wknml implementation was partially added as a `legacy` module. However, I adapted some parts (partly while I added typing). In the long run, the `legacy` module could probably be simply renamed/rebranded to the xml/nml scope.

### Issues:
- contributes to #380

### Todos:
 - [ ] Updated Changelog


philippotto:
@jstriebel I think you can have a first look at this PR. There are still some details missing, but the big picture should be alright.

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+    def get_node_positions(self) -> np.ndarray:
+        return np.array([node.position for node in self.nx_graph.nodes])
+
+    def get_node_by_id(self, node_id: int) -> Node:
+
+        # Todo: Use hashed access
+        for node in self.get_nodes():
+            if node.id == node_id:
+                return node
+        raise ValueError(f"No node with id {node_id} was found")
```
jstriebel, line 209: If we simply subclass `nx.Graph`, we might also want to enable lookups by id like this:
```python
node = g.nodes[3]
```

I think this might be possible using some magic, and I think it would have a better UX. What do you think about this?

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+            branchpoint_time=branchpoint_time,
+            enforce_id=_enforce_id,
+            nml=_nml or self._nml,
+        )
+        self.nx_graph.add_node(node)
+        return node
+
+    def add_edge(self, node_1: Node, node_2: Node) -> None:
+
+        self.nx_graph.add_edge(node_1, node_2)
```
jstriebel, line 247: I believe subclassing would be great ;-)

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+
+    def get_max_node_id(self) -> int:
+
+        # Chain with [0] since max is not defined on empty sequences
+        return max(itertools.chain((node.id for node in self.nx_graph.nodes), [0]))
```
jstriebel, line 252:
```suggestion
        return max(node.id for node in self.nx_graph.nodes, default=0)
```

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+
+    def __hash__(self) -> int:
+        return hash((self._nml.id, self.id))
```
jstriebel, line 255: Should we also re-define equality?

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+    return (
+        float(vec[0]),
+        float(vec[1]),
+        float(vec[2]),
+    )
+
+
+@attr.define()
+class Group:
+    id: int = attr.ib(init=False)
```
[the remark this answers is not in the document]
jstriebel: PS: Also, we should discourage the user to change the id. So we might use `__id` or `_id` as the property, `id` for the init signature, and have a read-only `id` property, which returns `self.__id`.

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+    scale: Vector3
+    offset: Optional[Vector3] = None
+    time: Optional[int] = None
+    edit_position: Optional[Vector3] = None
+    edit_rotation: Optional[Vector3] = None
+    zoom_level: Optional[float] = None
+    task_bounding_box: Optional[IntVector6] = None
+    user_bounding_boxes: Optional[List[IntVector6]] = None
+
+    root_group: Group = attr.ib(init=False)
```
jstriebel, line 275: I was wondering if we could also subclass `Group` here directly, this would reduce a lot of the boiler-plate code and makes sense semantically IMO. Then we'd need a separate attribute for the `nml_id`, and the normal `id` is just the one of the group.

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+    element_id_generator: Iterator[int] = attr.ib(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        self.id = nml_id_generator.__next__()
```
jstriebel, line 279: I'm not sure if we want to use a global counter, since this can lead to problems with threading, pickling, etc… Instead we could use `id(self)` where we had `id` before:
https://docs.python.org/3/library/functions.html#id

jstriebel (commented):
Awesome work :rocket: :man_singer: :tada: 

I added quite a number of comments, some cosmetics and some more about initialization and inheritance. Might be worth to discuss in a call :call_me_hand:

> In the long run, the legacy module could probably be simply renamed/rebranded to the xml/nml scope.

I think this would be great :+1: If that's easy, I'd vote to just rename it to `nml` in this PR.

You can also mark TODOs which won't be resolved right now with `pylint: disable=fixme`.

jstriebel:
PS: @fm3's PR #421 introduces Vec3Int and Vec3Float classes which we should use in the Skeleton API as well.

webknossos/pyproject.toml
```diff
@@ -17,6 +17,9 @@ python-dateutil = "^2.8.0"
 scikit-image = "^0.16.0"
 scipy = "^1.4.0"
 wkw = "1.1.11"
+networkx = "^2.6.2"
+loxun = "^2.0"
+icecream = "^2.1.1"
```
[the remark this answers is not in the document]
philippotto: good point. i moved it to dev dependencies now.

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+    def get_total_node_count(self) -> int:
+        return sum(
+            len(synapse_graph.get_nodes()) for synapse_graph in self.flattened_graphs()
+        )
+
+    def get_max_graph_id(self) -> int:
+        # Chain with [0] since max is not defined on empty sequences
+        return max(
+            itertools.chain((graph.id for graph in self.flattened_graphs()), [0])
+        )
```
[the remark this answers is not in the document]
philippotto: ah, cool :facepalm:

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+
+        return self.nx_graph.nodes
+
+    def get_node_positions(self) -> np.ndarray:
+        return np.array([node.position for node in self.nx_graph.nodes])
+
+    def get_node_by_id(self, node_id: int) -> Node:
+
+        # Todo: Use hashed access
+        for node in self.get_nodes():
```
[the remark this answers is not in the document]
philippotto: This is done  now :man_dancing: I just managed to fix my then broken approach to do this. Initially, I forgot to adapt the `add_edge` code and networkx then implicitly added the node objects (since these were passed instead of ids) to the graph, which is why half the nodes were broken.

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+        float(vec[2]),
+    )
+
+
+def vector3_as_float(vec: Tuple[Number, Number, Number]) -> Vector3:
+    return (
+        float(vec[0]),
+        float(vec[1]),
+        float(vec[2]),
+    )
```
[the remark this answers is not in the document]
philippotto: I'm not sure what you mean? The function itself is not for the sake of typing, but for enforcing floats which is important for the roundtrip-test (nml -> skeleton -> nml) to succeed (otherwise, there are differences between floats and ints in the NML files). There's probably another way to fix this, but this seemed like an okay solution for now.

webknossos/tests/test_skeleton.py
```diff
@@ -0,0 +1,164 @@
...
+
+    g = nml.add_graph(
+        "A WkGraph",
+        color=(0.9988844805996959, 0.09300433970039235, 0.13373766240135082, 1.0),
+    )
+
+    n1 = g.add_node(
+        position=(0, 1, 2),
+        comment="A comment 1",
+        is_branchpoint=True,
```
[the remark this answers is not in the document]
philippotto: > This is something we set in wk, right?

Yes.

> Does it mean this node has more than 2 edges attached? Or rather here is a biological branchpoint which might need more exploration?

Kind of both. In webKnossos you can mark a node explicitly as a branch point which then allows to jump back to that node later (typically when the first branch was explored). So, typically a branchpoint has more than 2 edges (after tracing was complete), but not necessarily. Also branchpoints are highlighted in the UI etc.

webknossos/webknossos/skeleton/__init__.py
```diff
@@ -0,0 +1,5 @@
+from webknossos.skeleton.entities import NML, Group, Node, WkGraph
+
+
+def open_nml(file_path: str) -> "NML":
+    return NML.from_path(file_path)
```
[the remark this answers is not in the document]
philippotto: Sounds good, but there is no top-level `__init__.py` currently, is there?

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+        else:
+            self.id = self._nml.element_id_generator.__next__()
+
+        self.position = vector3_as_float(self.position)
+
+    def __hash__(self) -> int:
+        return hash((self._nml.id, self.id))
+
+    def __eq__(self, other: object) -> bool:
+        return hash(self) == hash(other)
```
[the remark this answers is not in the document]
philippotto: These methods are not necessary anymore, since I changed the nxgraph usage to use ids as keys. Therefore, I would  simply remove the methods and rely on the python default behavior. Do you agree?

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,473 @@
...
+        float(vec[2]),
+    )
+
+
+def vector3_as_float(vec: Tuple[float, float, float]) -> Vector3:
+    return (
+        float(vec[0]),
+        float(vec[1]),
+        float(vec[2]),
+    )
```
jstriebel, line 42: I think those two function should not be necessary at all anymore, do you agree? They explicitly convert ints to floats, but I don't think that this should change anything.

webknossos/webknossos/skeleton/__init__.py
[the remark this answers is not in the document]
jstriebel: Yes, this one:
https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/__init__.py

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,479 @@
...
+
+
+@attr.define()
+class Group:
+    id: int = attr.ib(init=False)
+    name: str
+    children: List[GroupOrGraph]
```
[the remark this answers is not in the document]
philippotto: make attribute immutable and only expose iterator so that the caller cannot call `append` on that

webknossos/webknossos/skeleton/entities.py
philippotto: remove counter

webknossos/webknossos/skeleton/entities.py
jstriebel: Let's remove the hashing for now as discussed :+1:

webknossos/webknossos/skeleton/entities.py
jstriebel: Let's leave it as-is for now, we might look into nx-graph subclassing again later.

webknossos/webknossos/skeleton/entities.py
jstriebel: We'll leave it like this for now.

webknossos/webknossos/skeleton/entities.py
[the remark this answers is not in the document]
philippotto: I made the ids and children read-only for now. I think, this is what we decided on during our discussion yesterday :)

webknossos/tests/test_skeleton.py
```diff
@@ -0,0 +1,164 @@
...
+
+        g6 = nml.get_graph_by_id(6)
+        assert g6.name == "Graph in Group"
+        assert g6.get_node_by_id(7).position == (10.0, 3.0, 4.0)
+
+        nml.write(export_path)
+        diff_files(snapshot_path, export_path)
+
+
+def test_code_example() -> None:
```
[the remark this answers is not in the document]
philippotto: I had to move the examples folder into `webknossos/examples` so that I could import them from the tests. I also followed your `WIP` suggestion.

philippotto:
@jstriebel I think I covered nearly all of the todos. Please have another look and feel free to merge the PR while I'm on vacation. I don't know why the CI currently fails, but the merge conflict with the `poetry.lock` file also need to be resolved. Would be great if you could do that, since I don't have much time left for today.
Also feel free to change the PRs code if you feel the need when looking at my latest changes :see_no_evil:

webknossos/webknossos/skeleton/entities.py
```diff
@@ -0,0 +1,440 @@
...
+    zoom_level: Optional[float] = None
+    task_bounding_box: Optional[IntVector6] = None
+    user_bounding_boxes: Optional[List[IntVector6]] = None
+
+    root_group: Group = attr.ib(init=False)
+    element_id_generator: Iterator[int] = attr.ib(init=False)
+
+    def __attrs_post_init__(self) -> None:
+        self.element_id_generator = itertools.count()
+        self.root_group = Group(name="Root", children=[], nml=self, is_root_group=False)
```
hotzenklotz, line 254: Not sure if correct/incorrect, but I stumbled upon it just now: `self.root_group` ist initialized as `is_root_group=False`... this feels wrong ;-)

webknossos/webknossos/skeleton/entities.py
jstriebel: I tried out the subclassing, and it worked really nicely :+1:

webknossos/webknossos/skeleton/entities.py
jstriebel: This was incorrect before, but is completely removed, since the Skeleton now inherits from Group directly.

jstriebel:
I'm quite happy with the current state of the PR. There are of course more open points, but I'd defer them to another PR later:
* further cleanups in the NML import / export
* should skeleton.Graph inherit from nx.Graph?
* discourage use of Graph/Group/Node inits (e.g. via warnings?)
* add warning if exported graph is not a tree, add conversion methods
* check that _enforced_id is unused

I noted those in the original issue #380

jstriebel (approved):

jstriebel:
See https://static.webknossos.org/lib-docs/nml_lib/api/webknossos/skeleton.html

PR #417, opened 2021-09-09
Improve generated REST API

jstriebel:
### Description:
- Uses the newest swagger.json which includes nicer names, a `datastore` tag (used as the module name in the generated client) and fixes that the timestamp is required now.
- Adds more response schemas.
- Adds some more tests.


webknossos/__generate_client.py
```diff
@@ -58,25 +58,45 @@ def add_api_prefix_for_non_data_paths(openapi_schema: Dict) -> None:
...
+    assert dataset_info_response.status_code == 200
+    yield "datasetInfo", dataset_info_response.content
+
+    for api_endpoint in [datastore_list, build_info]:
+        api_endpoint_name = api_endpoint.__name__.split(".")[-1]
+        # convert name to camelCase
+        api_endpoint_name_parts = api_endpoint_name.split("_")
+        api_endpoint_name = "".join(
+            s.title() if i > 0 else s for i, s in enumerate(api_endpoint_name_parts)
+        )
```
philippotto, line 95: you can use https://github.com/scalableminds/webknossos-libs/blob/bf278931b83276a9a8ea331c52ae34483ef01589/webknossos/webknossos/dataset/properties.py#L60-L60

philippotto (approved):
:bee: :bee: :bee:

PR #418, opened 2021-09-11
Use Dataset API inside `cubing`

rschwanhold:
### Description:
- abc

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/wkcuber/wkcuber/convert_image_stack_to_wkw.py
+++ b/wkcuber/wkcuber/convert_image_stack_to_wkw.py
@@ -1,7 +1,5 @@
-from wkcuber import downsample_mags
 from .cubing import cubing, create_parser as create_cubing_parser
 from .compress import compress_mag_inplace
-from .metadata import write_webknossos_metadata, refresh_metadata
 from .utils import (
     add_isotropic_flag,
     setup_logging,
--- a/wkcuber/wkcuber/convert_image_stack_to_wkw.py
+++ b/wkcuber/wkcuber/convert_image_stack_to_wkw.py
@@ -55,7 +55,7 @@ def main(args: Namespace) -> None:
 
     arg_dict = vars(args)
 
-    bounding_box = cubing(
+    ds = cubing(
         args.source_path,
         args.target_path,
         args.layer_name,
@@ -70,29 +70,37 @@ def main(args: Namespace) -> None:
         args,
     )
 
-    write_webknossos_metadata(
-        args.target_path,
-        args.name,
-        args.scale,
-        compute_max_id=False,
-        exact_bounding_box=bounding_box,
-    )
+    #write_webknossos_metadata(
+    #    args.target_path,
+    #    args.name,
+    #    args.scale,
+    #    compute_max_id=False,
+    #    exact_bounding_box=bounding_box,
+    #)
 
     if not args.no_compress:
-        compress_mag_inplace(args.target_path, args.layer_name, Mag(1), args)
+        compress_mag_inplace(args.target_path, args.layer_name, Mag(1), args)  # TODO: why is this always Mag(1) and not target_mag?
 
-    downsample_mags(
-        path=args.target_path,
-        layer_name=args.layer_name,
-        from_mag=Mag(1),
+    ds.get_layer(args.layer_name).downsample(
+        from_mag=Mag(1),  # TODO: same here
         max_mag=None if args.max_mag is None else Mag(args.max_mag),
-        interpolation_mode="default",
         compress=not args.no_compress,
         sampling_mode=args.sampling_mode,
         args=get_executor_args(args),
     )
 
-    refresh_metadata(args.target_path)
+    #downsample_mags(
+    #    path=args.target_path,
+    #    layer_name=args.layer_name,
+    #    from_mag=Mag(1),
+    #    max_mag=None if args.max_mag is None else Mag(args.max_mag),
+    #    interpolation_mode="default",
+    #    compress=not args.no_compress,
+    #    sampling_mode=args.sampling_mode,
+    #    args=get_executor_args(args),
+    #)
+
+    #refresh_metadata(args.target_path)
 
 
 if __name__ == "__main__":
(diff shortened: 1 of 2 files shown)


wkcuber/wkcuber/cubing.py
```diff
@@ -299,18 +293,24 @@ def cubing(
         batch_size = BLOCK_LEN
 
     target_mag = Mag(target_mag_str)
-    target_wkw_info = WkwDatasetInfo(
-        target_path,
+
+    target_ds = Dataset.get_or_create(target_path, scale=scale)
+    target_layer = target_ds.get_or_add_layer(
         layer_name,
+        LayerCategories.COLOR_TYPE,
```
rschwanhold, line 300: Is this always a color layer?

wkcuber/wkcuber/cubing.py
```diff
@@ -319,25 +319,28 @@ def cubing(
 
     logging.info("Found source files: count={} size={}x{}".format(num_z, num_x, num_y))
 
-    ensure_wkw(target_wkw_info)
-
     with get_executor_for_args(executor_args) as executor:
         job_args = []
         # We iterate over all z sections
         for z in range(start_z, num_z + start_z, BLOCK_LEN):
```
rschwanhold, line 325: Why is it enough to use the `block_len` (instead of `block_len*file_len`)?

wkcuber/wkcuber/convert_image_stack_to_wkw.py
```diff
@@ -68,32 +66,20 @@ def main(args: Namespace) -> None:
...
-    write_webknossos_metadata(
-        args.target_path,
-        args.name,
         args.scale,
-        compute_max_id=False,
-        exact_bounding_box=bounding_box,
     )
 
     if not args.no_compress:
         compress_mag_inplace(args.target_path, args.layer_name, Mag(1), args)
```
rschwanhold, line 73: Shouldn't this be `args.target_mag` instead of `Mag(1)`? (Same few lines below)

wkcuber/wkcuber/cubing.py
fm3: I think what you want to avoid is concurrent writes to the same block/bucket. Writing to the same wkw file should be fine, as long as the writes are in distinct buckets. So parallelizing over BLOCK_LEN should be ok. Does this make sense?

wkcuber/wkcuber/cubing.py
```diff
@@ -319,25 +319,28 @@ def cubing(
...
             # Prepare z batches
```
fm3, line 326: If I see this correctly you don’t prepare z batches anymore, but rather pass the views directly?  Maybe the comments can be adapted

wkcuber/wkcuber/cubing.py
fm3: I would assume that it can also be segmentation, there should have been some heuristic (or user parameter) in place deciding about that. I can’t find that currently, though. Can you comment on that @normanrz or @youri-k ?

wkcuber/wkcuber/cubing.py
```diff
@@ -165,8 +161,7 @@ def cubing_job(
     ]
 ) -> None:
     (
-        target_wkw_info,
-        z_batches,
```
fm3, line 169: my understanding is that `z_batches` was actually a list of individual “dense” `z` values (so basically just one batch) – is that correct? In this case, replacing it by a single view seems fine by me, but maybe there was some specific reason behind `z_batches`? Also compare the old comment at line 184

fm3 (commented):
Great stuff! I’m really looking forward to this :) The code LGTM, but since this is very critical code, and I’m not very familiar with how this worked previously, I’d ask for a second review, possibly by @normanrz ? Are there any specific cases that need testing but are not included in the CI?
Since the CI currently fails at formatting, I am also not certain that there are no tests failing, so that should be fixed too.

wkcuber/wkcuber/cubing.py
rschwanhold: Yes, `z_batches` was a list of "dense" `z`: (`z_batch = list(range(z, max_z))`)

wkcuber/wkcuber/cubing.py
rschwanhold: Ah ok. I thought that concurrent writes to the same file are a problem.

normanrz: > Ah ok. I thought that concurrent writes to the same file are a problem.

It is only a problem, if the files are compressed, which should not be the case here.

wkcuber/wkcuber/cubing.py
rschwanhold: I have updated the comments

wkcuber/wkcuber/cubing.py
normanrz: I think there was a simple heuristic `segmentation if folder_name == "segmentation" else color`, that was implemented in the metadata inferral

rschwanhold: I have now modified the code to create a segmentation layer if the name is `"segmentation"`. However, this also requires the `largest_segment_id`. Currently I have set it simply to `-1`. Should I compute the largest_segment_id at the end, even though this might take a while for very large datasets?

normanrz: I don't know if `-1` is a valid value. I would default to `0` instead.
The cubing tool is reading all data anyway, would it be possible to accumulate the max value or is that too complex to  implement?

wkcuber/wkcuber/convert_image_stack_to_wkw.py
fm3: @youri-k can you comment on this?

wkcuber/wkcuber/cubing.py
rschwanhold: Yes, that makes sense. I have modified `wait_and_ensure_success` to return a value for each cube (in this case the largest_segment_id).

PR #419, opened 2021-09-11
Add option to rename symlink/copy layer

### LINKED ISSUE #415: dataset add_symlink_layer and add_copy_layer should have rename option
use case:
Dataset A has layer “color”, dataset B has another layer, also named “color”

I want to link A’s layer into dataset B, not currently possible because there is a name conflict
I would like to `b.add_symlink_layer(color_a_path, make_relative=True, rename_to="color_from_a")`

(should default to None, which takes the layer name from source dataset as is currently the case)

What do you think?

rschwanhold:
### Description:
- Add optional parameter `new_layer_name`

### Issues:
- fixes #415 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -385,7 +385,10 @@ class Dataset:
         self._export_as_json()
 
     def add_symlink_layer(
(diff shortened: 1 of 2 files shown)


fm3 (approved):
❤️

PR #421, opened 2021-09-15
Unify Properties in Mag and BoundingBox, add Vec3Int

### LINKED ISSUE #407: method naming: `as_np` vs `to_array`
Those are functions in the `Mag` class. Should for consistency possibly also be `as_array`? Also, should it be `as_list` instead, since the method returns a python list?
There is also other methods with `to`, e.g. `to_layer_name`.
What are your thoughts on this?

fm3:
### Description:
- immutable `Vec3Int` class was introduced that holds three integers, can be added/subtracted/subscripted etc, see tests
- Dataset API now takes `Vec3IntLike` (`Union[Vec3Int, Tuple[int, int, int], Tuple[int, ...], np.ndarray, List[int]]`, uses `Vec3Int` internally
- `Vec3Int` can be constructed from any `Vec3IntLike`
- There are breaking changes in
  - `BoundingBox`
    - is now immutable, use convenience methods, e.g. `bb.with_topleft((0,0,0))`
    - properties topleft and size are now Vec3Int instead of np.array, they are each immutable as well
    - all `to_`-conversions return a copy, some were renamed:
    - `to_array` → `to_list`
    - `as_np` → `to_np`
    - `as_wkw` → `to_wkw_dict`
    - `from_wkw` → `from_wkw_dict`
    - `from_config` → `from_config_dict`
    - `as_config` → `to_config_dict`
    - `as_checkpoint_name` → `to_checkpoint_name`
    - `as_tuple6` → `to_tuple6`
    - `as_csv` → `to_csv`
    - `as_named_tuple` → `to_named_tuple`
    - `as_slices` → `to_slices`
    - `copy` → (gone, immutable)

  - `Mag`
    - is now immutable
    - `mag.mag` is now `mag._mag` (considered private, use to_list instead if you really need it as list)
    - all `to_`-conversions return a copy, some were renamed:
    - `to_array` → `to_list`
    - `scale_by` → (gone, immutable)
    - `divide_by` → (gone, immutable)
    - `as_np` → `to_np`

### Issues:
- fixes #407

### Todos:
 - [x] Updated Changelog


webknossos/Changelog.md
```diff
@@ -0,0 +1,55 @@
...
+
+- Breaking changes were introduced for geometry classes in [#421](https://github.com/scalableminds/webknossos-libs/pull/421):
+  - `BoundingBox`
+    - is now immutable, use convenience methods, e.g. `bb.with_topleft((0,0,0))`
+    - properties topleft and size are now Vec3Int instead of np.array, they are each immutable as well
+    - all `to_`-conversions return a copy, some were renamed:
+    - `to_array` → `to_list`
+    - `as_np` → `to_np`
+    - `as_wkw` → `to_wkw_dict`
+    - `from_wkw` → `rom_wkw_dict`
```
jstriebel, line 22:
```suggestion
    - `from_wkw` → `from_wkw_dict`
    - `from_config` → `from_config_dict`
```

webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -12,51 +12,58 @@
...
         return BoundingBox(tuple6[0:3], tuple6[3:6])
 
     @staticmethod
-    def from_tuple2(tuple2: Tuple[Shape3D, Shape3D]) -> "BoundingBox":
+    def from_tuple2(tuple2: Tuple[Vec3IntLike, Vec3IntLike]) -> "BoundingBox":
         return BoundingBox(tuple2[0], tuple2[1])
 
     @staticmethod
-    def from_points(points: Iterable[Shape3D]) -> "BoundingBox":
+    def from_points(points: Iterable[Vec3IntLike]) -> "BoundingBox":
```
jstriebel, line 64:
```suggestion
    def from_points(points: Iterable[Vec3IntLike]) -> "BoundingBox":
    """Returns a bounding box exactly containing all points."""
```

webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -101,7 +108,7 @@ def from_auto(
             else:
                 return BoundingBox.from_csv(obj)
         elif isinstance(obj, dict):
-            return BoundingBox.from_wkw(obj)
+            return BoundingBox.from_wkw_dict(obj)
```
jstriebel, line 111: We might also disambiguate wkw_dicts and config dicts:
```suggestion
            if "size" in obj:
                return BoundingBox.from_config_dict(obj)
            else:
                return BoundingBox.from_wkw_dict(obj)
```

webknossos/webknossos/geometry/mag.py
```diff
@@ -1,87 +1,112 @@
...
-from typing import Any, List
+from typing import Any, List, Optional, Tuple, cast
 
+import attr
 import numpy as np
 
+from .vec3_int import Vec3Int, Vec3IntLike
+
+
+def import_mag(mag_like: Any) -> Vec3Int:
```
jstriebel, line 12: I'd like to mark it as private, this also ensure that it does not show up in the docs.
```suggestion
def _import_mag(mag_like: Any) -> Vec3Int:
```

webknossos/webknossos/geometry/mag.py
```diff
@@ -1,87 +1,112 @@
...
-        if self.mag is None or len(self.mag) != 3:
-            raise ValueError(
-                "magnification must be int or a vector3 of ints or a string shaped like e.g. 2-2-1"
-            )
-
-        for m in self.mag:
-            assert log2(m) % 1 == 0, "magnification needs to be power of 2."
+@attr.frozen
+class Mag:
+    _mag: Vec3Int = attr.ib(converter=import_mag)
```
jstriebel, line 46:
```suggestion
    _mag: Vec3Int = attr.ib(converter=_import_mag)
```

webknossos/webknossos/geometry/mag.py
```diff
@@ -1,87 +1,112 @@
...
+
+def import_mag(mag_like: Any) -> Vec3Int:
+    as_vec3_int: Optional[Vec3Int] = None
+
+    if isinstance(mag_like, int):
+        as_vec3_int = Vec3Int(mag_like, mag_like, mag_like)
+    elif isinstance(mag_like, Vec3Int):
+        as_vec3_int = mag_like
+    elif isinstance(mag_like, list) or isinstance(mag_like, tuple):
+        as_vec3_int = Vec3Int(cast(Vec3IntLike, mag_like))
```
jstriebel, line 20: Using a stricter typing for `mag_like`, this should not be necessary anymore (I hope):
```suggestion
        as_vec3_int = Vec3Int(mag_like)
```

webknossos/webknossos/geometry/mag.py
jstriebel: Also, can we specify `mag_like` more, type-wise? E.g.
```suggestion
MagLike = Union[Mag, Vec3IntLike, int, str]

def import_mag(mag_like: MagLike) -> Vec3Int:
```

webknossos/webknossos/geometry/mag.py
```diff
@@ -1,87 +1,112 @@
...
+    elif isinstance(mag_like, str):
+        if re.match(r"^\d+$", mag_like) is not None:
+            as_vec3_int = Vec3Int(int(mag_like), int(mag_like), int(mag_like))
+        elif re.match(r"^\d+-\d+-\d+$", mag_like) is not None:
+            as_vec3_int = Vec3Int([int(m) for m in mag_like.split("-")])
+    elif isinstance(mag_like, Mag):
+        as_vec3_int = Vec3Int(mag_like.to_tuple())
```
jstriebel, line 27: We could directly return `mag_like` at the start of the function in this case to ensure `Mag(Mag(5))` is as simple as possible, similar as with Vec3Int.

webknossos/webknossos/geometry/mag.py
```diff
@@ -1,87 +1,112 @@
...
 
-    def divide_by(self, d: int) -> None:
-        self.mag = [mag // d for mag in self.mag]
+    def scaled_by(self, factor: int) -> "Mag":
+        return Mag([mag * factor for mag in self._mag])
 
-    def divided_by(self, d: int) -> "Mag":
-        return Mag([mag // d for mag in self.mag])
+    def divided(self, coord: List[int]) -> List[int]:
+        return [c // m for c, m in zip(coord, self._mag)]
```
jstriebel, line 106: For such arithmetics we might also have Mag inherit from Vec3Int directly, but I'm not really sure if this is helpful. Let's defer this discussion to another time ;-)

jstriebel (commented):
Awesome work :rocket: I'm very much in love with this PR :love_letter:



I added a couple of comments, mostly nitpicks or minor changes.

webknossos/tests/test_vec3_int.py
```diff
@@ -0,0 +1,81 @@
...
+
+from webknossos.geometry import Vec3Int
+
+
+def test_import() -> None:
+
+    assert Vec3Int(1, 2, 3) == Vec3Int(1, 2, 3)
+    assert Vec3Int((1, 2, 3)) == Vec3Int(1, 2, 3)
+    assert Vec3Int([1, 2, 3]) == Vec3Int(1, 2, 3)
+    assert Vec3Int([1, 2, 3]) == Vec3Int(1, 2, 3)
```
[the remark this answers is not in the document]
fm3: Iterator/Generator is not currently part of Vec3IntLike. Do you think we should extend it? I removed the redundant line in the test for now.

webknossos/webknossos/geometry/mag.py
fm3: Unfortunately I ran into problems with attr when specifying the import parameter as MagLike (Cannot determine __init__ type from converter). MagLike can only be defined after Mag (as mag is in the Union), but the converter apparently needs a real type 🤷 If you know a trick there, let me know :)

webknossos/webknossos/geometry/mag.py
fm3: The import method needs to return a Vec3IntLike, not a Mag, since this is not the constructor but the converter for the attr.ib. Still, I moved up the `Mag` case in the switch, and removed the superfluous tuple conversion

webknossos/webknossos/geometry/mag.py
fm3: Yeah I am unsure about that as well. Let’s defer

webknossos/tests/test_vec3_int.py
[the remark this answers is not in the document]
fm3: As discussed I added Iterable[int] now

jstriebel (approved):
:green_heart:

PR #428, opened 2021-09-22
Add jpeg2000 support

youri-k:
### Description:
- adds `jp2` support by including it in supported file endings
- fixes the type translation for `signed int` layer

### Todos:
 - [X] Updated Changelog


wkcuber/wkcuber/image_readers.py
```diff
@@ -448,6 +448,7 @@ def __init__(self) -> None:
             ".tiff": TiffImageReader(),
             ".jpg": PillowImageReader(),
             ".jpeg": PillowImageReader(),
+            ".jp2": PillowImageReader(),
```
jstriebel, line 451: https://en.wikipedia.org/wiki/JPEG_2000 suggests more possible file-endings:
.jp2, .j2k, .jpf, .jpm, .jpg2, .j2c, .jpc, .jpx, .mj2

Should we add them all?

youri-k: Sounds good. I'll add them.

youri-k: I just checked whether there are more file endings for JPEG as well, so I added `.jpe, .jfif, .jif` as suggested here https://en.wikipedia.org/wiki/JPEG#JPEG_filename_extensions

jstriebel (approved):
LGTM :+1:

PR #429, opened 2021-09-23
Fix Bug in BoundingBox.to_slices

fm3:
- fix copy paste error introduced in #421 
- remove superfluous Mag.scaled_by


jstriebel (approved):
:+1:

PR #431, opened 2021-09-28
CI: fix conditions on complete step

jstriebel:
### Description:
The `complete` step was skipped previously if other earlier skips failed. Skipped is taken as `succeeded` for mergeability checks. This is changed now, by failing the step if any earlier steps failed or were cancelled.

`complete` should fail with the current master, and succeed when the CI is fixed on master again.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -318,11 +318,13 @@ jobs:
 
   complete:
     needs: [webknossos_linux, wkcuber_linux, wkcuber_win, wkcuber_mac, wkcuber_docker, docs]
-    if: |
-      always() &&
-      !contains(needs.*.result, 'failure') &&
-      !contains(needs.*.result, 'cancelled')
+    if: always()
     runs-on: ubuntu-latest
     steps:
+    - name: Check failure
+      if: |
+        contains(needs.*.result, 'failure') ||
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -318,11 +318,13 @@ jobs:
 
   complete:
     needs: [webknossos_linux, wkcuber_linux, wkcuber_win, wkcuber_mac, wkcuber_docker, docs]
-    if: |
-      always() &&
-      !contains(needs.*.result, 'failure') &&
-      !contains(needs.*.result, 'cancelled')
+    if: always()
```
philippotto, line 321: can this line be removed completely?

philippotto (approved):
great :+1:

.github/workflows/ci.yml
jstriebel: Nope, unfortunately not, since the default is to skip the step if any predecessors failed or were cancelled.
https://docs.github.com/en/actions/learn-github-actions/workflow-syntax-for-github-actions#example-not-requiring-dependent-jobs-to-be-successful

philippotto: ah, good to know. `if: always()` reminds me of `if true`, but I hope that I'll remember this in the future :nerd_face:

PR #432, opened 2021-09-28
calculate_mags_to_downsample: raise on infinite loop

jstriebel:
### Description:
`calculate_mags_to_downsample` can run into an infinite loop when having a scale and max-mag which don't result in a roughly isotropic resolution. This case is detected now and an `RuntimeError` is raised.


rschwanhold (approved):
LGTM

PR #433, opened 2021-09-30
Move Changelog portions to respective releases

fm3:
@rschwanhold @youri-k Could you both please have a look if the mapping from changes to releases is correct now (and fix it here if you notice an error)?


rschwanhold (approved):
LGTM

PR #434, opened 2021-09-30
Assert that numpy arg to vec3int has integer values

fm3:
### Description:
- Previously you could pass a numpy array with float values to vec3int and they would then be converted by int(val). I think it would be better to be strict here and demand the conversion from the user, lest something fails silently.
- Do you agree @jstriebel ?

### Todos:
 - [ ] Updated Changelog


jstriebel (approved):
:+1:

PR #435, opened 2021-10-04
Update scikit image to 0.18

daniel-wer:
### Description:
- Update scikit image to be able to update it in voxelytics. I tested that the update is compatible with voxelytics and that no significant changes are introduced due to the h-maxima fix.

### Issues:
- contributes to https://github.com/scalableminds/voxelytics/issues/2697

### Todos:
 - [x] Updated Changelog


jstriebel (commented):
Nice, thanks :+1: I think you should also update the lock-file in the `docs` repo. We should probably add a script for those things at some point…

jstriebel (approved):
:+1:

PR #437, opened 2021-10-05
Add shallow_copy_dataset method

valentin-pinkau:
### Description:
- This PR adds a new method to the dataset class: shallow_copy_dataset
- This new method makes publishing datasets to webknossos easier
- In addition, the _add_foreign_mag method does now consume a MagView instead of a path.

### Todos:
 - [ ] We might want to introduce a read-only flag to linked layers/mags.
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -545,8 +545,7 @@ class Dataset:
         Create a new dataset at the given path. Link all mags of all existing layers.
         In addition, link all other directories in all layer directories
         to make this method robust against additional files e.g. layer/mappings/agglomerate_view.hdf5.
-        This method becomes useful when exposing a dataset to webknossos
-        by e.g. linking several layers together or adding downsampled mags
+        This method becomes useful when exposing a dataset to webknossos.
         """
         new_dataset = Dataset.create(
             new_dataset_path, scale=self.scale, name=name or self.name
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -14,7 +14,7 @@ import numpy as np
 import wkw
 
 from webknossos.geometry import BoundingBox, Vec3Int
-from webknossos.utils import get_executor_for_args, copy_directory_with_symlinks
+from webknossos.utils import copy_directory_with_symlinks, get_executor_for_args
 
 from .layer import (
     Layer,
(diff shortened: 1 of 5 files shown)


webknossos/webknossos/utils.py
```diff
@@ -103,3 +105,22 @@ def time_since_epoch_in_ms() -> int:
...
+    Links all directories in src_path / dir_name to dst_path / dir_name.
+    """
+    for item in os.scandir(src_path):
+        if item.name not in ignore:
+            symlink_path = dst_path / item.name
+            if make_relative:
+                rel_or_abspath = os.path.relpath(Path(item.path), symlink_path.parent)
+            else:
+                rel_or_abspath = os.path.abspath(item.path)
+            symlink_path.symlink_to(rel_or_abspath)
```
jstriebel, line 126: Why not simply use `iterdir`?
```suggestion
    for item in src_path.iterdir():
        if item.name not in ignore:
            symlink_path = dst_path / item.name
            if make_relative:
                rel_or_abspath = os.path.relpath(item, symlink_path.parent)
            else:
                rel_or_abspath = os.path.abspath(item)
            symlink_path.symlink_to(rel_or_abspath)
```

webknossos/webknossos/dataset/layer.py
```diff
@@ -372,50 +371,41 @@ def delete_mag(self, mag: Union[int, str, list, tuple, np.ndarray, Mag]) -> None
...
-            )
+        is_foreign_mag_compressed = (
+            foreign_mag_view.header.block_type != wkw.Header.BLOCK_TYPE_RAW
+        )
+        added_mag_view = self.add_mag(
+            foreign_mag_view.mag,
+            foreign_mag_view.header.block_len,
+            foreign_mag_view.header.file_len,
+            is_foreign_mag_compressed,
+        )
```
jstriebel, line 384: This might be problematic, since there are multiple compression possibilities and the mag object caches the header.

webknossos/webknossos/dataset/layer.py
```diff
@@ -372,50 +371,41 @@ def delete_mag(self, mag: Union[int, str, list, tuple, np.ndarray, Mag]) -> None
...
-        self.bounding_box = self.bounding_box.extended_by(original_layer.bounding_box)
-        self._properties.wkw_resolutions += [mag_properties]
-        self._setup_mag(mag)
-        self.dataset._export_as_json()
-        return self.mags[mag]
+        return added_mag_view
 
     def add_symlink_mag(
-        self, foreign_mag_path: Union[str, Path], make_relative: bool = False
+        self, foreign_mag_view: MagView, make_relative: bool = False
```
jstriebel, line 408: We might simply support  `Union[os.PathLike, str, MagView]`

webknossos/webknossos/dataset/layer.py
jstriebel: PS: I've added `add_mag_for_existing_files` in the REST API work, this might help here. I'll open a separate PR for it, so you might already use it.

jstriebel (commented):
Nice, thanks! See my few comments:

webknossos/webknossos/dataset/layer.py
valentin-pinkau: Perfect! Thanks!

webknossos/webknossos/dataset/dataset.py
```diff
@@ -535,6 +556,36 @@ def copy_dataset(
...
+            for mag_view in layer.mags.values():
+                new_layer.add_symlink_mag(mag_view, make_relative)
+
+            # copy all other directories with a dir scan
+            copy_directory_with_symlinks(
+                layer.path,
+                new_layer.path,
+                ignore=[str(mag) for mag in layer.mags] + [PROPERTIES_FILE_NAME],
+                make_relative=make_relative,
+            )
```
jstriebel, line 585: Alternatively you could also link all directories first and use `add_layer_for_existing_files` afterwards, or does this do something different? But no strong opinion.

jstriebel (approved):
Nice, mostly LGTM :+1:

> What do you think of having a read-only flag in symlinked layers/mags?

I think this only makes sense if it can also be persisted in the datasource-properties.json. I'd wait with such change to see if we really want those, we can easily add them later.

webknossos/webknossos/dataset/dataset.py
valentin-pinkau: Currently, the links are on the mag level so adding another mag (through downsampling) would not invalidate the source dataset. I could still link on mag level and use the `add_layer_for_existing_files` afterwards, but I like explicit `add_layer_like` and `add_symlink_mag` more.

jstriebel: Exactly, I thought also about linking on the mag level, but feel free to leave as-is.

PR #438, opened 2021-10-06
Allow full Vec3IntLike as input for get_buffered_slice_writer

### LINKED ISSUE #436: get_buffered_slice_writer should take Vec3IntLike for offset
currently the user has to construct Vec3Int

fm3:
### Description:
- allowing Vec3IntLike in public functions and then converting to Vec3Int is more user-friendly, and already implemented in many libs functions
- Parameters with default value `None` should be marked as `Optional[…]` (do you agree?)
- I moved up the 0,0,0-default where it seemed appropriate (again, do you have an opinion on this?)

### Issues:
- fixes #436

### DIFF
--- a/webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
+++ b/webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
@@ -8,7 +8,7 @@ from typing import TYPE_CHECKING, Generator, List, Optional, Tuple, Type, cast
 import numpy as np
 import psutil
 
-from webknossos.geometry import Vec3Int
+from webknossos.geometry import Vec3Int, Vec3IntLike
 
 if TYPE_CHECKING:
     from webknossos.dataset import View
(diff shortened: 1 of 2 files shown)


jstriebel:
> Parameters with default value None should be marked as Optional[…] (do you agree?)

@fm3 You can supply `--no-implicit-optional` to mypy, which stops to implicitly make this Optional, which I'd also prefer. Feel free to try this out or to add an issue. See also the last note in this section: https://mypy.readthedocs.io/en/stable/kinds_of_types.html#optional-types-and-the-none-type
> Optional[...] does not mean a function argument with a default value. However, if the default value of an argument is None, you can use an optional type for the argument, but it’s not enforced by default. You can use the --no-implicit-optional command-line option to stop treating arguments with a None default value as having an implicit Optional[...] type. It’s possible that this will become the default behavior in the future.

and on SO: https://stackoverflow.com/questions/62732402/can-i-omit-optional-if-i-set-default-to-none#answer-62732458

fm3:
Thanks for your input! I created https://github.com/scalableminds/webknossos-libs/issues/443 as a follow-up issue for this :)

PR #448, opened 2021-10-07
Do not expose open() and close() to user in dataset api

### LINKED ISSUE #445: Dataset API: Do not expose open() / close() to user
Currently the user can worry about when a dataset's view should be opened or closed. Instead we should simply open it when initializing the view, and closing on on destruction. Opening it for every read/write might have performance implications (many small writes become twice as slow when always re-opening the wkw-ds), but opening it upon construction seems fine.

rschwanhold:
### Description:
- The `wkw.Dataset` is now opened lazy instead of opening it manually or for each read/write. This still supports that `View`s can easily be passed around (as long as it has not been opened) without any danger of destroying the handle. The advantage is that the user does not have to worry about opening or closing.
- The other alternative would be to open and close the dataset every time on `read` and `write`. However, this approach is slower (in a toy example, see below). Opening and closing the handle each time resulted in +30% runtime (this is only relevant in this artificial example where the number of file accesses is the dominating factor)

```
# Test for the old implementation
def test_runtime(tmp_path: Path):
    mag_view = Dataset.create(tmp_path/"ds", scale=(1, 1, 1)).add_layer("color", "color").add_mag(1)
    mag_view.write(data=np.random.random((5, 5, 10000)).astype(np.uint8))

    import time

    start = time.time()
    #mag_view.open()
    for i in range(1000000):
        mag_view.read(offset=(0, 0, i%10000), size=(5, 5, 1))
    end = time.time()
    #mag_view.close()
    print(end - start)
```

### Issues:
- fixes #445 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/tests/test_dataset.py
+++ b/webknossos/tests/test_dataset.py
@@ -9,7 +9,6 @@ from typing import Generator, Tuple, cast
 
 import numpy as np
 import pytest
-from wkw.wkw import WKWException
 
 from webknossos.dataset import (
     Dataset,


webknossos/webknossos/dataset/view.py
```diff
@@ -603,7 +558,7 @@ def __exit__(
         _value: Optional[BaseException],
         _tb: Optional[TracebackType],
     ) -> None:
-        self.close()
+        pass
```
jstriebel, line 561: Please add a warnin on `__enter__`, that the view does not need to be used as a context manager anymore.

webknossos/webknossos/dataset/view.py
```diff
@@ -616,6 +571,25 @@ def _mag_view_bounding_box_at_creation(self) -> BoundingBox:
...
+    def _close(self) -> None:
+        if self._dataset is None:
+            raise Exception("Cannot close View: the view is not opened")
+        else:
+            self._dataset.close()
+            self._dataset = None
+
+    def _open_if_necessary(self) -> None:
+        if self._dataset is None:
+            self._open()
```
jstriebel, line 591: I'd recommend to use a property for `_dataset` and rename it to `_wkw_dataset`, so it automatically gets opened when needed, and closing it in the destructor `__del__` (which can also be invoked explicitly using `del view` or just the dataset-cache using `del view._wkw_dataset`):

```suggestion
    @property
    def _wkw_dataset(self) -> None:
        if self._cached_wkw_dataset is None:
            _cached_wkw_dataset = Dataset.open(
                str(self._path)
            )  # No need to pass the header to the wkw.Dataset
        return self._cached_wkw_dataset
        
    @_wkw_dataset.deleter
    def _wkw_dataset(self) -> None:
        if self._cached_wkw_dataset is not None:
            self._cached_wkw_dataset.close()
            self._cached_wkw_dataset = None

    def __del__(self) -> None:
        del self._cached_wkw_dataset
```

PR #452, opened 2021-10-13
add webknossos-client with examples

### LINKED ISSUE #381: [wk-lib] add webknossos-REST client

jstriebel:
### Description:
* Adds a first version of a user-facing REST client for webknossos with support for
  * annotation download
  * dataset up- and download
* Also adds examples using the client. The examples are run during the tests, with cached network requests.

### Issues:
- fixes #381

## Followups
- Investigate why and how the dataset download should be chunked
- Add more unit-tests for the client
- Ensure the context-vars are propagated via executors
- Use a single httpx client for all requests and parallelize them where possible
- Reset invalid tokens (e.g. when datastore-tokens time out or user-tokens are revoked)
- Smooth out what can be specified in the download APIs: urls, ids, other orgas, additional sharing tokens, …
- [x] Speed up the `learned_segmenter` example in the tests

### Todos:
- [x] Investigate if the dataset bounding-box is actually a bugfix and why the tests are failing
- [x] Updated Changelog
- [x] Add issues for followups

### DIFF
--- a/wkcuber/wkcuber/metadata.py
+++ b/wkcuber/wkcuber/metadata.py
@@ -1,6 +1,4 @@
 import json
-import os
-import re
 from webknossos.dataset._utils.infer_bounding_box_existing_files import (
     parse_cube_file_name,
 )


jstriebel:
> I reviewed everything except for the resumable module. Was there something you based this on? Not sure how I should review this ideally.

Good point, should've mentioned this in the description. This is based on [resumable.py](https://github.com/acroz/resumable.py), but I changed it quite a bit (using httpx instead of requests, attrs for the config instead of a plain dict, added typing, added support for some features such as customized ids etc…). I'd recommend to also go through that code, I think it's fairly simple. We could think about re-publishing this at some point as a stand-alone package, but atm I don't think that's helpful, since it's still limited in comparison to resumable.js.

jstriebel:
@philippotto Thanks a ton for the great & detailed review :bee: :+1: I think I addressed all points by now (some are deferred to issues). Could you please double-check if this is the case and re-review my new changes? Thanks :pray:

PR #470, opened 2021-10-28
Add User client

jstriebel:
### Description:
Adds a `User` class to the client that can be used to get metainformation of users or their logged time. The currently logged in user can be accessed, as well as all managed users. Also adds an example and docs.

For the docs, please check https://static.webknossos.org/docs/user_client/webknossos-py/examples/user_times.html

### Todos:
 - [x] Updated Changelog


philippotto (approved):
Very cool :) I have looked very carefully, but don't have any feedback/nitpicks :laughing:

Only tangentially related: What do you think about having a short write-up about how the REST API can be extended? If I were to extend it with a new concept (such as the Users), I think, I wouldn't really know where/how to start :thinking:

jstriebel:
Thanks!

Good point, I added an issue about adding developer documentation: #473

PR #480, opened 2021-11-04
Improve Performance of Cubing

rschwanhold:
### Description:
- I have integrated the performance optimization suggested in #387

### Results:
- I have used layer `c0` of `humerus:/srv/datasets/yorick/14-36-08_AC307_Head_PCW7-5_4X_06Z_5-MHC_6-chAT_7-b3tub_4u_lr_mosaic` on my local machine to compare the performance of cubing on the master branch with this PR.
  - I skipped the first 1400 slcies (from the 1508 z-slices) due to memory limitations
- Results of the master branch:
```
Command being timed: "python -m wkcuber.cubing --batch_size 8 --layer_name color --scale 1,1,1 --skip_first_z_slices 1400 --pad /home/r/Documents/scalableminds/large_cubing_ds/c0 /home/r/Documents/scalableminds/large_cubing_ds_output/with_performance_skip_1400"
        User time (seconds): 66.10
        System time (seconds): 54.24
        Percent of CPU this job got: 80%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:30.24
        Maximum resident set size (kbytes): 1149208
        Exit status: 0
```
- Results of this branch:
```
Command being timed: "python -m wkcuber.cubing --batch_size 8 --layer_name color --scale 1,1,1 --skip_first_z_slices 1400 --pad /home/r/Documents/scalableminds/large_cubing_ds/c0 /home/r/Documents/scalableminds/large_cubing_ds_output/with_performance_skip_1400"
        User time (seconds): 50.92
        System time (seconds): 48.39
        Percent of CPU this job got: 70%
        Elapsed (wall clock) time (h:mm:ss or m:ss): 2:20.31
        Average total size (kbytes): 0
        Maximum resident set size (kbytes): 591260
        Exit status: 0
```
- The changes reduce the total run time slightly from `2:30.24` to `2:20:31` and reduced the maximum memory from `1,14` to `0.59` (which is almost 50%)

### Todos:
 - [x] Test `tile_cubing.py`
 - [x] Updated Changelog


wkcuber/wkcuber/tile_cubing.py
```diff
@@ -292,39 +292,81 @@ def tile_cubing(
...
-            file_count, tile_size[0], tile_size[1]
+            file_count, num_x, num_y
         )
     )
     if args is None or not hasattr(args, "dtype") or args.dtype is None:
         dtype = image_reader.read_dtype(arbitrary_file)
+    else:
+        dtype = args.dtype
+
+    target_ds = Dataset.get_or_create(target_path, scale=(1, 1, 1))
```
rschwanhold, line 308: I have hardcoded the scale. Maybe a parameter would be nicer.

wkcuber/wkcuber/tile_cubing.py
```diff
@@ -292,39 +292,81 @@ def tile_cubing(
...
+                Vec3Int(0, 0, min_dimensions["z"]),
+                Vec3Int(num_x, num_y, num_z),
+            )
+    else:
+        target_layer.bounding_box = target_layer.bounding_box.extended_by(
+            BoundingBox(
+                Vec3Int(0, 0, min_dimensions["z"]),
+                Vec3Int(num_x, num_y, num_z),
+            )
+        )
```
rschwanhold, line 338: I have set the z-offset to `min_dimensions["z"]`, but I am not sure if this is correct. The test dataset `/testdata/temca2` starts with `z=1`. Currently this would lead to a dataset with a bounding box that starts at 1. Is this intended or does the index for tiles always start at "1"?

rschwanhold (commented):
I have tested the performance (see description). I have also checked the equality of the results between this PR and the master after cubing `testdata/tiff`. Both of these datasets do not require padding, but there is a test case in the CI that runs cubing on tiffs that need padding.

I have only tested `tile_cubing` on `testdata/temca2`.

I have added comments regarding the two points where I am unsure.

wkcuber/wkcuber/tile_cubing.py
```diff
@@ -185,89 +182,92 @@ def find_file_with_dimensions(
...
+                            image = np.zeros(
+                                tile_size + (1,),
+                                dtype=target_view.header.voxel_type,
                             )
+                        buffer[:, :, :, z - z_batch[0]] = image.transpose((2, 0, 1, 3))[
+                            :, :, :, 0
+                        ]
+
+                    if np.any(buffer != 0):
+                        target_view.write(data=buffer)
```
philippotto, line 255: I think, this line is missing the offset which is constituted by `[x * tile_size[0], y * tile_size[1]`. Otherwise, all tiles would be written to `[0, 0]`.

philippotto (changes requested):
Great improvement :+1: I think, there is one bug, but other than that this looks very good to me. Since you probably don't have much time, anymore, for this PR, I will try to overtake this next week or so.

wkcuber/wkcuber/tile_cubing.py
philippotto: I think, the default `cubing` strategy writes moves the data so that it starts at `z=0`. However, I think, we should try to keep the original indexing by default (unless specified differently). This is a broader issue, though.
For this PR, reusing `z` seems okay to me. Would be a bummer, though, if this causes tests to fail :smile:

wkcuber/wkcuber/tile_cubing.py
```diff
@@ -292,39 +292,81 @@ def tile_cubing(
...
+
+    target_ds = Dataset.get_or_create(target_path, scale=(1, 1, 1))
+    is_segmentation_layer = layer_name == "segmentation"
+    if is_segmentation_layer:
+        target_layer = target_ds.get_or_add_layer(
+            layer_name,
+            LayerCategories.SEGMENTATION_TYPE,
+            dtype_per_channel=dtype,
+            num_channels=num_channels,
+            largest_segment_id=0,
```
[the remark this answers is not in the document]
rschwanhold: Updating the `largest_segment_id` in the `write` function does not work because `View.write()` does not know the dataset. However, for this particular case, we can simply compute the `largest_segment_id` for each buffer in `tile_cubing_job` and then update the id once all jobs have finished cubing. (We do the same thing in `cubing` since recently)

wkcuber/wkcuber/tile_cubing.py
rschwanhold: The test dataset `temca2` starts at `z=1` and ends at `z=40`.
The new code generates a datasource-config with bounding box `topLeft=(143360, 61440, 1), size=(6144, 2024, 40)`. I have run the test with the code from the master and this PR. They both have the same data in this boundingbox (which contains indeed some data). Therefore, I conclude that the code works.
As far as I can see it, the previous code also wrote the data to the "original" index.

philippotto: Great, sounds good then :+1:

philippotto (approved):
Looks awesome to me! :ship:

PR #484, opened 2021-11-22
Adapt to new wk dataset upload api

fm3:
### Description:
- The webknossos dataset upload API changes in https://github.com/scalableminds/webknossos/pull/5831
- This PR adapts the client to use that new API
- They should be released together. In the future, we should introduce proper API versioning, but for now this should do the trick.

### Todos:
 - [x] Updated Changelog


fm3:
@jstriebel Do you know if there is a way to run the tests against the new wk version?

jstriebel:
Sure: You need to update the `DOCKER_TAG` here (if you have webknossos running with the desired version anyways that works too, but please still update it so future runs work out of the box): https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/test.sh#L13
Then you can run `./test.sh --refresh-snapshots`

fm3:
@jstriebel thank you for your help :) We’ve come quite far already here

I adapted the dockerfile so that the new branch is executed (thanks for the support with the sampleOrganization overwrite!)

When executing refresh-snapshots I get a few assertion errors still, even after adapting for the new external-datastore setup of wk.org – For example, I did not understand why the dataset download is broken now, does this work only with internal datastores? In this case, nobody can download any dataset from wk.org right now.

I also got a surprisingly large amount of diff in the snapshots, and felt kind of lost about which parts of that I should commit.

I would be very grateful if next week you could have a look at this PR and what is missing to get the CI green again, since this is blocking the wk pr https://github.com/scalableminds/webknossos/pull/5831 and also, more such api changes are going to follow soon.

webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -35,6 +35,23 @@ def upload_dataset(dataset: Dataset) -> str:
...
+                        "organization": context.organization,
+                        "name": dataset.name,
+                        "totalFileCount": 1,
+                        "initialTeams": [],
+                    },
+                    timeout=60,
+                ).raise_for_status()
+                break
+            except httpx.HTTPStatusError as e:
+                http_error = e
```
jstriebel, line 54: This won't fail after 5 failed retries, this else clause helps:
```suggestion
        for _ in range(5):
            try:
                httpx.post(
                    f"{datastore_url}/data/datasets/reserveUpload?token={datastore_token}",
                    params={"token": datastore_token},
                    json={
                        "uploadId": upload_id,
                        "organization": context.organization,
                        "name": dataset.name,
                        "totalFileCount": 1,
                        "initialTeams": [],
                    },
                    timeout=60,
                ).raise_for_status()
                break
            except httpx.HTTPStatusError as e:
                http_error = e
        else:
            raise http_error
```

jstriebel (approved):
Nice, mostly LGTM :+1: Just see my two comments:

webknossos/webknossos/client/_defaults.py
```diff
@@ -1 +1,2 @@
 DEFAULT_WEBKNOSSOS_URL = "https://webknossos.org"
+DEFAULT_WEBKNOSSOS_DATASTORE_URL = "https://data-humerus.webknossos.org"
```
jstriebel, line 2: PS: Is this used for something?

webknossos/webknossos/client/_upload_dataset.py
fm3: Nice, learned something today :D

webknossos/webknossos/client/_defaults.py
fm3: No, just a merge artifact ( I had originally started out building something like #497 myself ). Good catch!

webknossos/test.sh
```diff
@@ -6,11 +6,11 @@ if [ $# -eq 1 ] && [ "$1" = "--refresh-snapshots" ]; then
         WK_DOCKER_DIR="tests/.webknossos-server"
         if [ ! -d "$WK_DOCKER_DIR" ]; then
             git clone git@github.com:scalableminds/webknossos.git $WK_DOCKER_DIR --depth 1
-            cp tests/wk-docker-compose.override.yml $WK_DOCKER_DIR/docker-compose.override.yml
         fi
         pushd $WK_DOCKER_DIR > /dev/null
+        sed -i -e 's/webKnossos.sampleOrganization.enabled=false/webKnossos.sampleOrganization.enabled=true/g' docker-compose.yml
         mkdir -p binaryData
-        export DOCKER_TAG=21.11.0
+        export DOCKER_TAG=master__16085__TODO
```
fm3, line 13: I’ll add the new master tag once https://github.com/scalableminds/webknossos/pull/5831 is merged

PR #491, opened 2021-11-29
Integrate cluster_tools

### LINKED ISSUE #446: Add clustertools to wk-libs repo
clustertools is tightly integrated with the webknossos and wkcuber packages, we should move it into this repo as well.

philippotto:
### Description:
- integrated cluster_tools into mono-repo structure
- migrated the history of the old cluster_tools repo into this PR
- converted the circleci config to GH actions
- converted clustertools to use poetry
- use pre-built docker image for slurm environment

### Issues:
- fixes #446

### Todos:
 - [ ] Updated Changelog


wkcuber/pyproject.toml
```diff
@@ -8,13 +8,13 @@ license = "AGPL-3.0"
 include = ["wkcuber/version.py"]
 
 [tool.poetry.dependencies]
-python = "^3.7"
+python = "^3.7,>=3.7.1"
```
philippotto, line 11: @jstriebel I had to make this change, because poetry complained that this requirement didn't match `webknossos/pyproject.toml` (which also uses `python = "^3.7,>=3.7.1"`). After I adapted it, I ran `poetry lock` which might have been a mistake since `wkcuber/poetry.lock` has now quite a lot of changes? Not sure whether I could have circumvented this and if yes if it's worth it to go back and remove these changes again?

What's your take on that? If we stick to the updates (which we would probably use at some point anyway), I need to fix some new linting errors. Then, the CI should turn green. Please let me know which route you prefer :)

jstriebel: Ah, ok. I guess you could revert the `poetry.lock` changes and run `poetry upgrade webknossos`, which should be enough I think. Maybe we should have a separate script which cross-updates all internal dependencies we have. This could also run in CI with an assertion that nothing changes. But doesn't need to be part of this PR for sure.

cluster_tools/dockered-slurm/docker-compose.yml
```diff
@@ -0,0 +1,86 @@
+version: "2.2"
```
jstriebel, line 1: I guess most files in `dockered-slurm` can be removed, right? If so, remove them and update the usages everywhere.

.github/workflows/ci.yml
```diff
@@ -18,6 +19,58 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Build/pull dockered-slurm image
+        run: |
+          cd ./dockered-slurm
+
+          echo build
+          # Build image
+          # docker build -t slurm-docker-cluster:latest .
+          # docker pull scalableminds/slurm-docker-cluster:slurm__1531478467
+
```
jstriebel, line 42: Probably outdated and can be removed?
```suggestion
```

cluster_tools/README.md
```diff
@@ -0,0 +1,39 @@
...
+  with cluster_tools.get_executor(strategy) as executor:
+    result = list(executor.map(square, [2, 3, 4]))
+    assert result == [4, 9, 16]
+```
+
+## Dev Setup
+
+```
+cd dockered-slurm
+docker build -t slurm-docker-cluster:latest .
```
jstriebel, line 27:
```suggestion
```

.github/workflows/ci.yml
```diff
@@ -303,22 +356,22 @@ jobs:
...
+        run: pip3 install -r requirements.txt
+      - name: Publish python packages
+        env:
+          PYPI_USERNAME: ${{ secrets.PYPI_USERNAME }}
+          PYPI_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
+        run: ./publish.sh
 
   complete:
-    needs: [webknossos_linux, wkcuber_linux, wkcuber_win, wkcuber_mac, wkcuber_docker, docs]
+    needs: [webknossos_linux, wkcuber_linux, wkcuber_win, wkcuber_mac, wkcuber_docker, docs, cluster_tools]
```
jstriebel, line 374: Should we also run the webknossos or wkcuber package CIs if we change cluster_tools? We could run them in parallel without waiting for the status of cluster_tools.

cluster_tools/README.md
```diff
@@ -0,0 +1,39 @@
...
+docker-compose up -d
+docker exec -it slurmctld bash
+docker exec -it c1 bash
+```
+
+Tests can be executed with `python3 -m pytest -s test.py` after entering the container.
+Linting can be ran with `python3 -m pylint cluster_tools/`.
+Code formatting (black) can be ran with `python3 -m black cluster_tools *.py`.
```
jstriebel, line 35:
```suggestion
Tests can be executed with `poetry run pytest -s test.py` after entering the container.
Linting can be run with `./lint.sh`.
Code formatting (black) can be run with `./format.sh`.
```

cluster_tools/dockered-slurm/how-to.md
```diff
@@ -0,0 +1,36 @@
+# Install segtools
```
jstriebel, line 1: What is this file used for? Some parts seem outdated, and maybe it should be merged with the README?

jstriebel:
Nice stuff, thanks a ton for the migration! :muscle: I left a couple of comments, but mostly LGTM :+1:

cluster_tools/dockered-slurm/how-to.md
philippotto: you are right. i removed it as the readme should contain all the relevant parts.

cluster_tools/dockered-slurm/docker-compose.yml
philippotto: I cleaned them up now, but the docker-compose.yml has to stay since it's used to spawn the slurm cluster.

wkcuber/pyproject.toml
philippotto: Since there was a merge conflict in the poetry.lock file, anyway, I restored the lock files from master and then only ran poetry upgrade {webknossos,wkcuber}`. The line change count is now far smaller :tada:

.github/workflows/ci.yml
philippotto: Yes, good point. I would assume it makes sense to always run the CI for modules when their dependencies change. I.e., if clustertools change webknossos and wkcuber need to run. If webknossos changes, wkcuber needs to run. Do you agree?

While we are at it:
image

Do you know why `wkcuber_{mac,win}` don't consume `changes`? Is this intended? I would have thought they should (together with `webknossos_linux`) consume `changes` and then only run if some of `{wkcuber,webknossos,clustertools}` change.

jstriebel: I guess the idea was that wkcuber has to run in any case since it depends on all other packages. But we can still enumerate all, probably nicer to check. Also, currently it depends on webknossos, which is not necessary and just prolongues the run-times. The whole pipeline will fail if any of them fails anyways.

wkcuber/Dockerfile
```diff
@@ -18,6 +18,12 @@ COPY webknossos/poetry.lock /webknossos/
 COPY webknossos/pyproject.toml /webknossos/
 COPY webknossos/README.md /webknossos/
 
+RUN mkdir /cluster_tools
+COPY cluster_tools/cluster_tools /cluster_tools/cluster_tools
+COPY cluster_tools/poetry.lock /cluster_tools/
+COPY cluster_tools/pyproject.toml /cluster_tools/
+COPY cluster_tools/README.md /cluster_tools/
+
```
jstriebel, line 26: Were they linked instead before? I guess this was simply missing, or did something else make this necessary?

philippotto: Earlier versions simply required them as a package via poetry. Now that it has changed to `cluster_tools = { path = "../cluster_tools/", develop = true }`, it has to be copied into the docker image. This is analogous to how webknossos is added to the docker image.

jstriebel (approved):
Thanks a lot :+1:

PR #492, opened 2021-11-29
Resolve path when symlinking layer and make_relative is False

daniel-wer:
### Description:
- abspath resulted in paths pointing into specific user directories which may not always be accessible. Instead, all symlinks should be resolved so that the resulting symlink points to the actual location on disk.
- I'll share the test workflow results privately

### Todos:
 - [x] Updated Changelog


georgwiese (approved):
❤️

PR #494, opened 2021-11-30
Fix wkcuber --version

daniel-wer:
### Description:
- While upgrading to the newest webknossos-libs version, I stumbled upon this error:
```
  File "/home/runner/.venv/lib/python3.8/site-packages/wkcuber/__main__.py", line 3, in <module>
    from versioning import get_available_version
ModuleNotFoundError: No module named 'versioning'
```
- Looking at the original PR introducing the --version flag, I'm not sure this has ever worked (https://github.com/scalableminds/webknossos-libs/pull/471) or I'm seriously misunderstanding how this was supposed to work. The problems I'm seeing and that I fix in this PR:
  - The version.py file was created after publishing, not ending up in the published package
  - ~~It was created in the wkcuber root (so not ending up in the published package), but imported as `from wkcuber.version import __version__`~~
  - The versioning.py file was created in the wkcuber root instead of in wkcuber/wkcuber, therefore also not ending up in the published package

### Issues:
- fixes wkcuber package

### Todos:
 - [ ] Updated Changelog


publish.sh
```diff
@@ -15,13 +15,13 @@ for PKG in */pyproject.toml; do
     cp pyproject.toml pyproject.toml.bak
     PKG_VERSION="$(dunamai from git)"
 
+    echo "__version__ = '$PKG_VERSION'" > ./"$PKG"/version.py
```
normanrz, line 18: Needs to be `./"$PKG"/version.py`, because we already pushd-ed into `$PKG`

normanrz (approved):

publish.sh
daniel-wer: Thank you, great catch!

PR #497, opened 2021-12-01
Allow downloads from remote datastore, update & fix tests, longer timeouts

### LINKED ISSUE #487: Allow to download dataset from remote-datastore
Currently, downloading a dataset only works if the datastore is available under the same url as webknossos. Downloading from remote datastores should be possible, too.

jstriebel:
### Description:
- Downloads from remote datastores are possible now, but auth via a token is still missing, this is blocked by scalableminds/webknossos#5875 (I'll do this afterwards in a separate PR)
- The snapshot tests are refreshed and fixed
- Timeouts are much longer (default was 5s previously), which was necessary especially for uploads
- All examples use the `if __name__ == "__main__"` pattern and can still be executed in the tests

### Issues:
- fixes #487

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/tests/test_examples.py
+++ b/webknossos/tests/test_examples.py
@@ -1,6 +1,8 @@
+import inspect
 import os
 from contextlib import contextmanager
 from tempfile import TemporaryDirectory
+from types import ModuleType
 from typing import Any, Iterator
 
 import numpy as np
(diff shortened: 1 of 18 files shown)


webknossos/tests/test_examples.py
```diff
@@ -21,38 +23,45 @@ def tmp_cwd() -> Iterator[None]:
             os.chdir(prev_cwd)
 
 
+def exec_module_as_main(module: ModuleType) -> None:
+    source = inspect.getsource(module)
+    new_source = source.replace('\nif __name__ == "__main__":\n', "\nif True:\n")
+    exec(new_source, module.__dict__)  # pylint: disable=exec-used
```
philippotto, line 29: alternatively, you could define a `def main():` in the examples which can be imported then. I would find this more elegant, but I can understand if you want to stick to the current approach for the sake of keeping the examples more "script-like".

philippotto (approved):
Looks good to me :+1: Only left one suggestion.

webknossos/tests/test_examples.py
jstriebel: I actually had that in place first, but got problem with testing: To assert the variables one needs them to be returned or in the global scope, which I both found weird in the example. However, I could do a similar trick with changing the code of the module to do any of those options, I'll give it another try.

philippotto: I think, returning something in the examples would be fine. The example could even use it like this:

```
def main():
  ...
  return some_object

if main...:
  some_object = main()
  print(some_object)
```

jstriebel: I changed this now to use the `main` function, but without adding returns in the examples (I think they might just confuse users). I used the same magic to make all needed variables global in the tests. Adding a return statement works, but the replace-code is much more fragile, since the end of the function needs to be detected.

Issue #274, opened 2021-02-12
Support .raw and .vol files


normanrz:
These formats house a flat stream of voxels.

raw files have their dimensions in the file name `test.123x234x345.raw`.

vol files have an accompanying `.vol.info` file attached. Example:
```
! PyHST_SLAVE VOLUME INFO FILE
NUM_X =  2800
NUM_Y =  2800
NUM_Z =  2048
voxelSize =  5.000830e-02
BYTEORDER = LOWBYTEFIRST
ValMin =  -3.66486481e+02
ValMax =  1.36156952e+02
s1 =  -2.40931052e+02
s2 =  7.82870530e+01
S1 =  -1.14686499e+02
S2 =  4.82676492e+01
```

PR #498, opened 2021-12-02
Adding Raw data file conversion tool

t20100:
### Description:
We would like to deploy webknossos at our institute (ESRF) and need to convert datasets from raw files (with dataset description in a .info or .xml file).

So, this PR adds the `wkcuber.convert_raw` module to convert raw datasets to wkw format.
I added the options that I needed, but it sounds to me generic enough.
Yet, some features that are not implemented in this PR: Support for segmentation datasets, multi-channels support, provide a byte offset in the file, ~~no downsampling processing integrated~~.
 
I also initially added support for `.info` files to find the data shape and guessing data type, but it is probably best to do that outside of the `wkcuber` tool since this is by no means standard.
Anyway if you are interested, the code is in this [branch](https://github.com/t20100/webknossos-libs/tree/raw-converter-dev)

### Issues:
- related to #274

### Todos:
 - [x] Updated Changelog
 - [x] Updated README
 - [x] More testing


t20100:
Added downsampling feature to `python -m wkucber.convert_raw` command.
Also directly import from `webknossos` rather than through `wkcuber.api`... please tell me if it is not the best way.

t20100:
Indeed!

I added the distribution options (only tested with default `multiprocessing` though) for both conversion and downsampling.
Got a 6x speed-up for the conversion on 12 cores machine, so it's good.

wkcuber/wkcuber/convert_raw.py
```diff
@@ -0,0 +1,239 @@
...
+    add_verbose_flag,
+    get_executor_args,
+    setup_logging,
+)
+
+
+logger = logging.getLogger(__name__)
+
+
+CHUNK_SIZE = 1024, 1024, 1024
```
t20100, line 26: Is there a place I can get the size of the block files from instead of having it hard-coded here?

normanrz: I think it would make sense to the chunk size dependent on the output wkw file sizes (default: 1024^3). You can compute that from `(block_len * file_len,) * 3`. We don't expose a constant for `block_len` atm, which we probably should, but it is always 32.

wkcuber/wkcuber/convert_raw.py
```diff
@@ -0,0 +1,239 @@
...
+    bounding_box: BoundingBox,
+    source_raw_path: Path,
+    target_mag_view: MagView,
+    input_dtype: str,
+    shape: Tuple[int, int, int],
+    order: str,
+    flip_axes: Optional[Union[int, Tuple[int, ...]]],
+):
+    logging.info(f"Conversion of {bounding_box.topleft}")
+    cube_data = np.memmap(
```
normanrz, line 141: I learned from experiments that it is beneficial to copy the data in-memory, when the input order differs from the output order (wkw uses F). Otherwise, numpy may perform a lot of tiny reads, which is inefficient for IO, and memmapping doesn't always mitigate this. Not sure if copying also brings perf benefits in case the orders are the same.

normanrz:
Excellent! Thanks again for providing this PR.
I just tested it on a slurm cluster and it seems to work and produces the expected results :+1:

wkcuber/wkcuber/convert_raw.py
t20100: I just update this, thanks!

wkcuber/wkcuber/convert_raw.py
t20100: I just added a copy of the array.

wkcuber/wkcuber/convert_raw.py
```diff
@@ -23,7 +23,7 @@
 logger = logging.getLogger(__name__)
 
 
-CHUNK_SIZE = 1024, 1024, 1024
+BLOCK_SIZE = 32
```
jstriebel, line 26: @normanrz There's also `DEFAULT_WKW_FILE_LEN` in `webknossos.dataset.defaults`, as well as `BLOCK_LEN`
and `DEFAULT_WKW_VOXELS_PER_BLOCK = 32` in `wkcuber.utils`.

normanrz (approved):
Thanks for making these changes!

wkcuber/wkcuber/convert_raw.py
normanrz: ah, ok. then `DEFAULT_WKW_VOXELS_PER_BLOCK` should probably be used here instead.

t20100: Perfect! I could have grabbed this from `convert_nifti.py`.

t20100:
I tested a bit more locally and added a `pytest` test around `order` and `flip_axes`.

normanrz (approved):

normanrz:
I fixed some typing issues and CI tests yesterday (tests on Windows need to be run with wkw-compression, because disk storage in the CI is limited). For some reason, the tests still don't run for this branch, but I made a copy of the branch in our repo where everything works fine. So, I'll go ahead and merge this one.

Thanks again for this excellent PR!

t20100:
Thanks!

Probably due to [using encrypted secrets in a workflow](https://docs.github.com/en/actions/security-guides/encrypted-secrets#using-encrypted-secrets-in-a-workflow):

> Note: With the exception of GITHUB_TOKEN, secrets are not passed to the runner when a workflow is triggered from a forked repository.

in the CI:

https://github.com/scalableminds/webknossos-libs/blob/6c9520fa004849e8e13259abfa04289b90cccda6/.github/workflows/ci.yml#L210-L211

PR #502, opened 2021-12-06
Update _upload_dataset.py

normanrz:
Finishing an upload can take very long, if the server needs to do stuff like unzipping large files.


jstriebel (approved):
:+1:

PR #503, opened 2021-12-06
GH Action for creating releases

### LINKED ISSUE #499: Automatically update changelogs during release
It would be very helpful if the changelogs were automatically updated when releasing a new webknossos-libs version. Similarly to https://github.com/scalableminds/voxelytics/pull/1899 except not automatically every first of the month but instead when deciding to create a new release.

Additionally, it would be beneficial to separate the changelogs into two files, one containing the released and one the unreleased changes (see webknossos for reference). This would avoid that changelog entries are assigned to the wrong release during merges as happened [here](https://github.com/scalableminds/webknossos-libs/commit/bcf4dd64c55ab04946ef8f9534632918aad2e2f8#r61060362). If it's easier to do the above with this change, it could be done as part of this issue, otherwise it could be a followup :)

normanrz:
### Description:
Adds a github action for creating releases. The action is triggered from the UI and takes a version number to build (e.g. 0.8.25). The action will do some checks (i.e. is there an equal or higher tagged version already present), prepare the changelogs for the release, create a commit and create a tag. The regular build action will then build and publish the package.

The Github action cannot be debugged from a branch :-( So, I built and debugged it on master (yolo): https://github.com/scalableminds/webknossos-libs/runs/4432677422?check_suite_focus=true

### Issues:
- fixes #499 

### Todos:
 - [ ] Updated Changelog


.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+        git add */Changelog.md
+          git status
+          git cat */Changelog.md
+
+        git commit -m "Release for v${VERSION}"
+        git tag v${VERSION}
+
+          git tag
+        # git push origin master
+        # git push --tags
```
philippotto, line 49: remove or uncomment before merging?

.github/workflows/release.yml
```diff
@@ -0,0 +1,49 @@
...
+        if git show-ref --tags "v${VERSION}" --quiet; then
+            echo "Version $VERSION already exists. Stopping."
+            exit 1
+        fi
+
+    - name: Make and push release
+      run: |
+        ./make_release.sh ${VERSION}
+        git add */Changelog.md
+          git status
```
philippotto, line 41: nit: formatting seems to a bit off

make_release.sh
```diff
@@ -0,0 +1,30 @@
+#!/usr/bin/env bash
+set -eEuo pipefail
+set +x
+
+if [[ $# -eq 0 ]] ; then
+    echo "Please supply a 'version' as argument."
```
philippotto, line 6:
```suggestion
    echo "Please supply a 'version' as argument (e.g., ./make_release.sh 1.0.0)."
```

or is it with a preceding "v"?

make_changelog.py
```diff
@@ -0,0 +1,76 @@
...
+    if line.startswith("## ") and i > unreleased_idx
+)
+
+# Clean up unreleased notes (i.e. remove empty sections)
+unreleased_notes = changelog_lines[(unreleased_idx + 2) : last_release_idx]
+
+empty_unreleased_sections = [
+    (i, line)
+    for i, line in enumerate(unreleased_notes)
+    if line.startswith("### ") and unreleased_notes[i + 1] == ""
```
philippotto, line 41: I think, there are two problems with this: 1) there might be a simple empty line after the heading as a "spacer" after which a entry follows <-- in that case, the section would be recognized as unreleased. 2) There might be an empty "bullet point" line which starts with a `-` <-- that section would be recognized as not-empty.

Both cases appear in the current changelog. One example:

```
### Changed

- `BoundingBox` and `Mag` are now immutable attr classes containing `Vec3Int` values. See breaking changes above.

### Fixed

-
```

I'd suggest to split by `### ` and then count the matches for `^- [^\n]+` (with `gm` for regex options)

philippotto (commented):
cool stuff :man_dancing: i left some suggestions

make_release.sh
normanrz: The "v"-prefix is only for the git tag, not for the scripts.

philippotto (approved):
Great stuff! Since I had to go through the trouble of updating the release notes manually, I'd vote to merge this quite soon, as it's sorely needed.

The newest changelog entries should be formatted correctly, so that the edge case I mentioned should not be blocking for now.

/edit: see my two newest comments

make_changelog.py
```diff
@@ -0,0 +1,76 @@
...
+    # Remove section heading
+    del unreleased_notes[i]
+    # Remove empty line
+    del unreleased_notes[i]
+
+
+# Update changelog
+lines_to_insert = [
+    "## Unreleased",
+    f"[Commits](https://github.com/scalableminds/voxelytics/compare/{this_version}...HEAD)",
```
philippotto, line 53: `voxelytics` is wrong (also see below)

philippotto: also `{this_version}` needs to be prefixed with `v` for the link to work.

jstriebel (commented):
Nice stuff ♥ Found nothing besides what @philippotto already commented

make_changelog.py
normanrz: Fixed in https://github.com/scalableminds/webknossos-libs/pull/503/commits/f1e67524c69afc61a7b8c5faa4f6f3e44b032aa6

make_changelog.py
normanrz: Fixed in https://github.com/scalableminds/webknossos-libs/pull/503/commits/e82cbc7ffecf62b793f73e9c86099e3d7b4244e4

make_changelog.py
```diff
@@ -0,0 +1,84 @@
...
+with open("Changelog.md") as f:
+    changelog_lines = list(s.rstrip() for s in f)
+
+# Determine new version
+this_version = sys.argv[1]
+today = datetime.date.today()
+today_str = f"{today.strftime('%Y')}-{today.strftime('%m')}-{today.strftime('%d')}"
+
+# Determine last version
+matches = re.finditer(r"^## \[(.*)\].*$", "\n".join(changelog_lines), re.MULTILINE)
```
philippotto, line 15:
```suggestion
matches = re.finditer(r"^## \[v?(.*)\].*$", "\n".join(changelog_lines), re.MULTILINE)
```

I just tested this and noticed that the changelog sometimes includes the "v" and sometimes it doesn't (which causes multiple problems, such as incorrect GH urls or the `last_version == this_version` check not working). Stripping the v away, makes the script more robust.

philippotto (approved):
Sweet! See my one last suggestion, but other than that it works very well (locally at least :smile:).

PR #504, opened 2021-12-07
Deprecate get_segmentation/color_layer and add plural form

philippotto:
### Description:
- `get_segmentation_layer()` and `get_color_layer()` should not be used, as they will fail if no or more than one layer exists for each category. Instead, `get_segmentation_layer()s` and `get_color_layer()s` should be used (if desired in combination with `[0]` to get the old, error-prone behavior).

### Todos:
 - [x] Updated Changelog


fm3 (approved):
LGTM. Not sure if we want to use  https://docs.python.org/3/library/exceptions.html#DeprecationWarning  here? But apparently it is not shown in all environments :shrug:

philippotto:
> Not sure if we want to use https://docs.python.org/3/library/exceptions.html#DeprecationWarning here? But apparently it is not shown in all environments shrug

Not sure about this, either, but I think it's not critical right now, which is why I skipped this. Hope that's alright!

PR #495, opened 2021-11-30
add a `wkcuber` command

normanrz:
### Description:
- Adds a env-wide `wkcuber` command as alias for `python -m wkcuber`

### Issues:
- fixes #447

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [x] Updated Changelog

PR #507, opened 2021-12-08
Restore cubing main method which takes namespace args 

philippotto:
### Description:
- [#495](https://github.com/scalableminds/webknossos-libs/pull/495) adapted the main-method so that no parameters are necessary so that `wkcuber` can be an executable. However, vx imports the `main` method and passes args which is why this was a breaking change. To fulfill both requirements I added a new `cube_with_args` function. In the long run, `typer` would probably help a great deal with this.

### Issues:
- follow-up for #495

### Todos:
 - [x] Updated Changelog


normanrz (approved):
Whoopsie!

PR #511, opened 2021-12-08
File dataset upload

### LINKED ISSUE #501: In webknossos client, during dataset upload, avoid zip file creation
Creating the zip file on the client can take a long time. On the server it is unpacked again, which can again take a long time and will create load on the server. Instead, the files should be sent individually. Since WKW will usually create 1GB-Files, this should not be too much overhead. The webknossos backend (and js frontend) already support both.

normanrz:
### Description:
- Refactors the dataset upload in the `webknossos` module to upload individual files of the dataset instead of zipping it upfront. This greatly reduces the time to upload and storage consumption locally. 
- In this PR, I also added support for relative file paths for uploading, which is required for the hierarchical structure of wkw.
- The default upload chunk size is now 100MB, which I think makes more sense for programmatic uploads.
- Simultaneous uploads are now set to 5.

### Issues:
- fixes #501 

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/client/_upload_dataset.py
+++ b/webknossos/webknossos/client/_upload_dataset.py
@@ -24,17 +24,19 @@ def _cached_get_upload_datastore(context: _WebknossosContext) -> str:
     raise ValueError("No datastore found where datasets can be uploaded.")
 
 
-def _walk_file(path: Path, base_path: Path) -> Iterator[Tuple[Path, int]]:
+def _walk_file(path: Path, base_path: Path) -> Iterator[Tuple[Path, Path, int]]:
     if path.is_dir():
         yield from _walk(path, base_path)
         return
     elif path.is_symlink():
         yield from _walk_file(path.resolve(), base_path)
         return
-    yield (path.relative_to(base_path), path.stat().st_size)
+    yield (path.resolve(), path.relative_to(base_path), path.stat().st_size)
 
 
-def _walk(path: Path, base_path: Optional[Path] = None) -> Iterator[Tuple[Path, int]]:
+def _walk(
+    path: Path, base_path: Optional[Path] = None
+) -> Iterator[Tuple[Path, Path, int]]:
     if base_path is None:
         base_path = path
     for p in Path(path).iterdir():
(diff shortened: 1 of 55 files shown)


webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -25,55 +24,74 @@ def _cached_get_upload_datastore(context: _WebknossosContext) -> str:
...
+    yield (path.resolve(), path.relative_to(base_path), path.stat().st_size)
+
+
+def _walk(
+    path: Path, base_path: Optional[Path] = None
+) -> Iterator[Tuple[Path, Path, int]]:
+    if base_path is None:
+        base_path = path
+    for p in Path(path).iterdir():
+        yield from _walk_file(p, base_path)
```
jstriebel, line 43: I think this can be simplified quite a bit, removing `_walk_file`:
```suggestion
def _walk(
    path: Path, base_path: Optional[Path] = None
) -> Iterator[Tuple[Path, Path, int]]:
    if base_path is None:
        base_path = path
    if path.is_dir():
        for p in Path(path).iterdir():
            yield from _walk(p, base_path)
    else:
        yield (path.resolve(), path.relative_to(base_path), path.stat().st_size)
```

I think it should be better to only resolve a path for the yield, not before in the symlink case, since the `relative_to` might yield something outside of the actual directory we're operating in.

jstriebel (approved):
Nice, LGTM :+1: This is really helpful! Just see my one comment below:

webknossos/webknossos/client/_upload_dataset.py
normanrz: Very good. I wrongly thought that `p.is_dir() == False` for symlink-dirs.

webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -25,55 +24,67 @@ def _cached_get_upload_datastore(context: _WebknossosContext) -> str:
     raise ValueError("No datastore found where datasets can be uploaded.")
 
 
+def _walk(
+    path: Path, base_path: Optional[Path] = None
+) -> Iterator[Tuple[Path, Path, int]]:
+    if base_path is None:
+        base_path = path
+    if path.is_dir():
+        for p in Path(path).iterdir():
```
jstriebel, line 33: One more minor thing I just found:
```suggestion
        for p in path.iterdir():
```

webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -67,7 +68,7 @@ def upload_dataset(dataset: Dataset) -> str:
     with Progress() as progress:
         with Resumable(
             f"{datastore_url}/data/datasets?token={datastore_token}",
-            simultaneous_uploads=5,
+            simultaneous_uploads=1 if "PYTEST_CURRENT_TEST" in os.environ else 5,
```
jstriebel, line 71: @normanrz The parallelism here was a problem when using the snapshots. I simply disabled it during the tests. Not ideal, but I think that should be ok.

jstriebel (approved):
:rocket:

webknossos/webknossos/client/_upload_dataset.py
normanrz: Thanks!

PR #516, opened 2021-12-09
Use `CONSTANT_Z` to downsample 2D data

### LINKED ISSUE #455: cuber should default to constant_z sampling for single-slice datasets
It does not make sense to downsample datasets that have only 2D (=single slice) layers with anything other than constant z

rschwanhold:
### Description:
- Downsampling 2D data with sampling modes other than `CONSTANT_Z` (`ISOTROPIC`, `ANISOTROPIC`) does not make sense. 
- If other modes are used, empty (black) data is loaded and combined with the actual data, which results in a darker image  

### Issues:
- fixes #455 

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/layer.py
+++ b/webknossos/webknossos/dataset/layer.py
@@ -558,6 +558,11 @@ class Layer:
         if max_mag is None:
             max_mag = calculate_default_max_mag(self.bounding_box.size)
 
+        if self._properties.bounding_box.size.z == 1:
+            if sampling_mode != SamplingModes.CONSTANT_Z:
+                warnings.warn("The sampling_mode was changed to 'CONSTANT_Z'. Downsampling 2D data with a differnt sampling mode lead to darker images.")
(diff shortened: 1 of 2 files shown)


webknossos/webknossos/dataset/layer.py
```diff
@@ -558,6 +558,13 @@ def downsample(
         if max_mag is None:
             max_mag = calculate_default_max_mag(self.bounding_box.size)
 
+        if self._properties.bounding_box.size.z == 1:
+            if sampling_mode != SamplingModes.CONSTANT_Z:
+                warnings.warn(
+                    "The sampling_mode was changed to 'CONSTANT_Z'. Downsampling 2D data with a differnt sampling mode lead to darker images."
```
fm3, line 564:
```suggestion
                    "The sampling_mode was changed to 'CONSTANT_Z'. Downsampling 2D data with a different sampling mode mixes in black and thus leads to darkened images."
```

PR #521, opened 2021-12-13
update generated client & use for dataset upload

jstriebel:
### Description:
- updates the generated client to the newest wk API, including swagger documentation for the token and json body for dataset upload reserve & finish. The json body annotations needed some manual corrections, which are automated in `__generate_client.py`
- uses the generated client during dataset upload for reserve & finish


fm3 (approved):
LGTM

PR #461, opened 2021-10-21
Globalize Floodfill Script

youri-k:
### Description:
- execute script to merge segmentation layer and volume annotation
- globalizes all floodfill operations that were cut short by the wK frontend

### Issues:
- fixes https://github.com/scalableminds/webknossos/issues/5763

### Todos:
 - [x] Updated Changelog

PR #528, opened 2021-12-16
Misc additions to Annotation class

philippotto:
### Description:
- adds some new features to the Annotation class:
  - `save_to_file()` writes an annotation to disk
  - `temporary_annotation_view` (naming up for debate) allows to open the volume layer in an annotation like a normal wk Dataset
  - possibility to open the NML file within an annotation directly (this is usually rather edge cases which is why I prefixed this: `_open_nml`)
- adds some tests for the annotation functionality
- fixes local execution of `./test.sh` by using `python -m` to ensure the correct `sys.path`
- DRYs some paths in the tests

### Issues:
- related to #461

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/annotation/annotation.py
```diff
@@ -124,6 +146,63 @@ def download(cls, annotation_path: str) -> "Annotation":
 
         return download_annotation(annotation_type, annotation_id)
 
+    def _skeleton_zip_path(self) -> Union[Openable, PathLike, str]:
```
jstriebel, line 149: Why should this return a Union? Can't this simply return `_ZipPath` or `Openable` if we don't want to be too specific? I'm all in for Unions on arguments, but the return types should be rather specific IMO

jstriebel, line 149: Why should this return a Union? Can't this simply return `_ZipPath` or `Openable` if we don't want to be too specific? I'm all in for Unions on arguments, but the return types should be rather specific IMO

webknossos/webknossos/annotation/annotation.py
```diff
@@ -124,6 +146,63 @@ def download(cls, annotation_path: str) -> "Annotation":
...
+        nml_files = [i for i in self._filelist if i.endswith(".nml")]
+        assert len(nml_files) == 1
+        return _ZipPath(self._zipfile, nml_files[0])
+
+    @contextmanager
+    def _open_nml(self) -> Iterator[IO[bytes]]:
+        """
+        This method can be used to open the annotation's inner NML
+        file directly.
+        """
```
jstriebel, line 159: Wouldn't this simply be the same as
```
with annotation._nml_file.open(mode="rb") as f:
    yield f
```
if we simplify the return type of `_nml_file`? I'm not sure if we need this method then, but also ok I guess.

webknossos/webknossos/annotation/annotation.py
```diff
@@ -124,6 +146,63 @@ def download(cls, annotation_path: str) -> "Annotation":
...
+        file_path = self._skeleton_zip_path()
+
+        if isinstance(file_path, Openable):
+            with file_path.open(mode="rb") as file_handle:
+                yield file_handle
+        else:
+            with open(file_path, "rb") as file_handle:
+                yield file_handle
+
+    def save_to_file(self, path: Union[Path, str]) -> None:
```
jstriebel, line 169: I'd rather call this `save` according to our method-naming discussion. Please also add a link to this in the table in the top-level `__init__.py`, so it appears on
https://docs.webknossos.org/api/webknossos.html

webknossos/webknossos/annotation/annotation.py
```diff
@@ -124,6 +146,63 @@ def download(cls, annotation_path: str) -> "Annotation":
...
+        assert Path(path).suffix == ".zip", "The target path should be a zip file."
+
+        if isinstance(self.file, str) or isinstance(self.file, PathLike):
+            copyfile(self.file, str(path))
+        else:
+            with open(path, "wb") as f:
+                f.write(self.file.getbuffer())
+
+    @contextmanager
+    def temporary_annotation_view(self) -> Iterator[Layer]:
```
jstriebel, line 183: Nice! I'd maybe rename it to `temporary_volume_annotation_layer` or `temporary_volume_annotation_layer_copy`. (`volume_annotation` for clarity, and I'd rather not use `view`, since it's a copy and writes on the returned object won't be reflected in the annotation.

webknossos/tests/test_annotation.py
```diff
@@ -0,0 +1,45 @@
...
+    annotation = wk.Annotation.download(
+        "https://webknossos.org/annotations/Explorational/61c20205010000cc004a6356"
+    )
+    assert annotation.dataset_name == "l4dense_motta_et_al_demo_v2"
+    assert len(list(annotation.skeleton.flattened_graphs())) == 1
+
+    annotation.save_to_file(TESTOUTPUT_DIR / "test_dummy_downloaded.zip")
+    annotation = wk.Annotation(TESTOUTPUT_DIR / "test_dummy_downloaded.zip")
+    assert annotation.dataset_name == "l4dense_motta_et_al_demo_v2"
+    assert len(list(annotation.skeleton.flattened_graphs())) == 1
```
jstriebel, line 45: I'd rather extract this into another function, but otherwise LGTM :+1:

jstriebel (commented):
Nice work, thanks a lot! I added a couple of comments, mostly API related:

webknossos/webknossos/annotation/annotation.py
philippotto: Not sure why I did it this way. I also noticed that the method already exists as `_nml_file`. Maybe it was refactored in parallel in another PR. I cleaned it up now, though.

philippotto:
Thanks for your feedback! Please have another look :)

jstriebel (approved):
Awesome, LGTM :100:

jstriebel:
PS: Please also add a changelog entry :writing_hand:

PR #529, opened 2021-12-17
Annotation: read correct volume-data-zip from nml

jstriebel:
### Description:
- Uses the volume.location info from the annotation's nml to find the correct volume data zipfile.

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/annotation/annotation.py
+++ b/webknossos/webknossos/annotation/annotation.py
@@ -9,6 +9,7 @@ from zipfile import ZipFile
 from attr import dataclass
 from boltons.cacheutils import cachedproperty
 
+import webknossos.skeleton.nml as wknml
 from webknossos.dataset import Dataset, Layer, SegmentationLayer
 from webknossos.skeleton import Skeleton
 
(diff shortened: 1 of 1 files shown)


webknossos/webknossos/annotation/annotation.py
```diff
@@ -61,8 +71,10 @@ def save_volume_annotation(
     ) -> Layer:
         # todo pylint: disable=fixme
         # the name is about to change with multiple volume annotations
```
fm3, line 73: the comment can go now :)

fm3 (approved):
Thanks for jumping in so quickly!
LGTM, but I did not test this separately. Did you test it or should I set it up on my side?

webknossos/webknossos/annotation/annotation.py
jstriebel, line 73:
```suggestion
```

jstriebel:
>Did you test it or should I set it up on my side? 

@fm3 I tested it with an annotation that still uses `data.zip` and a new one with `data_Volume.zip` and both work fine :+1:

webknossos/webknossos/annotation/annotation.py
```diff
@@ -60,9 +70,10 @@ def save_volume_annotation(
         self, dataset: Dataset, layer_name: str = "volume_annotation"
     ) -> Layer:
         # todo pylint: disable=fixme
```
philippotto, line 72:
```suggestion
        # todo pylint: disable=fixme
```

PR #531, opened 2021-12-17
compute the largestSegmentId only from available bboxes

normanrz:
### Description:
- Only computes the `largestSegmentId` from available bboxes in a volume annotation

### Todos:
 - [x] Updated Changelog


jstriebel (approved):
LGTM, assuming you tested this already :+1:

Issue #525, opened 2021-12-16
Fix flaky windows tests 


philippotto:
`wkcuber_win` fails sometimes for no obvious reasons. See this run for example:

https://github.com/scalableminds/webknossos-libs/runs/4547090867?check_suite_focus=true

PR #533, opened 2021-12-17
Make flaky windows CI step optional

philippotto:
### Description:
- Make flaky windows CI step optional, since it sometimes fails for no obvious reasons.

See https://github.com/scalableminds/webknossos-libs/runs/4561269724?check_suite_focus=true#step:7:147 which failed, but the step is still successful.

### Issues:
- workaround for [#525](https://github.com/scalableminds/webknossos-libs/issues/525)

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -204,7 +204,7 @@ jobs:
     - name: CLI tests
       continue-on-error: true
       shell: bash
-      run: poetry run "C:\Program Files\Git\bin\bash.EXE" --noprofile --norc tests/scripts/all_tests.sh || true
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -203,7 +203,7 @@ jobs:
 
     - name: CLI tests
```
jstriebel, line 204: Alternatively you could also set `continue-on-error` to true, which might make it easier to find failing jobs later, but also fine as is.
```suggestion
    - name: CLI tests
      continue-on-error: true
```

jstriebel (approved):
LGTM :+1:

PR #536, opened 2021-12-23
Consistently use resolved instead of absolute path

daniel-wer:
### Description:
- Consistently use pathlib.path.resolve instead of os.path.abspath to avoid issues like https://github.com/scalableminds/voxelytics/pull/2851
- We have started to do that in https://github.com/scalableminds/webknossos-libs/pull/492 but overlooked these last two occurrences
- I checked that there are no other occurrences of abspath in the code base

- [x] Test downsampling to check whether this works correctly now
  - See results in https://github.com/scalableminds/voxelytics/pull/2853

### Todos:
 - [x] Updated Changelog


georgwiese (approved):
🚀

Issue #537, opened 2021-12-23
cluster_tools slurm timeout issue


elhuhdron:
Thank you for implementing the new features that query for `MaxArraySize` and `MaxSubmitJobs` in slurm. Unfortunately, on our system I am still having an issue. On our system `MaxSubmitJobs` is not defined (allows for unlimited queued jobs), however, after 9 rounds of array job submissions (each 1001 in size), slurm responds to the next submission with a timeout-type error:

```
2021-12-22 23:15:23,214 DEBUG Number of currently submitted jobs is 9053.
Exception in thread Thread-13452:
Traceback (most recent call last):
  File "/gpfs/soma_fs/home/watkins/anaconda3/envs/wkw-new/lib/python3.8/threading.py", line 932, in _bootstrap_inner
    self.run()
  File "/gpfs/soma_fs/home/watkins/anaconda3/envs/wkw-new/lib/python3.8/site-packages/cluster_tools/schedulers/slurm.py", line 337, in run
    job_id = SlurmExecutor.submit_text(script, self.cfut_dir)
  File "/gpfs/soma_fs/home/watkins/anaconda3/envs/wkw-new/lib/python3.8/site-packages/cluster_tools/schedulers/slurm.py", line 154, in submit_text
    job_id, stderr = chcall("sbatch --parsable {}".format(filename))
  File "/gpfs/soma_fs/home/watkins/anaconda3/envs/wkw-new/lib/python3.8/site-packages/cluster_tools/util.py", line 65, in chcall
    raise CommandError(command, code, stderr)
cluster_tools.util.CommandError: 'sbatch --parsable .cfut/_temp_slurm9LFK0819HxEcazy2sTtbM72vFlHDuUlX.sh' exited with status 1: b'sbatch: error: Slurm temporarily unable to accept job, sleeping and retrying\nsbatch: error: Batch job submission failed: Resource temporarily unavailable\n'
```

I'm not sure how easily reproducible this is, particularly if you are testing with a docker instance. Although I like the automatic query, would it be possible to add manual control of these, so they can be specified as part of the job resources (i.e., how large of array size jobs to submit and how many can be submitted/running at once)? This would be an easy workaround.

elhuhdron:
I opened pull request #538 as a potential suggestion for the workaround. For me this resolved the issue, as I can specify a maxsubmit value below 8000 (I used 3000) and my cubing jobs completed.

daniel-wer:
[...]

PR #538, opened 2021-12-24
Optional user control of max_array_size and max_submit_jobs

elhuhdron:
### Description:
Allows user to specify values other than the system values for the number of jobs per slurm array and the maximum number of jobs submitted to slurm at once.

### Issues:
Allows for workaround of issue #537 

### Notes:
To me the easiest solution was to use the job_resources dict to allow for optional control over `max_array_size` and  `max_submit_jobs` (only `max_submit_jobs` is really needed for the workaround). But any method of control from the command line would be sufficient. In the case where the values are not specified, it defaults to the automatic reading behavior, short of the batch_size computation; my preference in the cases where this is relevant (multiple simultaneous jobs for example) would be to allow the user to decide how to handle this.

### DIFF
--- a/cluster_tools/cluster_tools/schedulers/slurm.py
+++ b/cluster_tools/cluster_tools/schedulers/slurm.py
@@ -187,14 +187,26 @@ class SlurmExecutor(ClusterExecutor):
         job_resources_lines = []
         if self.job_resources is not None:
             for resource, value in self.job_resources.items():
-                job_resources_lines += ["#SBATCH --{}={}".format(resource, value)]
+                # job_resources_lines += ["#SBATCH --{}={}".format(resource, value)]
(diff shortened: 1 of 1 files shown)


daniel-wer:
Thanks a lot for this suggestion on how to implement a workaround for #537.

I've opted for a slightly different approach in https://github.com/scalableminds/webknossos-libs/pull/554 using environment variables. This allows for better encapsulation and allows for easier consistent configuration across separate runs.
Also, I've removed that at most 1/3 of MaxArraySize and MaxSubmitJobs jobs are submitted in one batch. As you point out, these new configuration options make it easier for the user to decide how to handle this.

For the pointed out reasons we'll be merging #554 instead of this PR. Still, thanks a lot for your submission.

Issue #509, opened 2021-12-08
Allow to symlink existing public dataset layers when uploading dataset


fm3:
The webKnossos server API now ( https://github.com/scalableminds/webknossos/pull/5863 ) allows to symlink existing public dataset layers into an uploaded dataset. The client should be able to use this field and expose it to a user.

philippotto:
I'm not sure what exactly this should look like. From a user perspective, I would probably imagine an interface like this:

```python
ds = download_dataset(dataset_name, organization_name)
existing_layer = ds.get_layer("existing_layer", ...)
new_layer = ds.add_layer("new_layer", ...)
foreign_layer = ds.add_symlink_layer(foreign_layer_descriptor)

ds.upload()
```

However, it's not clear to me how the `upload` method would know which layer to symlink. For example, for the `existing_layer` one would need to know whether the user has mutated the layer or whether it's equivalent to the server side.

An explicit approach would be to require the user to pass the necessary information. Potentially:

```python
ds.upload(layers_to_symlink=["existing_layer"])
```

Or, even more information:

```python
ds.upload(layers_to_symlink=[
      {
        organizationName: "...",
        dataSetName: "...",
        layerName: "foreign_layer",
        newLayerName: "foreign_layer"
      }
])
```

which already feels quite complicated.

-----------------

For the (probably simpler) usecase, where I'm creating a fresh dataset on disk, I'm not even sure whether it's realistic to symlink something on the server (why would I create a brand-new DS in the first place which matches existing data?).

------------

All in all, I'm not yet seeing the big picture here. I hope you can help me with getting the details straight. I'm happy to implement a basic version which works somehow, but I'm not sure if this solves a use case satisfactorily.

Happy to hear your thoughts! (also cc @jstriebel)

fm3:
CCing @normanrz who brought up this feature request and may have more insights about the desired use case

normanrz:
I think this is what I was imagining:
```python
ds.upload(layers_to_link=[
      {
        organizationName: "...",
        dataSetName: "...",
        layerName: "foreign_layer",
        newLayerName: "foreign_layer"
      }
])
```

`organizationName` and `newLayerName` should be optional.

The primary use case for this feature is to upload new layers (e.g. a segmentation) to an existing dataset, without modifying the original.

Nit: I wouldn't call it `symlink`, because that is an implementation detail on the server-side.

jstriebel:
[...]

PR #543, opened 2022-01-04
Allow to specify linked layers when uploading a dataset

philippotto:
### Description:
- When uploading a dataset, one can pass layers which should be linked additionally.

For example, this script:
```python
from time import gmtime, strftime

import numpy as np
from skimage import data

import webknossos as wk
from webknossos.client._upload_dataset import LayerToLink
from webknossos.dataset import COLOR_CATEGORY


def main() -> None:
    # load your data - we use an example 3D dataset here
    img = data.cells3d()  # (z, c, y, x)

    # make sure that the dimension of your data has the right order
    # we expect the following dimensions: Channels, X, Y, Z.
    img = np.transpose(img, [1, 3, 2, 0])

    # choose a name for our dataset
    time_str = strftime("%Y-%m-%d_%H-%M-%S", gmtime())
    name = f"cell_{time_str}"

    # scale is defined in nm
    ds = wk.Dataset(name, scale=(11.24, 11.24, 28))

    # The example microscopy data has two channels
    # Channel 0 contains cell membranes, channel 1 contains nuclei.
    layer_membranes = ds.add_layer(
        "cell membranes",
        COLOR_CATEGORY,
        dtype_per_layer=img.dtype,
    )

    layer_membranes.add_mag(1, compress=True).write(img[0, :])

    url = ds.upload(
        layers_to_link=[
            LayerToLink(
                "scalable_minds",
                "l4dense_motta_et_al_demo_v2",
                "color",
                "color",
            )
        ]
    )
    print(f"Successfully uploaded {url}")


if __name__ == "__main__":
    main()

```

will produce this dataset: https://webknossos.org/datasets/scalable_minds/cell_2022-01-04_16-57-55/view#156,194,88,0,0.621

### Issues:
- fixes [#509](https://github.com/scalableminds/webknossos-libs/issues/509)

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -94,16 +117,17 @@ def upload_dataset(dataset: Dataset) -> str:
...
             client=datastore_client.with_timeout(None),  # type: ignore[arg-type]
             token=datastore_token,
             json_body=DatasetFinishUploadJsonBody.from_dict(
                 {
                     "uploadId": upload_id,
                     "organization": context.organization,
-                    "name": dataset.name,
+                    "name": new_dataset_name,
                     "needsConversion": False,
+                    "layersToLink": [layer._asdict() for layer in layers_to_link],
```
jstriebel, line 130: I'd rather remap the names here in order to use snake-case in the Namedtuple.

jstriebel (commented):
Nice! I'd change some things about the namedtuple, see my comments below. And I think we could directly use this feature in the `learned_segmenter` example, linking the color data and training segmentation instead of re-uploading it.

webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -22,15 +21,33 @@
...
+        context = _get_context()
+
+        print(
+            {
+                "dataSetName": self.dataset_name,
+                "layerName": self.layer_name,
+                "newLayerName": self.new_layer_name,
+                "organizationName": self.organization_name or context.organization,
+            }
+        )
```
jstriebel, line 43: Probably not meant to be commited :-)

jstriebel (approved):
LGTM :+1:

webknossos/examples/learned_segmenter.py
```diff
@@ -75,10 +76,18 @@ def main() -> None:
...
-    ):
-        url = dataset.upload()
+    with wk.webknossos_context(url="http://localhost:9000", token="secretSampleUserToken"):
+        url = dataset.upload(
+            layers_to_link=[
+                LayerToLink(
+                    annotation.dataset_name,
+                    "color",
+                    "color",
+                    "scalable_minds",
```
jstriebel, line 86: The `new_layer_name` is superfluous I think:

```suggestion
                    organization_name="scalable_minds",
                    dataset_name=annotation.dataset_name,
                    layer_name="color",
```

webknossos/examples/learned_segmenter.py
```diff
@@ -6,6 +6,7 @@
 from skimage.future import TrainableSegmenter
 
 import webknossos as wk
+from webknossos.client._upload_dataset import LayerToLink
```
jstriebel, line 9: Maybe we should re-import `LayerToLink` at some higher level, e.g. on the `client/__init__.py`, then it's also automatically re-exported toplevel, so it can be used as `wk.LayerToLink`? What do you think?

Issue #366, opened 2021-07-28
Use progress bar for downsampling


hotzenklotz:
I just started a new downsampling job through the CLI. However I did not find the log statements to be very helpful (see below). 

Ideas:
- show progress as progress bar (e.g. https://github.com/tqdm/tqdm) with interesting stats instead of printing each chunk bounding box
  - show the number of jobs/chunks which are done and are still queued, e.g. `3/10`
  - show an ETA for finishing
  - (show progress per minute)
  - likely all of the is already handled by a progress bar lib
- I would really love to know what Mag the downsampling is working on, e.g. "Starting to downsample 2-2-1 mag". This is especially important for the `auto` sampling mode, where I have no idea what the cuber is about to do. Perhaps, all mag planned downsampling steps could be displayed at start 
- log output is mixed in INFO and DEBUG log levels... It should probably be one or the other and hidden per default (e.g. enabled with --verbose flag)

Current state:
```
2021-07-28 11:07:48,846 INFO Using pool of 12 workers.
2021-07-28 11:07:48,869 INFO Downsampling of (32, 32, 64)
2021-07-28 11:07:48,875 INFO Downsampling of (32, 1024, 64)
2021-07-28 11:09:26,876 DEBUG Downsampling of (32, 32, 64) took 97.93967581s
2021-07-28 11:09:30,558 DEBUG Downsampling of (32, 1024, 64) took 101.68316174s
2021-07-28 11:09:30,562 INFO Downsampling of (32, 2048, 64)
2021-07-28 11:09:41,556 DEBUG Downsampling of (32, 2048, 64) took 10.99285913s
2021-07-28 11:09:43,620 INFO Downsampling of (512, 512, 64)
2021-07-28 11:10:12,622 INFO Downsampling of (512, 1536, 64)
```

PR #544, opened 2022-01-04
Adds a CLI for upload

### LINKED ISSUE #523: CLI tool for data upload
A CLI tool that wraps `Dataset.upload` would be nice. I think a good signature would be:

```
python -m wkcuber.upload --token ... --url https://webknossos.org --jobs 4 /path/to/dataset

WK_TOKEN=... WK_URL=https://webknossos.org python -m wkcuber.upload /path/to/dataset
```

normanrz:
### Description:
- Adds a CLI for Upload `python -m wkcuber.upload ...`
- Refactors `add_verbose_flag`: Rename old flag into `--silent` (because that is what it actually did) and add new `--verbose` flag to most CLI commands
- Re-adds some of the debug messages that #366 removed. They are `DEBUG` log level, so that they should not be displayed by default.

### Issues:
- fixes #523 

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/utils.py
```diff
@@ -22,6 +22,7 @@
 
 def time_start(identifier: str) -> None:
     times[identifier] = time.time()
+    logging.debug("{} started")
```
philippotto, line 25: Won't this just print `{} started`?

webknossos/webknossos/utils.py
```diff
@@ -147,14 +148,22 @@ def setup_logging(args: argparse.Namespace) -> None:
...
 
     parser.set_defaults(verbose=True)
 
 
+def add_verbose_flag(parser: argparse.ArgumentParser) -> None:
+    parser.add_argument(
+        "--verbose", help="Verbose output", dest="verbose", action="store_true"
+    )
+
+    parser.set_defaults(verbose=False)
```
philippotto, line 164: So, depending on which utility function a module uses, either `--verbose` OR `--silent` is available? I imagine this to be tedious to use, since it's hard to remember. Why not always make both args available, but only differ on the default? So, one function to add both flags which accepts a "default_verbose" boolean parameter.

wkcuber/wkcuber/upload.py
```diff
@@ -0,0 +1,65 @@
...
+
+if __name__ == "__main__":
+    args = create_parser().parse_args()
+    setup_logging(args)
+    url = (
+        args.url
+        if args.url is not None
+        else environ.get("WK_URL", "https://webknossos.org")
+    )
+    token = args.token if args.token is not None else environ.get("WK_TOKEN", None)
```
philippotto, line 60: This duplicates the logic from https://github.com/scalableminds/webknossos-libs/blob/2922e874641e96b50e27819379966579600d8f07/webknossos/webknossos/client/context.py#L123. I think, we should try to DRY this, as this will probably occur more often in the future.

philippotto (changes requested):
Cool stuff :+1: I only left a couple of nitpicks.

webknossos/webknossos/dataset/dataset.py
```diff
@@ -225,10 +225,17 @@ def layers(self) -> Dict[str, Layer]:
         """
         return self._layers
 
-    def upload(self) -> str:
+    def upload(self, jobs: Optional[int] = None) -> str:
+        """
+        Uploads this dataset to webKnossos.
+
+        If supplied, the `jobs` parameter will determine the number of simultaneous chunk uploads.
```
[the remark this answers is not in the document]
hotzenklotz: Warum nicht `If supplied, the `jobs` parameter will determine the number of simultaneous chunk uploads. Default  value 123.`?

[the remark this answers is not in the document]
philippotto: The downside would be that the default value is repeated in the docstring and could get out of sync with its actual definition. But it might be a valid trade-off. I don't have a strong opinion on this.

jstriebel (commented):
A more general question: Should this still be a part of wkcuber? I could imagine that we rather switch towards a CLI as a part of the webknossos package, basically duplicating the wkcuber functionalities for a while and deprecating wkcuber then. If that's the case, we might directly start implementing new jobs in webknossos, e.g. as `webknossos dataset upload …` (e.g. using typer). But also fine to merge this as-is, we can still discuss this later.

wkcuber/wkcuber/upload.py
normanrz: The difference is that the `url` has a default value, whereas `token` needs to be supplied. So I don't think it is a duplicate.

normanrz:
> A more general question: Should this still be a part of wkcuber? I could imagine that we rather switch towards a CLI as a part of the webknossos package, basically duplicating the wkcuber functionalities for a while and deprecating wkcuber then. If that's the case, we might directly start implementing new jobs in webknossos, e.g. as webknossos dataset upload … (e.g. using typer). But also fine to merge this as-is, we can still discuss this later.

I don't know how useful a CLI is for other tasks that operating on datasets. I haven't thought about good use cases for that, yet. I think of the `wkcuber` as the CLI for `webknossos`. We could rename that at some later point in time.

webknossos/webknossos/utils.py
normanrz: I agree that it would be better to decide whether to have verbose on by default or not uniformly for all CLI tasks.
I'd vote for silent by default (which is now possible with the new progress bars).

webknossos/webknossos/utils.py
normanrz: 🙈

wkcuber/wkcuber/upload.py
philippotto: Hm, what I meant is that that upload.py:58 basically duplicates the code from context.py:123 and upload.py:60 is a duplicate of context.py:124. You could at least import `DEFAULT_WEBKNOSSOS_URL` instead of using the hardcoded literal in L58 to make it a bit DRYer.

normanrz: Now, I understand what you meant. I re-used the constant :+1:

philippotto (approved):
Looks good to me :+1: Please add to the changelog that `--silent` was replaced with `--verbose` and that silent is the new default.

normanrz:
Thanks! I'll wait for your #543 to be merged in order to also support renaming from the CLI.

philippotto:
Oh, sorry, I mixed it up with #528. Ignore my comment.

philippotto:
Now the time has come ;)

PR #546, opened 2022-01-05
adapt to new wk sample user token

fm3:
The wk sample user token was changed in https://github.com/scalableminds/webknossos/pull/5931 – the examples need to be adapted here.

Do you know if the test cassettes need to be updated as well?


jstriebel:
> Do you know if the test cassettes need to be updated as well?

I think that's not necessary, since they are filtered from matching [here](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/tests/conftest.py#L28-L29).
But I guess we should bump webknossos then as well:
https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/test.sh#L13

jstriebel (approved):
:tada:

PR #553, opened 2022-01-10
[WIP] dataset view: cleanup signatures and bounding-box behavior

jstriebel:
### Description:
- abc

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog


webknossos/tests/test_dataset.py
```diff
@@ -1050,13 +1050,13 @@ def test_writing_subset_of_compressed_data() -> None:
...
             )
 
-        assert compressed_mag._mag_view_bounding_box_at_creation == BoundingBox(
+        assert compressed_mag.bounding_box == BoundingBox(
             topleft=(
                 0,
                 0,
                 0,
             ),
-            size=(120, 140, 160),
```
jstriebel, line 1059: `_mag_view_bounding_box_at_creation` used to be in the current mag. Instead `bounding_box` is now used for all checks and is handled in mag1.

webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -293,6 +295,20 @@ def align_with_mag(self, mag: Mag, ceil: bool = False) -> "BoundingBox":
...
+        roundup = self.topleft if ceil else self.bottomright
+        rounddown = self.bottomright if ceil else self.topleft
+        margin_to_roundup = roundup % mag_vec
+        aligned_roundup = roundup - margin_to_roundup
+        margin_to_rounddown = (mag_vec - (rounddown % mag_vec)) % mag_vec
+        aligned_rounddown = rounddown + margin_to_rounddown
+        if ceil:
+            return BoundingBox(aligned_roundup, aligned_rounddown - aligned_roundup)
+        else:
+            return BoundingBox(aligned_rounddown, aligned_roundup - aligned_rounddown)
```
jstriebel, line 310: I took the alignment implementation from the old `View._handle_compressed_write` implementation and compared it to the numpy version, it's quite a bit faster. Not sure if it's readable enough, though, I think the initial version is easier to grasp.
```suggestion
    def align_with_mag2(self, mag: Mag, ceil: bool = False) -> "BoundingBox":
        # TODO: consider alternative implementation using ints only  pylint: disable=fixme
        mag_vec = mag.to_vec3_int()
        roundup = self.topleft if ceil else self.bottomright
        rounddown = self.bottomright if ceil else self.topleft
        margin_to_roundup = roundup % mag_vec
        aligned_roundup = roundup - margin_to_roundup
        margin_to_rounddown = (mag_vec - (rounddown % mag_vec)) % mag_vec
        aligned_rounddown = rounddown + margin_to_rounddown
        if ceil:
            return BoundingBox(aligned_roundup, aligned_rounddown - aligned_roundup)
        else:
            return BoundingBox(aligned_rounddown, aligned_roundup - aligned_rounddown)
```
```suggestion
    def align_with_mag2(self, mag: Mag, ceil: bool = False) -> "BoundingBox":
        # TODO: consider alternative implementation using ints only  pylint: disable=fixme
        mag_vec = mag.to_vec3_int()
        roundup = self.topleft if ceil else self.bottomright
        rounddown = self.bottomright if ceil else self.topleft
        margin_to_roundup = roundup % mag_vec
        aligned_roundup = roundup - margin_to_roundup
        margin_to_rounddown = (mag_vec - (rounddown % mag_vec)) % mag_vec
        aligned_rounddown = rounddown + margin_to_rounddown
        if ceil:
            return BoundingBox(aligned_roundup, aligned_rounddown - aligned_roundup)
        else:
            return BoundingBox(aligned_rounddown, aligned_roundup - aligned_rounddown)
```
```
>>> timeit("bb.align_with_mag(wk.Mag(2), ceil=True)", globals=globals(), number=100000)
3.7116026320145465
>>> timeit("bb.align_with_mag2(wk.Mag(2), ceil=True)", globals=globals(), number=100000)
1.4745627499651164
```

jstriebel:
@philippotto The actual new signatures are not implemented yet, but I think most of the cleanup and preparation for it is done now. I think it might be easier to review this once now and again after I implemented the new signatures. Could you already have a look, I can also give you an intro in a call if you want :detective:

webknossos/webknossos/dataset/mag_view.py
```diff
@@ -82,19 +77,30 @@ def __init__(
...
             wkw.Dataset.create(str(self.path), self.header)
 
+    # Overwrites of View methods:
+    @property
+    def bounding_box(self) -> BoundingBox:
+        return self.layer.bounding_box.align_with_mag(Mag(self.name), ceil=True)
+
+    @property
+    def size(self) -> Vec3Int:
+        warnings.warn("[DEPRECATION] mag_view.size is deprecated.")
```
jstriebel, line 94: TODO: add alternative code :+1:

webknossos/webknossos/geometry/bounding_box.py
jstriebel: I'll rename the old version to `align_with_mag_slow` and add some tests that assert both methods do the same.

webknossos/pyproject.toml
```diff
@@ -31,6 +31,7 @@ wkw = "1.1.11"
 [tool.poetry.dev-dependencies]
 # autoflake
 black = "21.7b0"
+hypothesis = "^6.35.0"
```
jstriebel, line 34: @normanrz Using an MPL2 licensed lib as a dev-dependency should be fine, right?

normanrz: Yep

jstriebel:
@philippotto I think this is ready for review now. Two things are left open within this PR, but shouldn't block the review: Adapting all the docstrings, and also adapting the examples, so that the documentation uses the new API where possible.

webknossos/webknossos/dataset/view.py
```diff
@@ -29,53 +29,127 @@ def __init__(
...
         return self._read_only
 
+    def _get_mag1_bbox(
+        self,
+        abs_mag1_bbox: Optional[BoundingBox] = None,
+        rel_mag1_bbox: Optional[BoundingBox] = None,
+        abs_mag1_offset: Optional[Vec3IntLike] = None,
+        rel_mag1_offset: Optional[Vec3IntLike] = None,
+        mag1_size: Optional[Vec3IntLike] = None,
+        # current_mag_bbox: Optional[BoundingBox] = None,
```
philippotto, line 90: should this be removed?

webknossos/webknossos/dataset/view.py
```diff
@@ -29,53 +29,127 @@ def __init__(
...
+                abs_mag1_offset,
+                rel_mag1_offset,
+                abs_current_mag_offset,
+                rel_current_mag_offset,
+            ]
+        )
+        num_sizes = sum(i is not None for i in [mag1_size, current_mag_size])
+        if num_bboxes == 0:
+            assert num_offsets != 0, "You must supply an offset or a bounding box."
+            assert num_sizes != 0, "You must supply a size."
```
philippotto, line 108:
```suggestion
            assert num_sizes != 0, "You must supply a size or a bounding box."
```

webknossos/webknossos/dataset/view.py
```diff
@@ -29,53 +29,127 @@ def __init__(
...
+            assert num_offsets == 1, "Only one bounding-box can be supplied."
```
philippotto, line 109:
```suggestion
            assert num_offsets == 1, "Only one offset can be supplied."
```

webknossos/webknossos/dataset/view.py
```diff
@@ -29,53 +29,127 @@ def __init__(
...
+            assert num_sizes == 1, "Only one bounding-box can be supplied."
```
philippotto, line 110:
```suggestion
            assert num_sizes == 1, "Only one size can be supplied."
```

webknossos/webknossos/dataset/view.py
```diff
@@ -29,53 +29,127 @@ def __init__(
...
+        else:
+            assert num_bboxes == 1, "Only one bounding-box can be supplied."
+            assert (
+                num_offsets == 0
+            ), "A bounding-box was supplied, you cannot also supply an offset."
+            assert (
+                num_sizes == 0
+            ), "A bounding-box was supplied, you cannot also supply a size."
+
+        mag_vec = self._mag.to_vec3_int()
```
philippotto, line 120: Can be moved into the following else-block

webknossos/webknossos/dataset/view.py
```diff
@@ -139,40 +281,130 @@ def read(
...
-        offset = Vec3Int(offset)
-        size = self.size if size is None else Vec3Int(size)
+        current_mag_size: Optional[Vec3IntLike]
+        mag1_size: Optional[Vec3IntLike]
+        if absolute_bounding_box is None and relative_bounding_box is None:
+            if offset is None:
+                if size is None:
+                    assert (
+                        relative_offset is None and absolute_offset is None
+                    ), "You must supply size, when reading with an offset."
```
philippotto, line 291: the message makes sense, but at its execution `offset is None` is truthy, so something seems off to me?

webknossos/webknossos/dataset/view.py
```diff
@@ -139,40 +281,130 @@ def read(
...
+                        else:
+                            offset_param = "absolute_offset"
+                        warnings.warn(
+                            "[DEPRECATION] Using view.read(size=my_vec) only with a size is deprecated. "
+                            + f"Please use view.read({offset_param}=(0, 0, 0), size=size_vec * view.mag.to_vec3_int()) instead."
+                        )
+                        current_mag_size = size
+                        mag1_size = None
+                    else:
+                        current_mag_size = None
```
philippotto, line 307: current_mag_size is set to `None` in all branches of the level-2 if-block (`if offset is None`). I'd move it to the end of the if-block to deduplicate it.

webknossos/webknossos/dataset/view.py
```diff
@@ -203,24 +435,107 @@ def get_view(
...
+            ), "Failed to get subview. The calling view is read_only. Therefore, the subview also has to be read_only."
+
+        current_mag_size: Optional[Vec3IntLike]
+        mag1_size: Optional[Vec3IntLike]
+
+        if offset is None:
+            if size is None:
+                assert (
+                    relative_offset is None and absolute_offset is None
+                ), "You must supply a size, when using get_view with an offset."
```
philippotto, line 450: again offset is None at this place

webknossos/webknossos/dataset/view.py
jstriebel: I'd leave it like this, since an offset is already given in this case, so adding a size seems more natural I guess? But no strong opinion.

webknossos/webknossos/dataset/view.py
jstriebel: I'm not sure, what exactly seems off here? I guess this should be fine, since it fails in the case where `offset`, `size` and both bboxes are None, but one or both of `relative_offset` and `absolute_offset` are given. In this case, the size is missing, as we discussed in the chat.

webknossos/webknossos/dataset/view.py
jstriebel: There's still `current_mag_size = size`, so I don't think that's possible. `mag1_size` and `current_mag_size` are always set in pairs, since exactly one of them should be used.

webknossos/webknossos/dataset/view.py
jstriebel: Same as above, I think that's fine. Maybe the confusion is that offset, relative_offset and absolute_offset can be given separately?

jstriebel:
Thanks a lot for the review @philippotto. I applied the PR feedback and answered to some comments that were not yet clear to me. I'll add the documentation updates next.

> Not sure whether it's worth it to add some dedicated tests for the deprecation stuff?

I think that should be fine as-is.

webknossos/webknossos/dataset/view.py
philippotto: Whoops, you are right. Nevermind then.

webknossos/webknossos/dataset/view.py
philippotto: Ok, sorry, you are right of course. I think, I didn't negate the `assert` condition in my head and thought that the assertion is violated if `offset is None and size is None and relative_offset is None ...`, but it's obviously raised when `relative_offset is not None`. My bad!

webknossos/webknossos/dataset/view.py
philippotto: Yes, that probably contributed to my confusion :sweat_smile:

philippotto (approved):
Awesome :)

PR #554, opened 2022-01-11
Allow to configure MaxArraySize and MaxSubmitJobs via env variables

daniel-wer:
### Description:
- See title
- Also remove that the batch_size was set to be at most 1/3 of max_array_size and max_submit_jobs. This was added to avoid blocking other jobs if a user runs multiple programs, but it was undocumented and feels unexpected/intransparent. Since this PR allows to easily configure the slurm limits, that should be preferable imo.
- Test the new functionality

### Issues:
- fixes https://github.com/scalableminds/webknossos-libs/issues/537
- supersedes https://github.com/scalableminds/webknossos-libs/pull/538

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/cluster_tools/Changelog.md
+++ b/cluster_tools/Changelog.md
@@ -12,6 +12,7 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Breaking Changes
 
 ### Added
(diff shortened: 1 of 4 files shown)


philippotto (approved):
Looks great to me :+1:

elhuhdron:
Cool, the environment variable solution works for me. Thanks for the fix.

Issue #514, opened 2021-12-09
improve wk-lib snapshot tests


jstriebel:
Currently, some points should be addressed for the wk-lib snapshot tests:
* [x] Some requests are run against webknossos.org, which might leak user data. For this reason some snapshots are edited manually, which makes it hard to upgrade them.
* [x] The snapshots contain many timestamps, which makes them change, so that refreshing the snapshots creates large diffs. We should filter those timestamps so that a refresh only shows the necessary changes. There are [hooks for this in vcrpy](https://vcrpy.readthedocs.io/en/latest/advanced.html#filter-sensitive-data-from-the-request)
* [x] The current version of the local webknossos setup is a bit hidden in the `test.sh`, maybe we could default to the currently running webknossos.org version?
* [x] The database should be dropped or evolutions must be run when upgrading the local webknossos setup.
* [ ] The tests should be run regularly without snapshots, e.g. as a nightly test, so that we get notified when webknossos.org has breaking changes we didn't anticipate.
* [ ] `__generate_client.py` needs some cleanup:
  * use isolated wk
  * use extra module for responses
  * use test data
  * integrate into wk? see also https://github.com/scalableminds/webknossos/issues/6017

jstriebel:
pointers for myself:
* https://github.com/scalableminds/webknossos/wiki/Set-up-a-standalone-datastore-locally
* https://github.com/scalableminds/webknossos/blob/master/tools/postgres/prepareTestDB.sh

jstriebel:
Also, `__generate_client.py` needs some cleanup:
* use isolated wk
* use extra module for responses
* use test data
* integrate into wk?

PR #555, opened 2022-01-12
add check for dataset name before uploading #545

### LINKED ISSUE #545: dataset upload should check isValidNewName route first
When a dataset upload is attempted with a name that is already in use in webKnossos, the request will fail, but not with a super readable message.
The wK frontend first queries `/api/datasets/<organizationName>/<dataSetName>/isValidNewName` – the python client should probably do this too.

normanrz:
### Description:
- Checks for dataset availability before uploading
- Refactors `__generate_client.py` to work with local webknossos instance

### Issues:
- fixes #545 

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/client/_upload_dataset.py
```diff
@@ -88,6 +91,14 @@ def upload_dataset(
...
     if "PYTEST_CURRENT_TEST" in os.environ:
         simultaneous_uploads = 1
+    response = new_dataset_name_is_valid.sync_detailed(
+        organization_name=context.organization,
+        data_set_name=new_dataset_name,
+        client=context.generated_auth_client,
+    )
+    assert (
+        response.status_code == 200
+    ), f"Dataset name {context.organization}/{new_dataset_name} does not seem to be valid: {response}"
```
jstriebel, line 101: It seem's weird that it's not necessary to update the snapshots, since the upload is also tested. Probably some bug related to the testing, I'll check it when working on #514

jstriebel (commented):
LGTM, assuming you tested this already :+1: :shipit:

webknossos/webknossos/client/_upload_dataset.py
jstriebel: Nevermind, the tests are actually failing due to the tests :see_no_evil: I'll commit an update for those

normanrz: @jstriebel ping for cassette update

jstriebel: It was a bit more complex, please double-check my changes in commit 93f7398. I removed the cloning of the wk-repo and used a docker-compose file instead, where the databases are not persisted anymore. Also, I added some filters for vcr.py, so that the timestamps in some requests are ignored. More cleanup is coming with #514

jstriebel (approved):

PR #565, opened 2022-01-24
Python client: Mark viewConfiguration as varying json field

fm3:
new varying field in wk json response for annotation json, introduced in https://github.com/scalableminds/webknossos/pull/5967


jstriebel (approved):
:+1:

PR #549, opened 2022-01-06
Zarr to wkw converter

erjel:
While advertised [differently](https://webknossos.org/features) there seems no export from `zarr` arrays into the `wkw` format. Trivial approaches, such as loading the full data into memory and writing in bulk, become quickly prohibitive for large datasets which easily exhaust RAM resources of the the used system.

Therefore, I created the `convert_zarr.py` routine for the conversion of segmentation layers. The file borrows large parts from the `convert_raw.py` and probably can be extended to a base class for the chunk-wise conversion of numpy-like file formats such as `.npy`, `.hdf5`, `n5`, and others.

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [ ] Updated Changelog

PR #566, opened 2022-01-26
convert_zarr: add changelog entry

jstriebel:
### Description:
Followup for #549, adding a changelog entry and getting CI green :green_circle:

### DIFF
--- a/wkcuber/wkcuber/convert_zarr.py
+++ b/wkcuber/wkcuber/convert_zarr.py
@@ -117,9 +117,6 @@ def _zarr_chunk_converter(
 
     slices = bounding_box.to_slices()
     zarr_file = zarr.open(source_zarr_path, "r")
-    shape = zarr_file.shape
-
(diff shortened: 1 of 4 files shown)


wkcuber/wkcuber/convert_zarr.py
```diff
@@ -117,9 +117,6 @@ def _zarr_chunk_converter(
 
     slices = bounding_box.to_slices()
     zarr_file = zarr.open(source_zarr_path, "r")
-    shape = zarr_file.shape
-
-    slices = tuple(slice(s.start, min([s.stop, s_])) for s, s_ in zip(slices, shape))
```
jstriebel, line 122: This was superfluous, since the chunked bounding box was already set to the zarr-files' shape, see https://github.com/scalableminds/webknossos-libs/blob/1192ed5d757c53804e988c2681c8fb93d491cdf9/wkcuber/wkcuber/convert_zarr.py#L158

normanrz (approved):

PR #568, opened 2022-01-27
[wk-libs] make network-snapshots for tests readable & idempotent

jstriebel:
### Description:
Makes the network-snapshots for tests idempotent and more readable. There's still the exception of `webknossos/tests/cassettes/test_examples/test_user_times.yaml` which would currently yield user-data and is mostly left as-is for now.

The responses and requests are now adapted, so that
* zips are unpacked
* timestamps are filtered
* tokens are filtered
* lastUsedByUser is filtered
* loggedTime is filtered to years <= 2021
* resumable upload UUIDs are filtered
* varying metadata `content-type: multipart/form-data; boundary=…` is filtered

### Issues:
- part of #514


webknossos/tests/cassettes/test_generated_client/test_user_list.yaml
```diff
@@ -17,49 +17,42 @@ interactions:
...
+      minds","isTeamManager":false}],"experiences":{},"lastActivity":1532875515002,"isAnonymous":false,"isEditable":true,"organization":"scalable_minds","novelUserExperienceInfos":{"lastViewedWhatsNewTimestamp":1641906588957,"shouldSeeModernControlsModal":false,"hasSeenDashboardWelcomeBanner":true},"selectedTheme":"dark","created":1532875515006,"lastTaskTypeId":null},{"id":"5b8e93aa1e0000b90a4aa940","email":"robert.schwanhold@scalableminds.com","firstName":"Robert","lastName":"Schwanhold","isAdmin":false,"isDatasetManager":false,"isActive":false,"teams":[{"id":"5b3ce78557b7414e59269427","name":"scalable
       minds","isTeamManager":false}],"experiences":{},"lastActivity":1536070570797,"isAnonymous":false,"isEditable":true,"organization":"scalable_minds","novelUserExperienceInfos":{"shouldSeeModernControlsModal":true,"hasSeenDashboardWelcomeBanner":true},"selectedTheme":"auto","created":1536070570798,"lastTaskTypeId":null},{"id":"6006a44d0100007e0070d734","email":"jonathan@scm.io","firstName":"Jonathan","lastName":"Striebel","isAdmin":true,"isDatasetManager":false,"isActive":true,"teams":[{"id":"5b3ce78557b7414e59269427","name":"scalable
```
fm3, line 36: I am surprised to still see firstName, lastName and email of real users in this public repository. In my mind, this is unacceptable (while without explicit user consent) and in direct violation of https://webknossos.org/privacy – please [take steps to remove those entries](https://docs.github.com/en/authentication/keeping-your-account-and-data-secure/removing-sensitive-data-from-a-repository).

jstriebel: Good point, glad you noticed it! Not sure how this slipped through. For now I removed the entries in this list in this PR, we'll need to merge it first to master before fixing this with bfg since it does not change the master's HEAD.

fm3: Thanks for looking into this! Maybe we could add an assertion that no fields of these names are left at some point, so that this does not happen again. Note that annotation jsons also contain user information, not sure if those are around here somewhere (may only be _your_ info, since they are your annotations, which maybe you are comfortable publishing)

webknossos/tests/conftest.py
```diff
@@ -15,6 +20,9 @@
 
 from .constants import TESTOUTPUT_DIR
 
+### HYPOTHESIS STRATEGIES
```
fm3, line 23:
```suggestion
### HYPOTHESIS STRATEGIES (library to test many combinations for data class input)
```

webknossos/tests/conftest.py
```diff
@@ -62,13 +86,135 @@ def vcr_config() -> Dict[str, Any]:
...
+}
+
+_REPLACE_IN_REQUEST_MULTIFORM = {
+    r"\r\n": "\n",
+    r"--[0-9a-f]+": "--",
+    r"\d{4}-\d{2}-\d{2}T\d{2}-\d{2}-\d{2}__[0-9a-f\-]+": "2000-01-01T00-00-00__0011",
+}
+
+
+def _before_record_request(request: VcrRequest) -> VcrRequest:
```
fm3, line 101: please add comment that this is to create a readable (and relevant) diff even if the requests contain irrelevant changing fields and binary blobs

webknossos/tests/conftest.py
```diff
@@ -62,13 +86,135 @@ def vcr_config() -> Dict[str, Any]:
...
+
+def _unzip_if_possible(value: Any) -> Any:
+    if isinstance(value, bytes):
+        try:
+            with ZipFile(BytesIO(value)) as zipfile:
+                files = {}
+                for name in zipfile.namelist():
+                    entry: Any = zipfile.read(name)
+                    entry = _handle_special_formats(entry)
+                    if isinstance(entry, str) and name.endswith(".nml"):
```
fm3, line 186: extract to somewhere or adapt comment (this has nothing to do with zip blobs)

webknossos/tests/conftest.py
```diff
@@ -78,39 +224,42 @@ def _make_vcr_request(httpx_request: httpx.Request, **_kwargs: Any) -> VcrReques
...
-        "content": content,
+        "content": _handle_special_formats(httpx_reponse.content),
     }
 
 
 httpx_stubs._to_serialized_response = _to_serialized_response
 
 
+def _from_special_formats(value: Any) -> bytes:
+    if isinstance(value, dict) and "zip" in value:
```
fm3, line 239: add comment that this `zip` key is how this case was encoded while unpacking

fm3 (approved):
LGTM, please add some explaining high-level comments, especially for methods where the name was fixed and is not very descriptive. Looking forward to this change :) (and the next one with the isolated wk instance :D)

PR #572, opened 2022-01-28
[wk-libs] make network-snapshots for tests readable & idempotent

jstriebel:
### Description:
Makes the network-snapshots for tests idempotent and more readable. There's still the exception of `webknossos/tests/cassettes/test_examples/test_user_times.yaml` which would currently yield user-data and is mostly left as-is for now.

The responses and requests are now adapted, so that
* zips are unpacked
* timestamps are filtered
* tokens are filtered
* lastUsedByUser is filtered
* loggedTime is filtered to years <= 2021
* resumable upload UUIDs are filtered
* varying metadata `content-type: multipart/form-data; boundary=…` is filtered

### Issues:
- part of #514


jstriebel:
I re-opened #568 here, I don't think that was strictly necessary, but wanted to ensure that I keep the git history clean. All PR feedback from the original PR is applied @fm3

fm3 (approved):
I do like the diff line counter

PR #575, opened 2022-01-31
Fix automatic conversion of 3D tiffs which only have one page

### LINKED ISSUE #571: Auto-convert fails for single-page 3D tiffs
Tiff files which only have a single page (according to `Tifffile`) and still 3D cannot be converted to wkw automatically. The conversion process will properly detect that there are Z slices in z, but since there is only one page (and not Z pages) the z access is broken.

I don't know how we can use the `Tifffile` library to read the slices individually, but a simple workaround would be a preprocessing step like this:

```
from tifffile import imread, imwrite

def main():
    file_name = "problematic_file.tif"

    data = imread(file_name)
    imwrite(f"new_{file_name}", data)


if __name__ == "__main__":
    main()
```

The new file can then be converted without any problems. One could do this preprocessing step if there are more than z slices, but only one tiff page.

youri-k:
### Description:
- try to convert imagej file

### Issues:
- fixes #571 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/wkcuber/image_readers.py
+++ b/wkcuber/wkcuber/image_readers.py
@@ -176,6 +176,23 @@ class Dm4ImageReader(ImageReader):
 
 class TiffImageReader(ImageReader):
     @staticmethod
+    def _get_page_data(tif_file: TiffFile, page_index: int) -> np.ndarray:
+        if 0 <= page_index < len(tif_file.pages):
+            return tif_file.pages[page_index].asarray()
+        elif tif_file.is_imagej:
+            shape = tif_file.pages[0].shape
+            elements_per_page = 1
+            for i in shape:
+                elements_per_page *= i
+            series = tif_file.series[0]
+            tif_file.filehandle.seek(series.offset + page_index * np.dtype(series.dtype).itemsize * elements_per_page)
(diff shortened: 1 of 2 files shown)


wkcuber/wkcuber/image_readers.py
```diff
@@ -175,6 +175,26 @@ def read_dtype(self, file_name: Path) -> str:  # pylint: disable=unused-argument
 
 
 class TiffImageReader(ImageReader):
+    @staticmethod
+    def _get_page_data(tif_file: TiffFile, page_index: int) -> np.ndarray:
+        if 0 <= page_index < len(tif_file.pages):
+            return tif_file.pages[page_index].asarray()
+        elif tif_file.is_imagej:
+            shape = tif_file.pages[0].shape
```
philippotto, line 183:
```suggestion
            # ImageJ can export 3-dimensional, single-page tiff files.
            # In that case, we manually read from the correct filehandle
            # by seeking to the correct position.
            shape = tif_file.pages[0].shape
```

philippotto (approved):
Great workaround :) :tada:

hotzenklotz:
Such multipage! Much wow

PR #587, opened 2022-02-08
adds Google Analytics to docs

normanrz:
adds Google Analytics to docs


jstriebel (commented):
Don't we need a cookie-consent for this? I've read something about using GA4 without cookies, but not sure if you used that here.

jstriebel (approved):
> I configured GA to use a session-cookie, which imo does not require consent.

Cool! Don't know much about when cookie-consent is necessary, I guess it's fine then :+1:

PR #588, opened 2022-02-08
Add support for multiple volume layers in annotations

### LINKED ISSUE #490: Adapt volume annotation to support multiple layers
New webknossos versions will support volume annotations with multiple layers. When downloading such an annotation, there won't be one `data_Volume.zip`, anymore. Instead, multiple zips can be created with the following scheme:

```
  def volumeDataZipName(index: Int, isSingle: Boolean): String = {
    val indexLabel = if (isSingle) "" else s"_$index"
    val nameLabel = name.map(n => s"_$n").getOrElse("")
    s"data$indexLabel$nameLabel.zip"
  }
```

philippotto:
### Description:
- If ambiguous, the layer name has to be provided when reading from a volume layer in an annotation.

This PR is only fully compatible with webKnossos once https://github.com/scalableminds/webknossos/pull/6028#discussion_r801466842 is merged, since that PR fixes a naming issue in annotation downloads.

### Issues:
- fixes #490 

### Todos:
 - [x] Updated Changelog


webknossos/Changelog.md
```diff
@@ -12,7 +12,8 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Breaking Changes
 
 ### Added
- - Added AnnotationInfo, Project and Task classes for handling annotation information and annotation project administration. [#574](https://github.com/scalableminds/webknossos-libs/pull/574):
+- Added AnnotationInfo, Project and Task classes for handling annotation information and annotation project administration. [#574](https://github.com/scalableminds/webknossos-libs/pull/574):
```
philippotto, line 15: I removed a leading space.

webknossos/tests/test_annotation.py
```diff
@@ -36,6 +36,58 @@ def test_annotation_from_file() -> None:
         assert len(skeleton_lines) == 32
 
 
+@pytest.mark.vcr()
```
jstriebel, line 39: I guess that's not needed (anymore), since no download is used?
```suggestion
```

webknossos/tests/test_annotation.py
```diff
@@ -36,6 +36,58 @@ def test_annotation_from_file() -> None:
...
+def test_annotation_from_file_with_multi_volume() -> None:
+    annotation = wk.Annotation(
+        TESTDATA_DIR / "annotations" / "multi_volume_example_CREMI.zip"
+    )
+
+    volume_names = sorted(list(annotation.get_volume_layer_names()))
```
jstriebel, line 45: `sorted` can directly consume the iterable:
```suggestion
    volume_names = sorted(annotation.get_volume_layer_names())
```

webknossos/webknossos/annotation/annotation.py
```diff
@@ -92,11 +93,20 @@ def binary(self) -> Union[bytes, BinaryIO]:
...
+
     def save_volume_annotation(
         self,
         dataset: Dataset,
         layer_name: str = "volume_annotation",
         largest_segment_id: Optional[int] = None,
+        # If the annotation contains multiple volume layers,
+        # source_volume_name has to be provided.
+        # Use Annotation.get_volume_layer_names() to look up
+        # available layers.
```
jstriebel, line 108: Please add this or sth. similar to the docstring instead, so it shows up in the docs

webknossos/webknossos/annotation/annotation.py
```diff
@@ -107,9 +117,28 @@ def save_volume_annotation(
...
+                "The annotation contains multiple volume layers. "
+                "Please specify which layer should be used via `source_volume_name`."
+            )
+
+            volume = next(
+                filter(
+                    lambda volume: volume._get_name_or_id() == source_volume_name,
+                    self._nml.volumes,
+                )
+            )
```
jstriebel, line 135: Please also catch the case where nothing matches (I guess a `StopIteration` is reaised then)

jstriebel: Alternatively, you can also initialize `volume_zip_path` and use a regular for loop with an if and a `break`, might even be more readable, but no strong opinion :shrug:

jstriebel (approved):
Nice! Mostly LGTM, I left a few minor comments, but approving already :+1:

webknossos/tests/test_annotation.py
philippotto: You're right. I adapted this from `test_annotation_from_file` where it's probably not needed, either. I remember that I once wrote a test and it took me a while to get it working, since I forgot to add the decorator. So, maybe it would be an easier practice to always use the decorator? However, I see, that it's also a bit unusual to use something when it's not needed. What do you think?

webknossos/webknossos/annotation/annotation.py
philippotto: good idea. i went for the for-loop and added an assertion so that the error message looks nicer than the stopiteration.

webknossos/webknossos/annotation/annotation.py
```diff
@@ -92,24 +93,54 @@ def binary(self) -> Union[bytes, BinaryIO]:
...
         """
         Given a dataset, this method will save the
         volume annotation of this annotation into that dataset
         by creating a new layer.
         The largest_segment_id is computed automatically, unless provided
         explicitly.
+
+        source_volume_name has to be provided, if the annotation contains
+        multiple volume layers. Use Annotation.get_volume_layer_names() to look up
+        available layers.
```
jstriebel, line 116:
```suggestion
        `source_volume_name` has to be provided, if the annotation contains
        multiple volume layers. Use `get_volume_layer_names()` to look up
        available layers.
```
(the method `get_volume_layer_names()` will be auto-linked, since it's within the same module

webknossos/webknossos/annotation/annotation.py
```diff
@@ -92,24 +93,54 @@ def binary(self) -> Union[bytes, BinaryIO]:
...
+            volumeLocation = None
+            for volume in self._nml.volumes:
+                if volume._get_name_or_id() == source_volume_name:
+                    volumeLocation = volume.location
+                    break
+            assert (
+                volumeLocation is not None
+            ), f"The specified volume name {source_volume_name} could not be found in this annotation."
+
+            volume_zip_path = volumeLocation
```
jstriebel, line 139: Please use snake_case here. Also directly using `volume_zip_path` seems fine to me:
```suggestion
            volume_zip_path = None
            for volume in self._nml.volumes:
                if volume._get_name_or_id() == source_volume_name:
                    volume_zip_path = volume.location
                    break
            assert (
                volume_zip_path is not None
            ), f"The specified volume name {source_volume_name} could not be found in this annotation."
```

philippotto: argh, thanks for spotting this. context-switching between python and JS is killing me :laughing:

webknossos/tests/test_annotation.py
philippotto: cool idea! I just pushed that and removed the decorator everywhere else.

jstriebel (approved):
:tada:

jstriebel:
> @jstriebel mypy had some troubles with the `_get_name_or_id` method which is why I had to inline it :shrug: 

Weird error, but glad you found an easy fix :+1: 

> by the way, I noticed that mypy cannot really do typechecking on properties which use `@cachedproperty`. I noticed this because after I removed `_get_name_or_id`, mypy didn't complain even though it was still used. The decorator is currently only used in the annotation class. I briefly tried a newer mypy version because I saw [python/typing#985](https://github.com/python/typing/issues/985), but this didn't fix the issue. Just FYI..

Good point! I guess the main problem is, that boltons (from where we import`cachedproperty`) is not typed. There are some stubs available here:
https://github.com/wcooley/python-boltons-stubs/blob/master/boltons-stubs/cacheutils.pyi#L127-L130
Seems like this is not published to pypi, but we can add them (or at least the ones we need) to our repo: https://mypy.readthedocs.io/en/stable/stubs.html

philippotto:
> Good point! I guess the main problem is, that boltons (from where we import`cachedproperty`) is not typed. There are some stubs available here:
> https://github.com/wcooley/python-boltons-stubs/blob/master/boltons-stubs/cacheutils.pyi#L127-L130
> Seems like this is not published to pypi, but we can add them (or at least the ones we need) to our repo: https://mypy.readthedocs.io/en/stable/stubs.html

Ah, makes sense :+1: I added the cachedproperty stubs, since I consider them relatively important for the type coverage of the annotation module.

webknossos/stubs/boltons/strutils/__init__.py
```diff
@@ -0,0 +1,5 @@
+from typing import Any
+
+
+def __getattr__(name: str) -> Any:
+    ...
```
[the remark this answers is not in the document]
philippotto: Yes, otherwise mypy would find the stubs and then complain that strutils etc. don't exist.

webknossos/examples/learned_segmenter.py
```diff
@@ -20,7 +20,7 @@ def main() -> None:
 
     # Step 1: Download the dataset and our training data annotation from webKnossos to our local computer
     training_data_bbox = wk.BoundingBox.from_tuple6(
-        annotation.skeleton.user_bounding_boxes[0]
+        annotation.skeleton.user_bounding_boxes[0]  # type: ignore[index]
```
philippotto, line 23: Thanks for pushing this. Yesterday, I was unsure how to fix this (strictly speaking one would need to check that `user_bounding_boxes` is not None). Ideally, it would always default to an empty list. However, that's likely out-of-scope for this PR. So, let's tackle it at a different point.

jstriebel: I agree, cleaning things up a bit more in the skeleton class is definitely a good idea, but let's defer this.

PR #589, opened 2022-02-09
[wk-libs] Allow neg bbox topleft

jstriebel:
### Description:
It's useful that intermediate bboxes can have a negative topleft, since they might end up negative in an intermediate result (e.g. when padding), which usually is intersected with a valid positive bbox before reading/writing to it.
This PR removes the restriction on negative toplefts, and converts negative sizes to a positive counterpart, e.g.
```python
assert BoundingBox((10, 10, 10), (-5, 5, 5)) == BoundingBox((5, 10, 10), (5, 5, 5))
```
Both bboxes describe the same are, we simply ensure that the topleft is actually the topleft in all dimensions.

Also sneaked in two minor changes:
* [upgrade pdoc to fix some linkify behavior](https://github.com/scalableminds/webknossos-libs/commit/9886cf3d72294a1210bee71cfe71f614223f26cc)
* [warn when layers are not uploaded due to linking](https://github.com/scalableminds/webknossos-libs/commit/2e74b4134b245349b55a06ab715637d282556a91)
* [don't cancel ci builds on master](https://github.com/scalableminds/webknossos-libs/pull/589/commits/d772ecb08592abb7b474a09629bf49ccccfd592a)

### Issues:
- needed for https://github.com/scalableminds/voxelytics/issues/2882

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/geometry/bounding_box.py
+++ b/webknossos/webknossos/geometry/bounding_box.py
@@ -1,7 +1,7 @@
 import json
 import re
 from collections import defaultdict
-from typing import Any, Dict, Generator, Iterable, List, Optional, Tuple, Union, cast
+from typing import Dict, Generator, Iterable, List, Optional, Tuple, Union, cast
 
 import attr
 import numpy as np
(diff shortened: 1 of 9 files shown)


webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -16,15 +16,14 @@ class BoundingBox:
...
-    def validate_size(self, _attribute: Any, value: Vec3Int) -> None:
-        assert value.is_positive(), "size of a BoundingBox must not be negative"
-
     def __attrs_post_init__(self) -> None:
+        if not self.size.is_positive():
+            negative_size = self.size.pairmin(Vec3Int.zeros())
+            new_topleft = self.topleft + negative_size
+            new_size = self.size.pairmax(-self.size)
+            object.__setattr__(self, "topleft", new_topleft)
+            object.__setattr__(self, "size", new_size)
```
fm3, line 25: Needs explaining comment

fm3 (approved):
LGTM. Note that when integrating this in voxelytics, new assertions similar to the ones you added to layer and view may be neeeded.

jstriebel:
> Note that when integrating this in voxelytics, new assertions similar to the ones you added to layer and view may be neeeded.

We just introduced these restrictions in a new wk-libs version (0.9.0), vx is actually relying on the fact that the topleft may be negative.

PR #592, opened 2022-02-09
adds apply_merger_mode example

normanrz:
### Description:
- Adds a new example for applying merger mode annotations to segmentations
- It uses the LGPL-licensed `fastremap` module for performing the remapping, which is added as dev-dependency only
- In future work, this functionality could be integrated into the wklibs directly

### Todos:
 - [x] Updated Changelog


webknossos/examples/apply_merger_mode.py
```diff
@@ -0,0 +1,91 @@
...
+            for node in graph.nodes
+        )
+        for graph in nml.flattened_graphs()
+    ]
+
+    equiv_map = {}
+    for segment_ids in equiv_classes:
+        base = next(iter(segment_ids))
+        for segment_id in segment_ids:
+            equiv_map[segment_id] = base
```
jstriebel, line 50: This can be simplified a bit without having `equiv_classes` first. Also `ndarray.item()` is nice for this usecase:
```suggestion
    equivalence_map = {}
    for graph in nml.flattened_graphs():
        base = None
        for node in graph.nodes:
            segment_id = in_mag1.read(absolute_offset=node.position, size=(1, 1, 1)).item()
            if base is None:
                base = segment_id
            equivalence_map[segment_id] = base
```

webknossos/examples/apply_merger_mode.py
```diff
@@ -0,0 +1,91 @@
...
+
+    print(f"Found {len(equiv_classes)} equivalence classes with {len(equiv_map)} nodes")
```
jstriebel, line 52:
```suggestion
    print(f"Found {len(nml.flattened_graphs())} equivalence classes with {len(equiv_map)} nodes")
```

webknossos/tests/test_examples.py
```diff
@@ -56,6 +56,13 @@ def test_dataset_usage() -> None:
     assert data_in_mag2_subset.shape == (3, 256, 256, 16)
 
 
+@pytest.mark.block_network(allowed_hosts=[".*"])
+def test_apply_merger_mode() -> None:
+    import examples.apply_merger_mode as example
+
+    exec_main_and_get_vars(example)
```
jstriebel, line 63: please add some assertions about the output, e.g. reading some remapped parts and asserting the id

docs/src/webknossos-py/examples/apply_merger_mode.md
```diff
@@ -0,0 +1,13 @@
+# Apply Merger Mode Annotation
+
+This example applies a merger mode annotation to a segmentation.
+
+It downloads a [sample dataset](https://static.webknossos.org/data/l4_segmentation.zip) and uses the [Dataset API](dataset_usage.md) for reading and rewriting the `segmentation` layer.
```
jstriebel, line 5: Please explain a bit more about merger mode, that it generates the nml and link to the docs:
https://docs.webknossos.org/webknossos/volume_annotation.html#proof-reading-and-merging-segments

Is this a useful example which actually solves some splits? Then we could also link to the original dataset and annotation and the result, and maybe add a screenshot.

jstriebel (commented):
Nice example, great you added it! :rocket: Please see my comments:

webknossos/examples/apply_merger_mode.py
```diff
@@ -0,0 +1,102 @@
...
+        )
+        out_view.write(cube_data)
+
+    in_mag1.for_zipped_chunks(apply_mapping_for_chunk, out_mag1)
+
+    ########################################
+    # Optionally, downsample and re-upload #
+    ########################################
+
+    # out_layer.downsample()
```
normanrz, line 86: @jstriebel is there a way to disable this code from running in the tests, but showing it in the docs?

jstriebel: Yes, you can check if the env var `PYTEST_CURRENT_TEST` is set, see
https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/examples/learned_segmenter.py#L87

To not clutter it this much in the examples, we might also add some magic comment, where the line is then removed before execution by `exec_main_and_get_vars`, but let's just add this black magic when those checks keep piling up :magic_wand:

webknossos/examples/apply_merger_mode.py
```diff
@@ -0,0 +1,91 @@
...
+    nml = wk.Skeleton.load(Path("testdata/annotations/l4_sample__merger_mode.nml"))
+
+    ###############################################
+    # Download and open the corresponding dataset #
+    ###############################################
+
+    copyfileobj(
+        urlopen("https://static.webknossos.org/data/l4_segmentation.zip"),
+        open("testdata/l4_segmentation.zip", "wb"),
+    )
```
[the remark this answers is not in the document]
hotzenklotz: I assume the zip file contains only partial cutout / small bounding box of the segmentation. Can one specify a bounding box with `wk.Dataset.download()`? Otherwise the whole L4 segmentation will probably be downloaded :-/ ?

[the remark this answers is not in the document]
jstriebel: Yes, you can specify a bounding box (currently the arg is `bbox`, renaming to `bounding_box` would probably be appropriate): https://docs.webknossos.org/api/webknossos/dataset/dataset.html#Dataset.download

jstriebel (approved):
Awesome :+1: Please add some assertion to the test-code, apart from that everything LGTM

webknossos/examples/apply_merger_mode.py
```diff
@@ -0,0 +1,100 @@
...
+    nml = wk.Annotation.download(
+        "https://webknossos.org/annotations/Explorational/6204d2cd010000db0003db91"
+    ).skeleton
+
+    ###############################################
+    # Download and open the corresponding dataset #
+    ###############################################
+
+    dataset = wk.Dataset.download(
+        "l4_sample_dev", "scalable_minds", path="testoutput/l4_sample_dev"
```
jstriebel, line 24: We need to set the url here, since the context is set to localhost during the tests (to avoid leaked data):
```suggestion
        "l4_sample_dev",
        "scalable_minds",
        path="testoutput/l4_sample_dev",
        webknossos_url="https://webknossos.org",
```

PR #598, opened 2022-02-14
Warn when if-main-guard is missing when using multiprocessing

### LINKED ISSUE #496: Warn when multiprocessing fails due to missing if __name__ guard
If wklibs are used with multiprocessing (e.g., calling `downsample` on a layer), python's multiprocessing module will import the main module (if `spawn` is used which is the default on OS  X). If that module has side effects (i.e., it is not guarded with `if __name__ == "__main__":`), weird errors can occur.

On start up, we could spawn a dummy job with the clustertools and if that fails, we could emit

```
            raise AssertionError(
                """
                #########################################
                #########################################
                #########################################
                #########################################
                #########################################
                ## Multiprocessing setup does not work ##
                ## properly. Please check that you are ##
                ## using an if __name__ == "__main__"  ##
                ## guard in your main module.          ##
                #########################################
                #########################################
                #########################################
                #########################################
                #########################################
            """
            ) 
```

to help the user with what's wrong.

philippotto:
### Description:
- Warn when if-main-guard is missing when using multiprocessing.

### Issues:
- fixes #496

### Todos:
 - [x] Updated Changelog


cluster_tools/tests/tests.py
```diff
@@ -824,6 +824,18 @@ def test_preliminary_file_map():
...
+def test_multiprocessing_validation():
+
+    import sys
+    from subprocess import PIPE, STDOUT, Popen
+
+    cmd = [sys.executable, "guardless_multiprocessing.py"]
+    p = Popen(cmd, shell=False, stdin=PIPE, stdout=PIPE, stderr=STDOUT)
+    output = p.stdout.read()
+
+    assert "current process has finished its bootstrapping phase." in str(output), "S"
```
jstriebel, line 836: :heart:

jstriebel (approved):
Nice, LGTM :+1:

PR #602, opened 2022-02-16
Annotation improvements

### LINKED ISSUE #540: [client] Upload annotations
Add a function to assemble an annotation from the python lib (e.g. with a skeleton) that can be uploaded to webKnossos. Also, add documentation and an example for that.

### LINKED ISSUE #542: Create annotations from segmentation layers and/or NMLs
It would be great to be able to create volume annotations from the Python library. I think the interface could be to convert a layer of a Dataset into a volume annotation that can be uploaded to webKnossos.

jstriebel:
### Description:
This PR improves the Annotation handling. Annotations are not frozen anymore but may be changed. There are the following additions and changes:
- Allow to construct an Annotation from a skeleton.
- Allow to change various properties usually stored in the nml/skeleton, such as dataset-name (more to come)
- Allow to add & remove volume annotations
- Add annotation upload
- Allow to save an annotation
- Minor breaking changes (method & argument renames)

### Issues:
- fixes #540 
- fixes #542

### Todos:
- [x] Updated Changelog
- [x] Improve Tests
- [x] Improve Docs

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - master
+  pull_request: {}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}


jstriebel:
@fm3 I think this PR is now in a good state to get a first review. There are still some things todo, but let's check if the main refactoring looks valid to you as well, before I build more on top of that :cityscape:

fm3 (commented):
As discussed, I’d say move meta information about the annotation out of the Skeleton class :) Also added some naming suggestions. Otherwise, I think this is moving in a good direction!

webknossos/testdata/nmls/generated_skeleton_snapshot.nml
```diff
@@ -0,0 +1,41 @@
+<?xml version="1.0" encoding="utf-8"?>
+<things>
+  <parameters>
+    <experiment name="My NML" />
```
fm3, line 4: this is suspicious, should it by `My Dataset`? (or empty?)

webknossos/tests/test_skeleton.py
```diff
@@ -142,14 +136,13 @@ def test_import_from_nml() -> None:
 
 
 def test_simple_initialization_and_representations(tmp_path: Path) -> None:
-    nml = wk.Skeleton(name="my_skeleton", scale=(0.5, 0.5, 0.5), time=12345)
-    nml_path = tmp_path / "my_skeleton.nml"
+    nml = wk.Skeleton(dataset_name="ds_name", scale=(0.5, 0.5, 0.5))
+    nml_path = tmp_path / "ds_name.nml"
```
fm3, line 140: this could possibly be still my_skeleton.nml?

webknossos/webknossos/skeleton/skeleton.py
```diff
@@ -25,55 +20,49 @@
...
     """
     Contains metadata and is the root-group of sub-groups and graphs.
-    See the parent webknossos.skeleton.group.Group for methods about group and graph handling.
+    See the parent class Group for methods about group and graph handling.
     """
 
+    dataset_name: str
+    scale: Vector3
+    organization_id: Optional[str] = None
+    description: Optional[str] = None
```
fm3, line 29: I would have expected this in Annotation? :thinking:

fm3 (commented):
Wow, this is a lot of good stuff! Looking forward to this!

One thing I noticed is that the Skeleton class is still responsible for stuff like dataset name and scale, with the annotation delegating the properties. From our recent discussion I would have expected this in Annotation only. Is there a specific reason for this decision? It may be fair, I’d just like to understand in detail :)

Other than that I just added a few very minor comments.

As you already mentioned, tests are on the way, which I would have asked for now otherwise :D

webknossos/tests/test_skeleton.py
jstriebel: Yes, I search-replaced too much here.

fm3:
Fair point 🤔 This means that an annotation will always need a Skeleton, right? Even if it contains only a volume layer and no trees? I’m not 100% sure what is the right way to go here, maybe we should also discuss this with @philippotto

jstriebel:
> Fair point :thinking: This means that an annotation will always need a Skeleton, right? Even if it contains only a volume layer and no trees? I’m not 100% sure what is the right way to go here, maybe we should also discuss this with @philippotto

No, it does not need one during initialization, but an annotation always get's one. This is handled by `Annotation.__attrs_post_init__`. You can supply all arguments a skeleton gets also for the Annotation, which then automatically creates the internal skeleton.

webknossos/Changelog.md
```diff
@@ -10,20 +10,41 @@ For upgrade instructions, please check the respective *Breaking Changes* section
...
 ### Fixed
+- Fixed the download of skeleton-only annotations. [#602](https://github.com/scalableminds/webknossos-libs/pull/602)
 
 
 ## [0.9.8](https://github.com/scalableminds/webknossos-libs/releases/tag/v0.9.8) - 2022-02-28
 [Commits](https://github.com/scalableminds/webknossos-libs/compare/v0.9.7...v0.9.8)
 
 ### Added
-- Added `only_setup_mag` parameter to downsample-related methods in `Layer`. This parameter allows creating output magnifications before parallelizing downsampling invocations to avoid outdated writes to datasource-properties.json. [#610](https://github.com/scalableminds/webknossos-libs/pull/610)
-- Added `Task.create()` method to create tasks by prodiving a dataset name, location, and rotation. [#605](https://github.com/scalableminds/webknossos-libs/pull/605)
```
jstriebel, line 26: Those were already part of the previous release.

jstriebel:
> Fair for the user, but a little weird as a data model, to me. May still be the best option if the Skeleton needs those values.

True, it's a bit weird. I can think of two alternatives where we have those attributes on the Annotation, but also on the Skeleton:
1. We could make the parts that should be in an annotation and in the skeleton read-only on the skeleton (and maybe also the annotation). Then they can not diverge later (because they can't be set), and we could assert that they agree if a skeleton is added.
2. We try to backlink the annotation from the skeleton, and adapting the skeleton's parameters changes the one on the annotation as well. It's a bit messy, since we need to update the skeleton's annotation when we try to move it from one annotation to another, and then it's linked differently.

Atm, I think the current version is cleaner than approach 2. and has better usability than 1., so I'd keep it for the moment. As long as we don't introduce multiple skeletons (and I don't see a reason for that, since we have groups) that should be ok I think.

fm3 (approved):
I think I’ve looked at most of this now, and what I have seen looked good :) The changes are a bit too complex for a line-by-line review, though, so I might have missed small things. Still, if we encounter any problems we can still fix them then, so I’d say this is good to go!

Issue #541, opened 2022-01-04
[client] Add CRUD operations for Tasks, TaskTypes, Projects etc.


normanrz:
It should be possible to create, read, edit and delete tasks and their adjacent types from the Python library. It would also be great to have downloads of the annotations of projects etc.

PR #605, opened 2022-02-17
Added "standard" task creation

hotzenklotz:
### Description:
-  Added "standard" task creation without the need for any base annotation. Rather, tasks a specified by manually providing a dataset name, starting location for the task, and camera rotation at that location.
- Refactored code to use the task API responding handling for both "manual" task creation and "by base annotation" creation
- Equivalent to this UI option:
image

### Issues:
- adds to #541

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/administration/task.py
+++ b/webknossos/webknossos/administration/task.py
@@ -12,7 +12,7 @@ from webknossos.client._generated.api.default import (
     task_info,
 )
 from webknossos.client.context import _get_generated_client
-from webknossos.geometry import BoundingBox
+from webknossos.geometry import BoundingBox, Vec3Int
 
 logger = logging.getLogger(__name__)
 
@@ -64,9 +64,11 @@ class Task:
         bounding_box: Optional[BoundingBox] = None,
     ) -> List["Task"]:
         """Submits tasks in webKnossos based on existing annotations, and returns the Task objects"""
+
         assert (
             len(base_annotations) > 0
         ), "Must supply at least one base annotation to create tasks"
+
         client = _get_generated_client()
         url = f"{client.base_url}/api/tasks/createFromFiles"
         task_parameters = {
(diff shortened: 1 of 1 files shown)


webknossos/webknossos/administration/task.py
```diff
@@ -97,31 +100,56 @@ def create_from_annotations(
...
+        project_name: str,
+        needed_experience_domain: str,
+        needed_experience_value: int,
+        dataset_name: str,
+        starting_position: Vec3Int,
+        starting_rotation: Optional[Vec3Int] = Vec3Int(0, 0, 0),
+        instances: int = 1,
+        script_id: Optional[str] = None,
+        bounding_box: Optional[BoundingBox] = None,
+    ) -> List["Task"]:
```
jstriebel, line 119: I'd move `dataset_name` up a bit, since I hope that the experience arguments might be optional in the future. I'd also add a default for the starting position, this might be fine for small datasets and is coherent with the UI.
```suggestion
    def create(
        cls,
        task_type_id: str,
        project_name: str,
        dataset_name: str,
        needed_experience_domain: str,
        needed_experience_value: int,
        starting_position: Vec3Int = Vec3Int(0, 0, 0),
        starting_rotation: Optional[Vec3Int] = Vec3Int(0, 0, 0),
        instances: int = 1,
        script_id: Optional[str] = None,
        bounding_box: Optional[BoundingBox] = None,
    ) -> List["Task"]:
```

jstriebel (approved):
Nice, mostly LGTM :+1: Just added one nitpick, please see my comment:

webknossos/webknossos/administration/task.py
hotzenklotz: I thought about using a default value for the `starting_position` but discarded it soon:
1. The UI has position and rotation as required arguments
2. A good default could be the centroid of the dataset's BB. But there is no way of requesting that from the server yet.
3. Alternatively, one could use the centroid location provided by the (optional) bounding box

IMH, `0, 0, 0` is not a good default, since it will move the XY viewport in such a way that the data is only in the lower right corner - 3/4 of the screen remain black.

PR #610, opened 2022-02-25
Add only_setup_mag(s) parameter to downsample methods

philippotto:
### Issues:
- Avoids metadata consistency problems related to [#524](https://github.com/scalableminds/webknossos-libs/issues/524)

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [x] Updated Changelog

PR #614, opened 2022-02-28
Add allow_overwrite parameter to Layer.downsample()

philippotto:
### Description:
- Add allow_overwrite parameter to Layer.downsample()
- Necessary to make use of the only_setup_mags parameter. Otherwise, later calls to downsample will complain that the mag already exists.

### Issues:
- follow up to #610

### Todos:
 - [x] Updated Changelog


jstriebel (approved):
LGTM :+1: We should unify `allow_overwrite` vs `exists_ok` at some point, but that's fine for now I guess.

Issue #524, opened 2021-12-15
Dataset: improve handling of multiple instances of the same dataset


jstriebel:
In case a Dataset is opened multiple times, former writes of another instance would currently silently be overwritten. We should raise errors in those cases and completely prevent them where possible:
* We could cache the datasource-properties content in a dataset and assert that it's still unchanged before we rewrite the file (atomically if possible). If it's changed, we should raise an error.
* When a dataset is opened multiple times in the same thread, we might return the first dataset instead of creating a new one to ensure a shared state.

jstriebel:
cc @normanrz

normanrz:
A few thoughts:
* Maybe the data should not be cached at all, but always read from disk?
* `*.lock` files could help to avoid concurrent writes to the file.
* I would not optimize for the single-node multiprocess/thread case, but instead for the general slurm/cloud/multiprocess/etc case.

jstriebel:
I'm not sure if we can control what is cached, since the user-code might re-use previous results to change the datasource-properties in latter calls. So we don't have any control over what is cached as soon as we expose any data from the datasource-properties. As an extreme example, a user might expand the bounding box of a layer by different degrees in parallel:
```
def expand_bounding_box_by_50():
    ds = …
    ds.get_layer("color").bounding_box += (50, 50, 50)
    # This is not atomic and can not be locked implicitly on our side, since there's a read and write from user-code.

def expand_bounding_box_by_100():
    ds = …  # same ds as above
    ds.get_layer("color").bounding_box += (100, 100, 100)

```

In this case, it might happen that the bounding box is expanded by 150 as expected, but also it might happen that it's just expanded by 50 or 100, e.g. using the following order:
```
read_bbox_before_50 = ds.get_layer("color").bounding_box
read_bbox_before_100 = ds.get_layer("color").bounding_box
ds.get_layer("color").bounding_box = read_bbox_before_100 + (100, 100, 100)
ds.get_layer("color").bounding_box = read_bbox_before_50 + (50, 50, 50)
```

We can make some operations atomic where all read and writes happen within one user-call (e.g. downsampling) by using locking or not caching data. But as soon as datasource-properties data is read and modified outside of the lib, we can only detect such cases and offer nicer solutions (such as allowing the user to add a lock).

Also, I think that the single-threaded multi-instance case is quite important to manage, too, and can be fixed, in contrast to the parallel case:
```
ds_for_50 = …
[...]

PR #615, opened 2022-02-28
Share dataset instances by caching them

philippotto:
### Description:
- By default, dataset instances are now shared by caching them.
- Use `dont_use_instance_cache=True` to avoid this (used in one test, too)

### Issues:
- contributes to [#524](https://github.com/scalableminds/webknossos-libs/issues/524)

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/stubs/boltons/cacheutils/__init__.py
+++ b/webknossos/stubs/boltons/cacheutils/__init__.py
@@ -20,3 +20,6 @@ class cachedproperty(Generic[T, KT]):
 class LRU(dict):
     def __init__(self, max_size: Optional[int]):
         ...
+
+    def clear(self) -> None:
+        ...
--- a/webknossos/stubs/boltons/cacheutils/__init__.py
+++ b/webknossos/stubs/boltons/cacheutils/__init__.py
@@ -15,3 +15,8 @@ class cachedproperty(Generic[T, KT]):
 
     def __get__(self, obj: Any, objtype: Optional[Type]) -> T:
         ...
+
+
+class LRU(dict):
+    def __init__(self, max_size: Optional[int]):
+        ...
--- a/webknossos/tests/conftest.py
+++ b/webknossos/tests/conftest.py
@@ -17,6 +17,7 @@ from vcr.stubs import httpx_stubs
 
 import webknossos as wk
 from webknossos.client.context import _clear_all_context_caches
+from webknossos.dataset.dataset import _clear_instance_cache
 
 from .constants import TESTOUTPUT_DIR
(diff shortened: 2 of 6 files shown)


philippotto:
> I have thought about this a bit and I am not sure, if caching is a good idea. I am weary that this will have unintended consequences, such as difficulties when serializing the objects for cluster distribution and surprising behavior for library users. 

I'm also not super convinced by this approach after I had to dig through the failing tests. However, my concerns are a bit different. The serialization you mentioned is automatically tested and I don't see immediate problems for that. Surprising behavior for users would also be rather unlikely in my opinion. However, I see the following downsides:

- additional complexity in implementation/maintenance and program behavior
- potential memory leaks and dangling/hidden file handles (since the cache is usually not cleared)
- could (how likely?) create hard to debug problems (until now I only observed far fetched edge cases due to tests which deal with obscure scenarios)

In combination with the fact, that this doesn't do anything for multiprocessing/cluster setups, I feel like the downsides dominate.
I'd be fine with closing the PR. @jstriebel What are your thoughts?

> For Voxelytics, where we mostly have map-reduce-style tasks, I think it should be feasible to set/fix the bounding box in pre or post steps.

Yes, this has already been done in https://github.com/scalableminds/webknossos-libs/pull/610 and https://github.com/scalableminds/voxelytics/pull/2911.

jstriebel:
> In combination with the fact, that this doesn't do anything for multiprocessing/cluster setups, I feel like the downsides dominate.
I'd be fine with closing the PR. @jstriebel What are your thoughts?

I agree that caching the instances might also lead to confusion for users, just like the current problems, and not caching anything is simpler (implementation-wise, and to think about in weird cases). I still think detecting the problematic cases when another instance overwrites outdated datasource-properties is most important, we could have a warning for it (shouldn't be an exception, though). (It might be useful to elevante this warning to an error in vx since we should handle all cases there).

In my opinion this warning would be enough to handle #524, let's close this PR then :+1:

philippotto:
> I agree that caching the instances might also lead to confusion for users, just like the current problems, and not caching anything is simpler (implementation-wise, and to think about in weird cases). I still think detecting the problematic cases when another instance overwrites outdated datasource-properties is most important, we could have a warning for it (shouldn't be an exception, though). (It might be useful to elevante this warning to an error in vx since we should handle all cases there).

Ok, agree. I'll look into the detection thing. Should be less tedious than the caching (and also easier to test now that there will be no caching :smile:).

Closing this now :wave: (just when the CI turned green for the first time :disappointed:)

PR #616, opened 2022-02-28
CI: don't run everything twice in PRs

jstriebel:
Runs CI only for push on master, to get CI feedback on branches, please open a draft PR

Taken from https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662

before:
Screenshot from 2022-02-28 17-37-34

after (see the names, job status does not matter):
Screenshot from 2022-02-28 17-37-48

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
     - master
-  pull_request:
+  pull_request: {}
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ci.yml
```diff
@@ -1,6 +1,10 @@
 name: CI
 
-on: [push, pull_request]
+on:
+  push:
+    branches:
+    - master
+  pull_request:
```
normanrz, line 7: is this valid yaml?
```suggestion
  pull_request: {}
```

normanrz (approved):

.github/workflows/ci.yml
jstriebel: seems so, just copied it from the link. But the empty dict looks nicer :ok_hand:

PR #618, opened 2022-03-01
Fix macos network tests

leowe:
### Description:
- Since macOS uses `spawn` in python multiprocessing, it uses also sockets to communicate between them
- If all network is disabled, the tests that rely on multiprocessing in some form will fail
- To make the tests work, a marker is added to each test to allow for kind of socket communication (the address will be at something like `/var/abc/longname`)


webknossos/tests/conftest.py
```diff
@@ -316,3 +317,10 @@ def pytest_collection_modifyitems(items: List[pytest.Item]) -> None:
     for item in items:
         if item.get_closest_marker("vcr") is None:
             item.add_marker("vcr")
+
+        if (
+            item.get_closest_marker("block_network") is None
+            and multiprocessing.get_start_method() != "fork"
+        ):
+            marker = pytest.mark.block_network(allowed_hosts=["/.*"])
```
fm3, line 325: please add comment explaining that /.* means unix sockets for multiprocessing

fm3 (approved):
Code LGTM, please add explaining comment, though

Issue #551, opened 2022-01-06
nearest interpolation does not work


erjel:
Hi,

for me the nearest interpolation does not work in the downsampling.

Test code to create the dataset:
```python
import numpy as np
from webknossos import Dataset

data = np.random.randint(0, 255, (1024, 1024, 1024), dtype=np.uint8)

Dataset('sample_wkw', scale=(256.0, 256.0, 35.0))\
    .get_or_add_layer('segmentation', 'segmentation', dtype_per_layer=np.uint8, largest_segment_id=int(255))\
    .add_mag(1)\
    .write(data)
```
and the cli run:
```bash
poetry run python -m wkcuber.downsampling --layer_name segmentation -i nearest -j 10 ./debug
``` 
produces an error:
```
2022-01-06 10:28:49,644 INFO Using pool of 1 workers.
Downsampling layer segmentation from Mag 1 to Mag 1-1-2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% 0:00:00 | -:--:--2022-01-06 10:28:49,907 DEBUG the selected filtering strategy does not support anisotropic downsampling. Selecting 1 as uniform downsampling factor
Downsampling layer segmentation from Mag 1 to Mag 1-1-2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% 0:00:02 | -:--:--2022-01-06 10:28:51,877 DEBUG the selected filtering strategy does not support anisotropic downsampling. Selecting 1 as uniform downsampling factor
Downsampling layer segmentation from Mag 1 to Mag 1-1-2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━   0% 0:00:04 | -:--:--2022-01-06 10:28:53,978 ERROR Downsampling of target BoundingBox(offset=Vec3Int(0,0,0), size=Vec3Int(1024,1024,512)) failed with could not broadcast input array from shape (256,256,512) into shape (256,256,256)
2022-01-06 10:28:53,981 ERROR A future crashed with an exception: could not broadcast input array from shape (256,256,512) into shape (256,256,256). Future: <Future at 0x7fd7a8964370 state=finished raised ValueError>
[...]

PR #619, opened 2022-03-01
Fix nearest interpolation

leowe:
### Description:
- When executing `wkcuber.downsampling` with an anisotropic and `nearest` interpolation mode, the execution crashed, because of misaligned bounding boxes
- Currently the `nearest` interpolation mode doesn't support anisotropic downsampling
- Add comments and docstrings regarding the combination of `interpolation_mode`s and `anisotropic` when downsampling
- Add assertion for fit of `interpolation_mode` and `anisotropic` instead of only logging it

### Issues:
- contributes to #551


wkcuber/wkcuber/downsampling.py
```diff
@@ -109,6 +109,15 @@ def downsample_mags(
     sampling_mode: str = SamplingModes.ANISOTROPIC,
     force_sampling_scheme: bool = False,
 ) -> None:
+    """
+    Argument `path` expects the directory containing the dataset.
+    Argument `layer_name` expectes the name of the cubed layer (color or segmentation).
+    Argument `from_mag` expects the resolution to base downsampling on.
+
+    For the other parameters see the cli help or `Layer.downsample` and `Layer.downsampling_mag`.
```
philippotto, line 117:
```suggestion
    For the other parameters see the CLI help or `Layer.downsample` and `Layer.downsampling_mag`.
```

philippotto (approved):
Great :+1:

PR #621, opened 2022-03-02
add MagView.get_views_on_disk()

### LINKED ISSUE #532: Add MagView.get_views_on_disk
[MagView.get_bounding_boxes_on_disk](https://docs.webknossos.org/api/webknossos/dataset/mag_view.html#MagView.get_bounding_boxes_on_disk) is very useful. However, there may be some confusion, because it returns global offsets, even when the `MagView` has its own offset. It would be great to add a new method that returns `View`s directly.

Also, `MagView.get_bounding_boxes_on_disk` doesn't return bboxes but `(offset, size)` pairs.

jstriebel:
### Description:
- adds `MagView.get_views_on_disk()` and uses it

### Issues:
- fixes #532 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/skeleton/__init__.py
+++ b/webknossos/webknossos/skeleton/__init__.py
@@ -9,7 +9,7 @@ from webknossos.skeleton.skeleton import Skeleton
 
 
 def open_nml(file_path: Union[PathLike, str]) -> Skeleton:
-    """Deprecated."""
+    """open_nml is deprecated, please use Skeleton.load instead."""
     warnings.warn(
         "[DEPRECATION] open_nml is deprecated, please use Skeleton.load instead."
     )
--- a/webknossos/webknossos/skeleton/__init__.py
+++ b/webknossos/webknossos/skeleton/__init__.py
@@ -9,6 +9,7 @@ from webknossos.skeleton.skeleton import Skeleton
 
 
 def open_nml(file_path: Union[PathLike, str]) -> Skeleton:
+    """Deprecated."""
     warnings.warn(
         "[DEPRECATION] open_nml is deprecated, please use Skeleton.load instead."
     )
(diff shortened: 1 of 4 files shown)


webknossos/webknossos/skeleton/__init__.py
```diff
@@ -9,6 +9,7 @@
 
 
 def open_nml(file_path: Union[PathLike, str]) -> Skeleton:
+    """Deprecated."""
```
philippotto, line 12:
```suggestion
    """open_nml is deprecated, please use Skeleton.load instead."""
```

philippotto (approved):
Cool!

normanrz:
I was wondering if `MagView.get_bounding_boxes_on_disk()` and therefore `MagView.get_views_on_disk()` should be intersected with the layer's bounding box?

jstriebel:
Good question, I was wondering this as well^^ Not really sure tbh, but I think restricting this on the active bbox would be more correct. We might still add an argument to get all available bboxes later. Do you think we should warn, if there are wkw files outside the layer? I'll add a PR for this later.

philippotto:
> I think restricting this on the active bbox would be more correct.

I agree.

> We might still add an argument to get all available bboxes later.

If it's easy, I would provide a `respect_layer_bounding_box=True` parameter in the same PR, as there are certainly cases where setting it to False is desired.

> Do you think we should warn, if there are wkw files outside the layer?

Yes, if the intersection of a wkw file with the layers bounding box produces an empty bounding box, there should be a warning in my opinion, since this is a strong indicator of an invalid set layer bounding box (likely leading to ignoring volume data).

jstriebel:
> Yes, if the intersection of a wkw file with the layers bounding box produces an empty bounding box, there should be a warning in my opinion, since this is a strong indicator of an invalid set layer bounding box (likely leading to ignoring volume data).

Good point. I think we should check this then even in some more places, e.g. dataset upload, possilby removing wkw files when changing the layer-bbox, etc. Since this is a bit more I'll open an issue first, not sure if I can still squeeze it in today.

jstriebel:
added issue #629

PR #626, opened 2022-03-03
automate github release with changelog

### LINKED ISSUE #579: add content to gh & pypi releases
We should add some nice content to gh & pypi releases:
1. The github tags are pushed automagically via a gh action, also updating the Changelog (which is awesome!)
2. During the release CI, we should additionally
  * add a release on github, including the aggregated changelog entries per package
  * add some content to the pypi release, maybe also updating the main page with the newest README

jstriebel:
### Description:
- Automates the github release with the changelog
- Also added some small improvements for the pypi metadata

### Issues:
- fixes #579


philippotto (approved):
Awesome! Only tested `changelog_for_version.sh`, but that works well for me.

I only noticed that the "clustertools" changelog will appear at the top (if there are changes). Since the focus of this repository is on wklibs + wkcuber, I'd rather change the order. Not sure how to encode this counter-lexicographically maybe? Or hardcode the order and do an additional `for PKG in */pyproject.toml where PKG not in hardcoded_pkgs` to pickup potentially existing other packages? No strong opinion on this, though.

jstriebel:
> Since the focus of this repository is on wklibs + wkcuber, I'd rather change the order. 

Good point, I just set a fixed order. Since I could directly use it, I added a Makefile with the most important make commands and moved all the various scripts in the `_tooling` subfolder. Could you give 37159b1 another look, @philippotto? Thanks a lot :pray:

philippotto (approved):
:tada:

PR #631, opened 2022-03-04
Added method to validate dataset names

hotzenklotz:
### Description:
- During our webinar yesterday, I discovered that the webKnossos server blocks the upload of some datasets if the contain forbidden characters / ill-formed dataset names. 
- I added a simple name validation method to block the most common bad characters. I am not sure exactly what the backend check is using. This should hopefully catch most cases.

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -140,7 +140,7 @@ class Dataset:
             name = name or basename(normpath(dataset_path))
             assert self._is_valid_dataset_name(
                 name
-            ), "Invalid name. A dataset name can only contain letters, digits and underscores"
+            ), f"Invalid name {name}. A dataset name can only contain letters, digits and underscores."
 
             dataset_properties = DatasetProperties(
                 id={"name": name, "team": ""}, scale=scale, data_layers=[]
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -136,7 +136,12 @@ class Dataset:
             assert (
                 scale is not None
             ), "When creating a new dataset, the scale must be given, e.g. as Dataset(path, scale=(10, 10, 16.8))"
+
             name = name or basename(normpath(dataset_path))
+            assert self._is_valid_dataset_name(
+                name
+            ), "Invalid name. A dataset name can only contain letters, digits and underscores"
+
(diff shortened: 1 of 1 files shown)


webknossos/webknossos/dataset/dataset.py
```diff
@@ -136,7 +136,12 @@ def __init__(
             assert (
                 scale is not None
             ), "When creating a new dataset, the scale must be given, e.g. as Dataset(path, scale=(10, 10, 16.8))"
+
             name = name or basename(normpath(dataset_path))
+            assert self._is_valid_dataset_name(
+                name
+            ), "Invalid name. A dataset name can only contain letters, digits and underscores"
```
jstriebel, line 143:
```suggestion
            ), f"Invalid name {name}. A dataset name can only contain letters, digits and underscores."
```

jstriebel (changes requested):
> During our webinar yesterday, I discovered that the webKnossos server blocks the upload of some datasets if the contain forbidden characters / ill-formed dataset names.

Good idea checking this! However, I would only check this upon upload, which also allows to edit the name via an argument. I'm also not quite sure what exactly the patterns are, that are allowed. Please rather use a regex for checking for valid characters if it's that strict, and please ensure that the error message is descriptive (e.g. currently `-` is also allowed, I also used this successfully for uploads).

hotzenklotz:
>  I would only check this upon upload

I am not sure I agree. If we only check it on upload - and if it fails - one is stuck. There is no way to programmatically change a dataset name, is there?

jstriebel:
> I am not sure I agree. If we only check it on upload - and if it fails - one is stuck. There is no way to programmatically change a dataset name, is there?

Yes, you can set `dataset.name = "new_name"` or use `dataset.upload(new_dataset_name="new_name")`, as I wrote above.

jstriebel:
Btw, dataset-names are already validated upon uploaded, I assume this would catch those cases as well?
https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/client/_upload_dataset.py#L115-L122

hotzenklotz:
> Btw, dataset-names are already validated upon uploaded, I assume this would catch those cases as well?

That's exactly my issue. I need run any script/prepocess/downsampling etc and then fail on `.upload()` because the server rejects the dataset name. This is why I want to to check the name on creation.

PR #634, opened 2022-03-04
wk-libs docs: restructure index & add stability policy

jstriebel:
### Description:
* restructure index for wk-libs docs
* add stability policy

Screenshot from 2022-03-04 17-59-10


philippotto (approved):
:+1:

PR #638, opened 2022-03-08
Allow anisotropic mags when downsampling with --from

philippotto:
### Description:
- Allow anisotropic mags when downsampling with `--from`.

### Todos:
 - [x] Updated Changelog


hotzenklotz (approved):
Makes sense to me

PR #639, opened 2022-03-09
organization_id to organization_name

### LINKED ISSUE #558: Rename organization_name to organization_id

leowe:
### Description:
- rename all occurrences of `organization_name` to `organization_id` except in deprecated `webknossos.client.download_dataset`

### Issues:
- fixes #558 

### Todos:
 - [x] Updated Changelog


jstriebel (commented):
Nice stuff :+1: Just two things need to change IMO: The files in `webknossos/webknossos/client/_generated` should not be changed manually. I'm not sure if you changed the respective parts of the swagger schema as well to produce the changes, but I guess rather not? Then we should leave the parts there as they are, please see webknossos/client/README.md.
Also, please change `_WebknossosContext.organization` (in `context.py`) to `_WebknossosContext.organization_id` to use this more consistently internally.

jstriebel:
> Should I also rename `User.organization`?

Yes, good catch :+1: 

> We also have an inconsistency with `nml.param.organization` from the WkNML library

That's ok for the moment, since this directly mirrors the NML format which calls this parameter `organization`.

> Waiting for the PR for the `__generate_client.py` to run it.

see #656

leowe:
The only methods that are still using `organization_name` are the methods that were generated from wk swagger. Should we also change the usage in webknossos?

leowe:
I am a little unsure where the changes in the generated files come from since I used the newest webknossos master.

jstriebel (approved):
> The only methods that are still using `organization_name` are the methods that were generated from wk swagger. Should we also change the usage in webknossos?

I don't think so, at least for now, since that would change the REST interface and also need frontend changes.

> I am a little unsure where the changes in the generated files come from since I used the newest webknossos master.

I guess that depends on the linked dataset :see_no_evil: We'll need to use a fixed dataset at some point in the future, but the changes look fine :+1:

PR #641, opened 2022-03-09
Persist log file

### LINKED ISSUE #478: Persist log file
When running the cuber, there usually is a lot of output (overflowing terminal buffers). It would be great if all of that output would be persisted in a log file as well to allow to easily scan and search through it, afterwards.

leowe:
### Description:
- add logging to file
- to test: run a wkcuber workflow and check in the `./logs` directory

### Issues:
- fixes #478 

### Todos:
 - [x] Updated Changelog


philippotto (approved):
Awesome :+1:

PR #644, opened 2022-03-10
Use annotation-provided scale for temp volume annotation

normanrz:
### Description:
- Use the scale that is provided in the NML for the temp volume annotation in `Annotation.temporary_volume_layer_copy`

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/annotation/annotation.py
+++ b/webknossos/webknossos/annotation/annotation.py
@@ -7,7 +7,7 @@ from io import BytesIO
 from os import PathLike
 from pathlib import Path
 from tempfile import TemporaryDirectory
-from typing import (
+from typing import ((
     BinaryIO,
     ContextManager,
     Dict,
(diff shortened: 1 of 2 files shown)


philippotto (approved):

webknossos/webknossos/annotation/annotation.py
```diff
@@ -600,7 +600,7 @@ def temporary_volume_layer_copy(
             )
 
         input_annotation_dataset = Dataset(
-            str(tmp_annotation_dataset_path), scale=(1, 1, 1), exist_ok=True
+            str(tmp_annotation_dataset_path), scale=self.scale, exist_ok=True
```
jstriebel, line 603:
```suggestion
            tmp_annotation_dataset_path, scale=self.scale, name="tmp_annotation_dataset"
```

jstriebel (approved):
This is a nice improvement! Please add a short entry to the changelog as well.

Looking at it, I think the contextmanager does not work as intended here. The directory is removed before the dataset is created, so it is never cleaned up after the outer contextmanager is closed. Instead, I think the method should be
```python
        with TemporaryDirectory() as tmp_annotation_dir:
            input_annotation_dataset = Dataset(
                tmp_annotation_dir,
                name="tmp_annotation_dataset",
                scale=self.scale,
                exist_ok=True
            )

            input_annotation_layer = self.export_volume_layer_to_dataset(
                input_annotation_dataset,
                "volume_layer",
                volume_layer_name=volume_layer_name,
                volume_layer_id=volume_layer_id,
            )

            yield input_annotation_layer
```

Could you please adapt this part as well, since this already changes the relevant parts?

PR #645, opened 2022-03-10
More docs for Skeleton classes etc.

### LINKED ISSUE #582: improve the Skeleton documentation
Skeletons and the related classes have very few docstrings, this could need some more love :heart:

philippotto:
### Description:
- Adds more documentation to the skeleton classes.

### Issues:
- fixes #582

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/geometry/bounding_box.py
+++ b/webknossos/webknossos/geometry/bounding_box.py
@@ -12,6 +12,22 @@ from .vec3_int import Vec3Int, Vec3IntLike
 
 @attr.frozen
 class BoundingBox:
+    """
+    This classed is used to represent a cuboid in 3D.
+    The top-left coordinate is inclusive and the bottom-right coordinate is exclusive.
+
+    A small usage example:
+
+    ```python
+    from webknossos import BoundingBox
+
+    bbox_1 = BoundingBox((0, 0, 0), (100, 100, 100))
+    bbox_2 = BoundingBox((75, 75, 75), (100, 100, 100))
+
+    assert bbox_1.intersected_with(bbox_2).size == (25, 25, 25)
+    ```
+    """
+
     topleft: Vec3Int = attr.field(converter=Vec3Int)
     size: Vec3Int = attr.field(converter=Vec3Int)
     bottomright = attr.field(init=False)
@@ -110,7 +126,7 @@ class BoundingBox:
 
     @staticmethod
     def from_checkpoint_name(checkpoint_name: str) -> "BoundingBox":
-        """This function extracts a bounding box in the format x_y_z_sx_sy_xz which is contained in a string."""
+        """This function extracts a bounding box in the format `x_y_z_sx_sy_xz` which is contained in a string."""
         regex = r"(([0-9]+_){5}([0-9]+))"
         match = re.search(regex, checkpoint_name)
         assert (
--- a/webknossos/webknossos/geometry/vec3_int.py
+++ b/webknossos/webknossos/geometry/vec3_int.py
@@ -13,6 +13,23 @@ class Vec3Int(tuple):
         y: Optional[int] = None,
         z: Optional[int] = None,
     ) -> "Vec3Int":
+        """
+        Class to represent a 3D vector. Inherits from tuple and provides useful
+        methods and operations on top.
+
+        A small usage example:
+
+        ```python
+        from webknossos import Vec3Int
+
+        vector_1 = Vec3Int(1, 2, 3)
+        vector_2 = Vec3Int.full(1)
+        assert vector_2.x == vector_2.y == vector_2.y
+
+        assert vector_1 + vector_2 == (2, 3, 4)
+        ```
+        """
+
         if isinstance(vec, Vec3Int):
             return vec
 
--- a/webknossos/webknossos/skeleton/group.py
+++ b/webknossos/webknossos/skeleton/group.py
@@ -28,6 +28,22 @@ class Group:
     _skeleton: "Skeleton" = attr.ib(eq=False, repr=False)
     _enforced_id: Optional[int] = attr.ib(None, eq=False, repr=False)
 
+    @classmethod
+    def _set_init_docstring(cls) -> None:
+        Group.__init__.__doc__ = """
+        To create a group, it is recommended to use `webknossos.skeleton.skeleton.Skeleton.add_group` or
+        `Group.add_group`. That way, the newly created group
+        is automatically attached as a child to the object the method was
+        called on.
+
+        A small usage example:
+
+        ```python
+        subgroup = group.add_group("a subgroup")
+        graph = subgroup.add_graph("a graph")
+        ```
+        """
+
     def __attrs_post_init__(self) -> None:
         if self._enforced_id is not None:
             self._id = self._enforced_id
@@ -45,6 +61,8 @@ class Group:
         color: Optional[Union[Vector4, Vector3]] = None,
         _enforced_id: Optional[int] = None,
     ) -> Graph:
+        """Adds a graph to the current group with the provided name (and color if specified)."""
+
         if color is not None and len(color) == 3:
             color = cast(Optional[Vector4], color + (1.0,))
         color = cast(Optional[Vector4], color)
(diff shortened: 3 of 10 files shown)


webknossos/webknossos/geometry/vec3_int.py
```diff
@@ -44,6 +62,11 @@ def from_xyz(x: int, y: int, z: int) -> "Vec3Int":
         # the tolerant (and potentially) slow Vec3Int.__new__ method.
         return tuple.__new__(Vec3Int, (x, y, z))
 
+    @staticmethod
+    def from_vec3_float(vec: Tuple[float, float, float]) -> "Vec3Int":
+
+        return Vec3Int(floor(vec[0]), floor(vec[1]), floor(vec[2]))
```
jstriebel, line 68: I'm not quite sure why this should be floored, I'd either expect `int()` (which only differs on negative inputs from floor, which probably won't be the case), or `round()`.

webknossos/webknossos/skeleton/group.py
```diff
@@ -61,64 +79,87 @@ def add_graph(
...
+    def graphs(self) -> Iterator[GroupOrGraph]:
+        """Returns all (immediate) graph children as an iterator.
+        Use flattened_graphs if you need also need graphs within subgroups."""
+        return (child for child in self._children if isinstance(child, Graph))
+
+    @property
+    def groups(self) -> Iterator[GroupOrGraph]:
+        """Returns all (immediate) group children as an iterator.
+        Use flattened_groups if you need also need groups within subgroups."""
+        return (child for child in self._children if isinstance(child, Group))
```
jstriebel, line 95: I think those properties are great additions. However, I'd rather store the graphs and groups separate from each other and simply combine them in `children`, which saves all the `isinstance` checks.

webknossos/webknossos/skeleton/group.py
```diff
@@ -61,64 +79,87 @@ def add_graph(
...
                 yield from child.flattened_graphs()
             else:
                 yield child
 
     def flattened_groups(self) -> Generator["Group", None, None]:
+        """Returns an iterator of all groups within this group (and its subgroups)."""
         for child in self._children:
             if isinstance(child, Group):
                 yield child
                 yield from child.flattened_groups()
```
jstriebel, line 135: Please refactor those to use `self.groups` in `flattened_groups` directly and `self.graphs` in `flattened_graphs`. E.g.
```suggestion
    def flattened_graphs(self) -> Iterator[Graph]:
        """Returns an iterator of all graphs within this group (and its subgroups)."""
        yield from self.graphs
        for group in self.groups:
            yield from group.flattened_graphs()

    def flattened_groups(self) -> Iterator["Group"]:
        """Returns an iterator of all groups within this group (and its subgroups)."""
        for group in self.groups:
            yield group
            yield from group.flattened_groups()
```

webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -12,6 +12,22 @@
 
 @attr.frozen
 class BoundingBox:
+    """
+    This class is used to represent a cuboid in 3D.
```
jstriebel, line 16:
```suggestion
    This class is used to represent an axis-aligned cuboid in 3D.
```

jstriebel (commented):
Awesome stuff! It might not be ready for review yet, but I couldn't hold myself and already had a look :cat2:

webknossos/webknossos/geometry/vec3_int.py
philippotto: You are right. I changed it to `int'.

webknossos/webknossos/skeleton/group.py
philippotto: good point :+1: simplifies the code, too.

philippotto:
> Awesome stuff! It might not be ready for review yet, but I couldn't hold myself and already had a look :cat2:

Thanks for the review! The PR was review-ready by the way :smile: Feel free to have another look :)

webknossos/webknossos/skeleton/group.py
```diff
@@ -20,10 +20,15 @@
 class Group:
     _id: int = attr.ib(init=False)
     name: str
-    _children: Set[GroupOrGraph] = attr.ib(
+    _child_groups: Set["Group"] = attr.ib(
         factory=set,
         init=False,
-        repr=lambda children: f"<{unit_len(children, 'children')}>",
+        repr=lambda children: f"<{unit_len(children, 'child groups')}>",
```
jstriebel, line 26:
```python
>>> unit_len([1, 2], 'child groups')
'2 child groupses'
```
:rofl:

```suggestion
        repr=lambda children: f"<{unit_len(children, 'child group')}>",
```

webknossos/webknossos/skeleton/group.py
```diff
@@ -20,10 +20,15 @@
...
+    )
+    _child_graphs: Set[Graph] = attr.ib(
+        factory=set,
+        init=False,
+        repr=lambda children: f"<{unit_len(children, 'child graphs')}>",
```
jstriebel, line 31:
```suggestion
        repr=lambda children: f"<{unit_len(children, 'child graph')}>",
```

webknossos/examples/calculate_segment_sizes.py
```diff
@@ -0,0 +1,33 @@
+import numpy as np
```
jstriebel, line 1: Please also add a respective file in `docs/src/webknossos-py/examples/` for this example and include it in the index in `docs/mkdocs.yml`.

jstriebel (approved):
Thanks a ton, awesome stuff! :heart: I just noticed that the new example probably also should show up in the docs, but apart from that and the two nitpicks everything LGTM :+1:

webknossos/webknossos/skeleton/group.py
philippotto: :sweat_smile: 

I thought, one would need to provide the plural form. However, I now see, that this doesn't make much sense. However, I just wanted to say that `unit_len` also isn't perfect (`unit_len(["one"], 'children')` yields `1 children` even though I would have expected `1 child`).

jstriebel: > However, I just wanted to say that `unit_len` also isn't perfect (`unit_len(["one"], 'children')` yields `1 children` even though I would have expected `1 child`).

It is, if you use the singular form here as well:
```python
>>> unit_len([1, 2], 'child')
'2 children'
>>> unit_len([1], 'child')
'1 child'
```

philippotto: Ah ok, then it has been wrong before this PR ;)

webknossos/webknossos/skeleton/skeleton.py
```diff
@@ -47,12 +47,6 @@ class Skeleton(Group):
     _enforced_id: Optional[int] = attr.field(default=None, eq=False, repr=False)
     name: str = attr.field(default="Root", init=False, eq=False, repr=False)
 
-    """Should not be used with `Skeleton`, this attribute is only useful for sub-groups. Set to `Root`."""
```
philippotto, line 35: @jstriebel Do you know why the `_children` property was here redefined in the first place? I removed it now as it's redundant to `Group`, but it might exist for a reason? Maybe documenting that it should not be used for the `Skeleton` class? 

The comment `"""Should not be used with `Skeleton`, this attribute is only useful for sub-groups. Set to `Root`."""` probably belonged to `name`. Readding it now.

jstriebel: > @jstriebel Do you know why the `_children` property was here redefined in the first place? I removed it now as it's redundant to `Group`, but it might exist for a reason? Maybe documenting that it should not be used for the `Skeleton` class?

I'm guessing that it might have been to define a certain order for the repr, but not quite sure :shrug: 

> The comment `"""Should not be used with `Skeleton`, this attribute is only useful for sub-groups. Set to `Root`."""` probably belonged to `name`. Readding it now.

Yes, the attribute-docstrings unfortunately are under their attribute :+1:

docs/mkdocs.yml
```diff
@@ -77,6 +77,7 @@ nav:
       - webknossos-py/examples/apply_merger_mode.md
       - webknossos-py/examples/learned_segmenter.md
       - webknossos-py/examples/skeleton_synapse_candidates.md
+      - webknossos-py/examples/calculate_segment_sizes.md
```
jstriebel, line 80: We could group those as `Annotation Examples` or sth similar, but also fine as-is for now.

PR #646, opened 2022-03-11
First class bounding boxes

### LINKED ISSUE #482: Support reading name of user bounding boxes with skeleton API
currently, `wk.Skeleton.load(...).user_bounding_boxes` only contains the boundaries, but user bounding boxes also have a name which should be exposed.

philippotto:
### Description:
- Exposes proper bounding objects when interacting with an annotation.
- Adds the properties `id`, `name`, `is_visible` and `color` to the existing bounding box class so that all attributes of an NML-encoded bounding box can be represented.

### Issues:
- fixes #482

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -14,7 +14,11 @@
 class BoundingBox:
     topleft: Vec3Int = attr.field(converter=Vec3Int)
     size: Vec3Int = attr.field(converter=Vec3Int)
-    bottomright = attr.field(init=False)
+    bottomright: Vec3Int = attr.field(init=False)
+    name: Optional[str] = attr.field(default="Unnamed Bounding Box")
+    is_visible: bool = attr.field(default=True)
+    id: Optional[str] = attr.field(default=None)
+    color: Optional[Tuple[float, float, float, float]] = attr.field(default=None)
```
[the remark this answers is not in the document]
jstriebel: PS: Defaults can (and IMO should) be given using the standard syntax:
```suggestion
    name: Optional[str] = "Unnamed Bounding Box"
    is_visible: bool = True
    id: Optional[str] = None
    color: Optional[Tuple[float, float, float, float]] = None
```

webknossos/Changelog.md
```diff
@@ -13,8 +13,10 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 [Commits](https://github.com/scalableminds/webknossos-libs/compare/v0.9.9...HEAD)
 
 ### Breaking Changes
+- The annotation class now exposes bounding objects instead of tuples. [#646](https://github.com/scalableminds/webknossos-libs/pull/646)
 
 ### Added
+- The bounding box class now supports the following additional properties: `id`, `name`, `is_visible` and `color'. [#646](https://github.com/scalableminds/webknossos-libs/pull/646)
```
jstriebel, line 19:
```suggestion
- The `BoundingBox` class now supports the following additional properties: `id`, `name`, `is_visible` and `color'. [#646](https://github.com/scalableminds/webknossos-libs/pull/646)
```

webknossos/webknossos/_nml/parameters.py
```diff
@@ -3,7 +3,11 @@
 
 from loxun import XmlWriter
 
-from .utils import IntVector6, Vector3, enforce_not_null, filter_none_values
+from webknossos.geometry import BoundingBox
+
+from .utils import Vector3, enforce_not_null, filter_none_values
+
+DEFAULT_COLOR = [0.2, 0.5, 0.1, 1]
```
jstriebel, line 10: Where does this come from? Maybe rather `DEFAULT_BOUNDING_BOX_COLOR`?

webknossos/webknossos/_nml/parameters.py
```diff
@@ -112,29 +126,43 @@ def _dump(self, xf: XmlWriter) -> None:
...
             int(bounding_box_element.get("width", 0)),
             int(bounding_box_element.get("height", 0)),
             int(bounding_box_element.get("depth", 0)),
         )
+        color = (
+            float(bounding_box_element.get("color.r", 0)),
+            float(bounding_box_element.get("color.g", 0)),
+            float(bounding_box_element.get("color.b", 0)),
+            float(bounding_box_element.get("color.a", 0)),
+        )
```
jstriebel, line 146: I think the default should rather be `None` for the color, similar how the parsing is done for `Tree`.

jstriebel (commented):
Nice work :tada: I added some minor comments, and was wondering if the additional attributes for the `BoundingBox` should rather be part of a separate subclass, but not sure here. Please see the comments:

webknossos/webknossos/geometry/bounding_box.py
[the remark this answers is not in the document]
philippotto: > Do you think it makes sense to have those attributes in all cases (even with their default values)? Having just a single class as you propose might make it easier for users, since there's no distinction between the two.

Yes, I think having only one BB class is way easier to use (if the parameters are not used, they don't hurt, either). The only downside I see would be performance, but the number of bounding boxes is usually **very** low in comparison to the actual EM data. So, this argument isn't really strong for me. Do you have other points against a unified class?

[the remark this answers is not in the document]
jstriebel: Totally agree :+1: The only point I could imagine is that we use the data-structure for some algorithm, similar to Vec3Int in the flood-fill, and run into performance problems then. But let's solve this as it appears, I think we can still add this abstraction for exactly that purpose then (it wouldn't even need to be user-facing).

webknossos/webknossos/_nml/parameters.py
philippotto: I renamed it now. I just chose some arbitrary color. There is no specific reason for it being the way it is.

philippotto:
@jstriebel Thanks for your feedback! Please have another look :)

jstriebel (approved):
:+1: :rocket: :metal:

PR #655, opened 2022-03-15
bugfix: Missing default values for optional parameters in _get_volume_layer

erjel:
### Description:
Missing default parameters for method `_get_volume_layer`.

Currently, the method can not be used without parameters.
So either delete the `Optional` or add default parameters.

Otherwise the first `if` [statement](https://github.com/erjel/webknossos-libs/blob/a5550c1833e6b11f7ca5993f08925b98269c415d/webknossos/webknossos/annotation/annotation.py#L461
) is only a special case of the ones that follow.

### DIFF
--- a/webknossos/webknossos/annotation/annotation.py
+++ b/webknossos/webknossos/annotation/annotation.py
@@ -453,8 +453,8 @@ class Annotation:
 
     def _get_volume_layer(
         self,
-        volume_layer_name: Optional[str],
-        volume_layer_id: Optional[int],
+        volume_layer_name: Optional[str] = None,
+        volume_layer_id: Optional[int] = None,
     ) -> _VolumeLayer:
         assert len(self._volume_layers) > 0, "No volume annotations present."
 


jstriebel (approved):
Hi @erjel, thanks for the contribution!

> Currently, the method can not be used without parameters. So either delete the `Optional` or add default parameters.

That's because the method is only meant to be used internally at the moment as it's prefixed with an underscore, please also see our [stability policy, point 2](https://docs.webknossos.org/webknossos-py/stability_policy.html). 

> Otherwise the first `if` [statement](https://github.com/erjel/webknossos-libs/blob/a5550c1833e6b11f7ca5993f08925b98269c415d/webknossos/webknossos/annotation/annotation.py#L461) is only a special case of the ones that follow.

You might still supply `None` explicitly (e.g. as the two calling methods `delete_volume_layer` & `export_volume_layer_to_dataset` do), but you're right that this is not a great usability (mostly because it's not  intended to be used from users).

Since this method is not meant to be used directly, I think it's most important to support your actual use-case. Could you please elaborate what are you using the result for? E.g. do you want to access the fallback-layer to merge the annotation data? I'd be happy to prioritize supporting your use-case accordingly.

In the meantime merging this change does no harm and we'll add it, but please be aware that this method might change at any time even with stable releases.

erjel:
Hi @jstriebel ,

>   Could you please elaborate what are you using the result for? E.g. do you want to access the fallback-layer to merge the annotation data?

You guessed right, I used the function to implement a fallback layer merging. Currently, the [gist](https://gist.github.com/philippotto/fcba429c646d34733d045a14b390a17a) provided as solution for the corresponding [image.sc post](https://forum.image.sc/t/webknossos-web-ui-annotations-and-annotation-downloads-are-inconsistent/62962) requires the fallback `segmentation_layer_path` as file system path.

Given that the `webknossos_folder` in the file system is known, the `segmentation_layer_path` is a combination of the `annotation.dataset_name` and the corresponding `fallback_layer_name` of the corresponding volume layer. In the case of a single volume layer in the annotation I can write:
```python
fallback_layer_name = annotation._get_volume_layer(None, None).fallback_layer_name
segmentation_layer_path= webknossos_folder / annotation.dataset_name / fallback_layer_name
```

Currently, I do not see another option to get the `segmentation_layer_path` via the python API from the annotation directly. Yet, this is required to download the "full" annotation as displayed in the web UI.

 > please be aware that this method might change at any time even with stable releases

I already pin your libraries up to the patch number. I just can not keep up with your release pace ...

Eric

PR #656, opened 2022-03-16
cleanup generate_client, use local wk-test setup

jstriebel:
### Description:
- update wk version for the test setup
- regenerate test snapshots with the upgraded version
-  use wk-test setup for `generate_client.sh` (which wraps `__generate_client.py`)
- adds some ignored fields coming from the new wk version

### Issues:
- part of #514, more cleanup coming separately


webknossos/tests/binaryData/Organization_X/e2006_knossos/datasource-properties.json
```diff
@@ -0,0 +1,52 @@
+{
```
jstriebel, line 1: @fm3 Is it fair to assume that this dataset exists locally? It is a [part of the test db](https://github.com/scalableminds/webknossos/blob/master/test/db/dataSets.csv#L7), that's why I chose this name. It's needed for the annotation upload (and I think also the dataset-specific requests). Here, I just copied the small `testdata/simple_wk_dataset` to have a valid dataset.

webknossos/__generate_client.py
```diff
@@ -16,10 +15,9 @@
     _get_project_for_url_or_path,
 )
 
-from webknossos.utils import snake_to_camel_case
-
-# SCHEMA_URL = "https://master.webknossos.xyz/swagger.json"
-SCHEMA_URL = "http://localhost:9000/swagger.json"
+WK_URL = os.environ["WK_URL"]
```
jstriebel, line 18: Should we warn if this is not `localhost:9000`?

webknossos/tests/binaryData/Organization_X/e2006_knossos/datasource-properties.json
jstriebel: Hm, l4_sample is not visible for the test user `user_A@scalableminds.com`, I'm getting
```
AssertionError: Failed to upload annotation: 404: {"messages":[{"error":"Couldn't find the requested resource."}]}
```

 @fm3 Could you help me to adapt the test db that the dataset is visible?

webknossos/__generate_client.py
```diff
@@ -225,6 +268,7 @@ def make_properties_required(x: Any) -> None:
                 property
                 for property in properties.keys()
                 if property not in OPTIONAL_FIELDS
+                and not ("tracingId" in properties and property == "name")
```
jstriebel, line 271: Follow-up issue: use path or parent instead of sibling

fm3 (approved):
Great stuff, I think this will save us a lot of headaches :) as discussed in person, please check the type of tracingTime and how to make stuff optional (not only by name)

Issue #580, opened 2022-02-03
[wkcuber] cleanup testing


jstriebel:
Currently the test setup consists of different scripts as well as pytest tests. The following commands are run in CI:
```bash
tar -xzvf testdata/WT1_wkw.tar.gz
poetry run pytest tests
poetry run tests/scripts/all_tests.sh
```

There's also a `test.sh` which is outdated atm. We should ensure that this script works, is used in CI and reflects all tests needed and expected. Additionally, we might double-check some of the scripts, they might also be refactored to pytest tests.

Related: #409

jstriebel:
PS: http://docs.webknossos.org/webknossos-py/development.html#wkcuber-package should be updated accordingly

philippotto:
> Additionally, we might double-check some of the scripts, they might also be refactored to pytest tests.

Yes, I think, this would make a lot of sense. Could be done peu à peu. Also see
https://github.com/scalableminds/webknossos-libs/pull/657/files#r829123141

PR #657, opened 2022-03-17
Fix downsampling/compression of segmentation layers + misc

philippotto:
### Description:
- fix downsampling and compressing segmentation layers (cast largest_segment_id to int if necessary and safe; also pass when creating copy of layer)
- adds test for cubing/downsampling/compressing (end-to-end) for a color and segmentation layer
- improve logging (e.g., point out deprecated caller location) and remove some deprecation warnings
- upgrade zarr in wkcuber to be in sync with webknossos package
- adds/cleans up ./lint.sh, ./typecheck.sh and ./format.sh for wkcuber

### Issues:
- follow-up for https://github.com/scalableminds/webknossos-libs/commit/04938dea411be7f61c408c351f56c0bdf1e3b1d5

### Todos:
 - [x] Updated Changelog


wkcuber/tests/test_main.py
```diff
@@ -0,0 +1,65 @@
...
+)
+from webknossos import Dataset
+from wkcuber.__main__ import create_parser, cube_with_args
+from tifffile import TiffWriter
+
+TESTOUTPUT_DIR = Path("testoutput")
+
+
+@pytest.mark.parametrize("category", ["color", "segmentation"])
+def test_main(category: str) -> None:
```
normanrz, line 15: Nice. I'm starting to think that we should port your bash test scripts also to pytest. Asserting with the Dataset API makes a lot of sense.

normanrz (approved):
Thanks for cleaning this up 🧹

wkcuber/tests/test_main.py
philippotto: Yes, I agree. I'll highlight this point in https://github.com/scalableminds/webknossos-libs/issues/580, too.

PR #658, opened 2022-03-17
fix changelog entries

jstriebel:
Fixes some changelog entries, probably some leftovers from a merge conflict.


philippotto (approved):
Oops, thanks for cleaning up this mess!

PR #660, opened 2022-03-18
dataset download: allow url & sharing token, add tests

### LINKED ISSUE #559: Download a dataset from the dataset URL
To download a dataset, the user needs to know the organization_name (or _id #558) and dataset name. As an alternative, the download function could parse a user-provided URL (eg. https://webknossos.org/datasets/Demo_Organization/h01_20210601/view#620659,317098,2662,0,91.819) for these bits of information.

### LINKED ISSUE #599: Dataset: test upload-download roundtrip
We discovered some problems with dataset download (see #595), this would have been detected with a simple roundtrip test (using multiple mags, non 0, non-32-aligned bbox). Let's add this.

jstriebel:
### Description:
- Allows to download dataset by url
- Allows to pass a sharing token for dataset url (either by argument or via the url)
- Adds download tests, also upload-download roundtrip

### Issues:
- fixes #559 
- fixes #599

### Todos:
 - [x] Updated Changelog


jstriebel:
@philippotto I already added automerge, feel free to remove it if necessary.

webknossos/examples/download_image_data.py
```diff
@@ -15,7 +15,7 @@ def main() -> None:
 
     # Download the datset
     ds = wk.Dataset.download(
-        dataset_name=name,
+        dataset_name_or_url=name,
```
philippotto, line 18: Do we consider a change of a parameter name to be a breaking change?

philippotto (approved):
Cool! Only see my one question.

webknossos/examples/download_image_data.py
jstriebel: Good point, yes :+1: I'll add this to the changelog.

PR #662, opened 2022-03-20
non-fork start_method as default

### LINKED ISSUE #661: Always use `spawn` in multiprocessing
I just debugged an issue where my multiprocessing worker would hang on Linux, but not on macOS. After 4 hours of debugging, I found out that `s3fs` seems to have issues with the `fork` start method in multiprocessing. See https://github.com/fsspec/s3fs/issues/464.

I know there have been other occasions where switching the start method was the solution. I wonder if there are any downsides to always use `spawn` (or `forkserver` where available) within the ProcessPoolExecutors from `cluster_tools`?

normanrz:
### Description:
- Uses `forkserver` or `spawn` start methods for multiprocessing by default. `fork` is known to create hard-to-debug bugs.

### Issues:
- fixes #661 

### Todos:
 - [x] Updated Changelog


normanrz:
@philippotto is there any way I can test this PR with the Voxelytics CI?

philippotto:
> @philippotto is there any way I can test this PR with the Voxelytics CI?

I guess you would need to publish a beta or pre-release package of clustertools to pypi and then use that in a branch of VX so that the CI can run it? Haven't done such a thing myself yet, though.

normanrz:
Ok. I'll try that.

Other than that, do you think this PR is a good idea?

philippotto:
> Other than that, do you think this PR is a good idea?

Yes, I think, it makes sense. The only downside might be performance. I think, we've seen a performance hit in the past (especially for CI) when switching away from forking. However, stability (and keeping sanity) is more important than performance in my opinion. In case, it matters greatly for CI, we could think about sticking to `fork` in the CI context. However, I'd only do this when really needed.

normanrz:
What do think about `forkserver` vs `spawn` as default?
I found this resource that explains both all methods quite well: http://www.bnikolic.co.uk/blog/python/parallelism/2019/11/13/python-forkserver-preload.html

The pros/cons:
`forkserver`: fast; but needs network, which can be blocked through firewalls or in our CI
`spawn`: slow; but most reliable

normanrz:
Seems to work in vx: https://github.com/scalableminds/voxelytics/actions/runs/2025032655

philippotto:
> forkserver: fast; but needs network, which can be blocked through firewalls or in our CI

This was already problematic for this CI, right? Since you added `export MULTIPROCESSING_DEFAULT_START_METHOD=spawn`. If the default of forkserver works for vx, I'd keep it as is now (defaulting to forkserver and allowing to switch to spawn easily). If it turns out to be annoying in the future, we can still change the default to the safer `spawn` variant.

normanrz:
> This was already problematic for this CI, right? Since you added `export MULTIPROCESSING_DEFAULT_START_METHOD=spawn`. If the default of forkserver works for vx, I'd keep it as is now (defaulting to forkserver and allowing to switch to spawn easily). If it turns out to be annoying in the future, we can still change the default to the safer `spawn` variant.

I'm less concerned with our stack (where we can easily override the default), but with the setups of external users (e.g. on their PCs with a firewall).

philippotto:
> > This was already problematic for this CI, right? Since you added `export MULTIPROCESSING_DEFAULT_START_METHOD=spawn`. If the default of forkserver works for vx, I'd keep it as is now (defaulting to forkserver and allowing to switch to spawn easily). If it turns out to be annoying in the future, we can still change the default to the safer `spawn` variant.
> 
> I'm less concerned with our stack (where we can easily override the default), but with the setups of external users (e.g. on their PCs with a firewall). Not that we replace one source of hard-to-find bugs with another :trollface:

Ok, makes sense. Then let's default to spawn and change it to forkserver when necessary/compatible (e.g., in vx), right?

jstriebel:
@normanrz We already had this (or a similar) problem on MacOS, and fixed it here:
https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/tests/conftest.py#L321-L328

It might be enough to adapt the check, at least if the communication uses unix sockets. If not, you might need to adapt the `allowed_hosts` patterns. Another alternative might be to remove `block-network` [in the `test.sh`](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/test.sh#L20), but I'd prefer the first solution if this works out.

webknossos/tests/conftest.py
```diff
@@ -318,10 +317,7 @@ def pytest_collection_modifyitems(items: List[pytest.Item]) -> None:
...
             item.add_marker("vcr")
 
-        if (
-            item.get_closest_marker("block_network") is None
-            and multiprocessing.get_start_method() != "fork"
-        ):
+        if item.get_closest_marker("block_network") is None:
             # To allow for UNIX socket communication necessary for spawn multiprocessing
             # addresses starting with `/` are allowed
             marker = pytest.mark.block_network(allowed_hosts=["/.*"])
```
jstriebel, line 323: I think you might also add `"/.*" to the list of allowed hosts, if the marker exist already, e.g.
```python
marker = item.get_closest_marker("block_network")
if marker is None:
    marker = pytest.mark.block_network(allowed_hosts=["/.*"])
    item.add_marker(marker)
else:
    marker.kwargs["allowed_hosts"].append("/.*")
```

Then you can omit the additional `"/.*"` directly at the tests.

jstriebel (approved):
Nice! :+1: Mostly LGTM, see me one comment below:

PR #666, opened 2022-03-23
unify tests in test.sh

leowe:
### Description:
- unify test call in `./test.sh`
- remove all other scripts that might call collections of test scripts, to enfore usage of `./test.sh`
- usage of `./test.sh` in CI

### Issues:
- contributes to #580


leowe:
It seems like somehow the CI isn't running through although all tests are green...

jstriebel:
> It seems like somehow the CI isn't running through although all tests are green...

That's because a run was cancelled, probably manually:
Screenshot from 2022-03-28 11-58-59

> @jstriebel I am wondering whether that closes #580 for now or whether I should also look into substituting some of the bash scripts with python scripts?

I'd leave it open for the shell-script-to-pytest conversion. It would be great if you could look into this as well, but it can be in another PR :+1:

jstriebel (approved):
LGTM :+1:

PR #675, opened 2022-03-30
update generated client

jstriebel:
### Description:
- Updates the generated client. The diff mostly stems from slightly different datasets.

### DIFF
--- a/webknossos/webknossos/client/_generated/models/dataset_info_response_200_data_source.py
+++ b/webknossos/webknossos/client/_generated/models/dataset_info_response_200_data_source.py
@@ -18,7 +18,7 @@ class DatasetInfoResponse200DataSource:
 
     id: DatasetInfoResponse200DataSourceId
     data_layers: List[DatasetInfoResponse200DataSourceDataLayersItem]
-    scale: List[float]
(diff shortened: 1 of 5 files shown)


philippotto (approved):

webknossos/webknossos/client/_generated/models/dataset_info_response_200_data_source.py
```diff
@@ -18,7 +18,7 @@ class DatasetInfoResponse200DataSource:
 
     id: DatasetInfoResponse200DataSourceId
     data_layers: List[DatasetInfoResponse200DataSourceDataLayersItem]
-    scale: List[float]
+    scale: List[int]
```
philippotto, line 21: Just stumbled upon this: Aren't there scales such as `11.24, 11.24, 28` which would require floats?

jstriebel: Yes. Unfortunately, the checked in dataset simply has integers there, we probably should change that. Luckily, this is only used for type-checking and not a problem there, a list of floats is simply passed on at runtime.

PR #677, opened 2022-03-31
Improve default_view_configuration handling

### LINKED ISSUE #604: Move layer-wise default_view_configuration parts into Layer
Currently, the `default_view_configuration` property-attribute on the `Dataset` class can also contain layer-wise configurations. We should move the layer-wise portions into the `Layer` class and handle this nicely for users:

- [X] The Dataset's `default_view_configuration` should just have the global settings (zoom/position/interpolation), possibly typed. The layer-wise configurations should instead be properties on the layers, also possibly typed.
- [x] We should adapt the `upload_image_data` example to set a good default config for the nuclei layer, e.g. `{"color": [255, 0, 0], "intensityRange": [3000, 30000]}`
- [x] Removing a layer, and copying / linking a layer from another dataset should also handle the default-view config, e.g. copying the configuration to the new dataset when copying a layer.
- [x] idempotency for uploading + downloading a dataset with view configurations

philippotto:
### Description:
- Makes use of `default_view_configuration` in `upload_image_data` example.
- Ensures that default_view_configuration is saved correctly when downloading dataset.

### Issues:
- fixes #604 (most of it already worked)

### Todos:
 - [x] Updated Changelog


jstriebel (commented):
Nice, almost LGTM! Just two points:
* Should we also handle the global default view configuration on download? The new code is just for the layer-wise, right?
* Could you also please add this to the roundtrip test (`test_dataset_download_upload.py`)?

jstriebel:
> > * Should we also handle the global default view configuration on download? The new code is just for the layer-wise, right?
> 
> I think, the global default view configuration is not exposed in the info request at the moment. See `https://webknossos.org/api/datasets/scalable_minds/l4_sample_dev`. Instead, it's in `https://webknossos.org/api/dataSetConfigurations/scalable_minds/l4_sample_dev`. Therefore, I'd postpone this to a later PR.

Sounds fair! Let's tackle this again after https://github.com/scalableminds/webknossos/issues/6133.

jstriebel (approved):
LGTM :+1:

Issue #682, opened 2022-04-04
cluster_tools: add test.sh


jstriebel:
PR #683 added an empty test.sh, which should contain all tests from CI, and be used in CI instead.

Issue #681, opened 2022-04-04
cluster_tools: type everything


jstriebel:
After PR #683, initial typing is introduced to cluster_tools. However, we should enforce complete typing by uncommenting `# disallow_untyped_defs = true` in the pyproject.toml and adding the respective type-hints.

PR #683, opened 2022-04-04
add global make commands, add cluster-tools typechecking

### LINKED ISSUE #581: Use common scripts across packages
In my opinion it would be great to use the following scripts consistently across projects:
    * `format.sh`: black and isort
    * `lint.sh`: pylint
    * `typecheck.sh`: mypy
    * `test.sh`: pytest and custom scripts (see also #580 for wkcuber)

After they are introduced everywhere, we might also add some top-level helpers that run across all (changed) repos, possibly also for common poetry commands (`install`/`update`). `make` might be helpful.

jstriebel:
### Description:
- adds `make` commands for format, test, lint, typecheck across all packages
- cluster-tools:
  - adds initial `typecheck.sh` command (added #681 for full typechecking)
  - adds a `test.sh` stub doing nothing, added #682 for the real thing
- wkcuber: upgraded mypy, fixed some related issues

### Issues:
- fixes #581

### DIFF
--- a/cluster_tools/cluster_tools/pickling.py
+++ b/cluster_tools/cluster_tools/pickling.py
@@ -1,16 +1,7 @@
 import os
+import pickle
 import sys
 
-use_cloudpickle = "USE_CLOUDPICKLE" in os.environ
-
-if use_cloudpickle:
-    import cloudpickle
-
-    pickle_strategy = cloudpickle
-else:
-    import pickle
-
-    pickle_strategy = pickle
 from .util import warn_after
 
 WARNING_TIMEOUT = 10 * 60  # seconds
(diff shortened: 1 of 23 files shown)


cluster_tools/cluster_tools/pickling.py
```diff
@@ -45,27 +36,23 @@ def get_suitable_pickle_protocol():
...
 def loads(*args, **kwargs):
     assert (
         "custom_main_path" not in kwargs
     ), "loads does not implement support for the argument custom_main_path"
-    return pickle_strategy.loads(*args, **kwargs)
+    return pickle.loads(*args, **kwargs)
 
 
-class RenameUnpickler(pickle_strategy.Unpickler):
+class RenameUnpickler(pickle.Unpickler):
```
jstriebel, line 55: Since cloudpickle does not expose `cloudpickle.Unpickler`, the support for cloudpickle was broken anyways (probably since 3 years, see [this PR](https://github.com/scalableminds/cluster_tools/pull/68/files#diff-263c407a6e6c37c0e1033d2c2022c66c6036de62eb52480b549059327822e794R58). I simply removed it now.

normanrz (approved):
Can't comment on cloudpickle. Should it also be removed from the dep list?
The rest looks good!

cluster_tools/cluster_tools/pickling.py
philippotto: Yes, that's probably a good call. Since cloudpickle caused some trouble back then, we switched back to the default pickling and haven't really had any need to switch back. So, let's remove it altogether :+1:

PR #686, opened 2022-04-05
Also store success boolean to pickle output when using multiprocessing

philippotto:
### Description:
- When using the clustertools with slurm, the result pickle file was always written in the format of `(wasSuccessful, result_value)`. However, when using multiprocessing, this was not the case which was inconsistent.

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/cluster_tools/cluster_tools/__init__.py
+++ b/cluster_tools/cluster_tools/__init__.py
@@ -176,7 +176,6 @@ class WrappedProcessPoolExecutor(ProcessPoolExecutor):
         else:
             raise result[1]
 
-
     def map_unordered(self, func, args):
 
         futs = self.map_to_futures(func, args)
--- a/cluster_tools/cluster_tools/__init__.py
+++ b/cluster_tools/cluster_tools/__init__.py
@@ -1,3 +1,4 @@
+import logging
 import multiprocessing
 import os
 import tempfile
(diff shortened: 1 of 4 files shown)


cluster_tools/cluster_tools/remote.py
```diff
@@ -78,10 +78,10 @@ def worker(executor, workerid, job_array_index, job_array_index_offset, cfut_dir
         out = pickling.dumps(result)
 
     except Exception:
-        print(traceback.format_exc())
 
         result = False, format_remote_exc()
-        logging.warning("Job computation failed.")
+        logging.warning("Job computation failed with:")
+        print(traceback.format_exc())
```
jstriebel, line 84: I'm wondering why this is partially in logging.warning and in print? Maybe we should rather put the traceback also in the logging statement?

cluster_tools/cluster_tools/__init__.py
```diff
@@ -162,7 +162,12 @@ def _execute_and_persist_function(output_pickle_path, *args, **kwargs):
         func = args[0]
         args = args[1:]
 
-        result = func(*args, **kwargs)
+        try:
+            result = True, func(*args, **kwargs)
+        except Exception as exc:
+            result = False, exc
+            logging.warning("Job computation failed with:")
+            print(exc)
```
jstriebel, line 170: Same here, maybe also in other spots?

jstriebel (approved):
Nice, mostly LGTM :+1: Just a question about the logging:

cluster_tools/cluster_tools/remote.py
philippotto: good point. i cleared it up now :ok_hand:

PR #689, opened 2022-04-07
add BMP

### LINKED ISSUE #680: Support conversion from BMP to wkw/zarr
Next to tiff and the other image formats, bmp files should also be useful and easy to support (there was at least one external user who could have used this).

normanrz:
### Description:
- Adds BMP support

### Issues:
- fixes #680

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/wkcuber/wkcuber/TODO
+++ b/wkcuber/wkcuber/TODO
@@ -1,7 +1,0 @@
-+ export_wkw_as_tiff: remove wkw
-- export_wkw_as_tiff -> export_as_tiff
-- export_wkw_as_nifti -> export_as_nifti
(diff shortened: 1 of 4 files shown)


wkcuber/wkcuber/TODO
```diff
@@ -0,0 +1,7 @@
++ export_wkw_as_tiff: remove wkw
```
philippotto, line 1: probably added accidentally?

philippotto (approved):
:+1:

wkcuber/wkcuber/TODO
normanrz: It's too early ☕

PR #692, opened 2022-04-11
Don't create checkpoint in multiprocessing case if the job failed

philippotto:
### Description:
- Don't create checkpoint in multiprocessing case if the job failed.

### Issues:
- follow-up for https://github.com/scalableminds/webknossos-libs/commit/1cf6f509d46139f552adf09aa4b1bf5933809f55

### Todos:
 - [x] Updated Changelog


daniel-wer (approved):
Thanks for fixing, LGTM :+1:

PR #693, opened 2022-04-12
refactor uses of UPath

normanrz:
Now that https://github.com/fsspec/universal_pathlib/pull/56, https://github.com/fsspec/universal_pathlib/pull/53 and https://github.com/fsspec/universal_pathlib/pull/52 are merged, we can simplify our code a bit.

### DIFF
--- a/webknossos/webknossos/dataset/_array.py
+++ b/webknossos/webknossos/dataset/_array.py
@@ -141,7 +141,9 @@ class WKWArray(BaseArray):
 
     @classmethod
     def open(cls, path: Path) -> "WKWArray":
-        if (path / "header.wkw").is_file():
+        header_path = path / "header.wkw"
+
+        if header_path.exists() and header_path.is_file():
             return cls(path)
         raise ArrayException(
             f"Could not open WKW array at {path}. `header.wkw` not found."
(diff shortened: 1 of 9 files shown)


webknossos/webknossos/dataset/_array.py
```diff
@@ -141,7 +141,9 @@ def __init__(self, path: Path):
 
     @classmethod
     def open(cls, path: Path) -> "WKWArray":
-        if (path / "header.wkw").is_file():
+        header_path = path / "header.wkw"
+
+        if header_path.exists() and header_path.is_file():
```
normanrz, line 146: This is a fix for a bug that @leowe found.

webknossos/webknossos/utils.py
```diff
@@ -213,34 +212,21 @@ def warn_deprecated(deprecated_item: str, alternative_item: str) -> None:
     )
 
 
-def make_upath(maybe_path: Union[str, PathLike, Path]) -> UPath:
-    return maybe_path if isinstance(maybe_path, UPath) else UPath(maybe_path)
-
-
 def is_fs_path(path: Path) -> bool:
```
normanrz, line 215: We could get rid of this as well, but I actually think `is_fs_path` is more expressive than `not isinstance(path, UPath)`.

jstriebel:
@normanrz Is this PR ready for review?

normanrz:
> @normanrz Is this PR ready for review?

Yes!

webknossos/webknossos/dataset/_array.py
jstriebel: One might argue that this is also a bug in UPath, but always sending two requests for the is_file check also seems wrong. In any case the fix is good atm!

jstriebel (approved):
Nice, LGTM 👌

webknossos/webknossos/dataset/_array.py
normanrz: Yeah. I already have a fix for UPath in the pipeline.

PR #697, opened 2022-04-14
Allow newer numpy versions

daniel-wer:
### Description:
- This will allow to use newer numpy versions in tools that use this dependency, for example, voxelytics
- The webknossos package also uses version 1.21.1, currently


wkcuber/wkcuber/convert_nifti.py
```diff
@@ -196,7 +196,6 @@ def convert_nifti(
     # Writing wkw compressed requires files of shape (shard_size, shard_size, shard_size)
     # Pad data accordingly
     padding_offset = shard_size - np.array(cube_data.shape[1:4]) % shard_size
-    padding_offset = (0, 0, 0)
```
daniel-wer, line 199: No idea what this what supposed to accomplish :shrug: Looks like an oversight after a refactoring to me (and mypy complained) which is why I removed it.

daniel-wer: @philippotto It seems like you wrote this code. Do you remember whether you deactivated the padding deliberately?

jstriebel (approved):
LGTM :+1: (except for the `padding_offset = (0, 0, 0)` part, let's see)

wkcuber/wkcuber/convert_nifti.py
philippotto: I think this line simply slipped through while refactoring. Removing it should be fine in my opinion :+1:

PR #700, opened 2022-04-20
fix traps in test scripts

jstriebel:
### Description:
Since our test scripts used multiple traps, they overwrote each other, leading to dangling containers, e.g. in the nightly CI runs.
This fixes it, by appending the trap functions, using this [crazy bash-foo](https://stackoverflow.com/a/7287873/3581794), but slightly adapted to ensure the correct exit code.

I tested this locally.


normanrz:
We could also add this as a fixture to pytest?

normanrz:
This is actually nicely done in upath: https://github.com/fsspec/universal_pathlib/blob/main/upath/tests/conftest.py#L201

jstriebel:
Good idea! I'll add those as fixtures, minio to `test_dataset` and webknossos as a test-wide thing.

jstriebel:
I just changed the minio setup to fixtures now, since the local wk setup is a bit more involved. I'll refactor this, the next time I have to touch it anyways if you agree.

webknossos/tests/test_dataset.py
```diff
@@ -34,14 +34,42 @@
...
+        f" --name {container_name}"
+        " --rm"
+        " -d"
+        " minio/minio server /data"
+    )
+    print("BEFORE", flush=True)
+    subprocess.check_output(shlex.split(cmd))
+    REMOTE_TESTOUTPUT_DIR.fs.mkdirs("testoutput", exist_ok=True)
+    yield
+    subprocess.check_output(["docker", "stop", container_name])
```
normanrz, line 65: Does this need try/finally or will it always call the docker stop?

jstriebel: Good point, I think try-finally is needed.

normanrz (approved):

PR #702, opened 2022-04-20
FinishUpload: stop retrying on error

fm3:
In the webKnossos backend API, the finishUpload call is not idempotent. It should thus not be retried if the server has answered with 400 (only if there was a network error).

Note that the for/else statement had to go, as the break in the 400 case led to the assertion not being executed.

### Todos:
 - [x] Updated Changelog


jstriebel (approved):
:+1:

PR #704, opened 2022-04-20
Adapt terminology

jstriebel:
### Description:
Does some renaming according to https://github.com/scalableminds/webknossos/issues/6053. From the changelogs:

- Renamed some entities to comply with the [webKnossos terminology](https://docs.webknossos.org/webknossos/terminology.html). [#704](https://github.com/scalableminds/webknossos-libs/pull/704):
  * The `Graph` class changed to `Tree`, also related methods and attributes are renamed now, e.g. `add_graph` is now `add_tree`.
    All previous entities are still available, but deprecated.
  * `scale` has changed to `voxel_size` for datasets, skeletons and annotations.
    Changes in `Dataset` are backwards-compatible, but `scale` is deprecated.
    For `Annotation` and `Skeletons` the changes are breaking, `scale` is not supported anymore.
  * `get_best_mag` is deprecated, please use `get_finest_mag` instead
  * In `layer.upscale`, `min_mag` is deprecated in favor of `finest_mag`
- Deprecated `--scale` in favor of `--voxel_size`. [#704](https://github.com/scalableminds/webknossos-libs/pull/704)

### Issues:
- part of https://github.com/scalableminds/webknossos/issues/6053

### Todos:
 - [x] Updated Changelog


webknossos/tests/test_annotation.py
```diff
@@ -28,7 +28,7 @@ def test_annotation_from_zip_file() -> None:
         in annotation.metadata  # pylint: disable=unsupported-membership-test
     )
     assert len(list(annotation.get_volume_layer_names())) == 1
-    assert len(list(annotation.skeleton.flattened_graphs())) == 1
```
normanrz, line 31: Should we also test usage of deprecated identifiers?

webknossos/tests/test_dataset_deprecated.py
```diff
@@ -153,7 +153,7 @@ def assure_exported_properties(ds: Dataset) -> None:
 def test_create_dataset_with_layer_and_mag() -> None:
     rmtree(TESTOUTPUT_DIR / "wkw_dataset")
 
-    ds = Dataset(TESTOUTPUT_DIR / "wkw_dataset", scale=(1, 1, 1))
```
normanrz, line 156: This is the test file for deprecated usages. I think we should keep `scale` here.

webknossos/webknossos/annotation/annotation.py
```diff
@@ -143,13 +145,13 @@ def dataset_name(self, dataset_name: str) -> None:
         self.skeleton.dataset_name = dataset_name
 
     @property
-    def scale(self) -> Tuple[float, float, float]:
```
normanrz, line 146: Add `scale` property with deprecation warning?

webknossos/webknossos/dataset/downsampling_utils.py
```diff
@@ -40,40 +40,42 @@ def determine_buffer_shape(array_info: ArrayInfo) -> Vec3Int:
 
 
 def calculate_mags_to_downsample(
```
normanrz, line 42: Btw should we mark this module as internal (ie. prefix with `_`)?

webknossos/webknossos/skeleton/skeleton.py
```diff
@@ -7,39 +7,36 @@
...
-    graph.add_edge(node_1, node_2)
+    tree.add_edge(node_1, node_2)
     ```
 
     Also see [this example](/webknossos-py/examples/skeleton_synapse_candidates.html) for a more
     complex interaction.
     """
 
-    scale: Vector3
+    voxel_size: Vector3
```
normanrz, line 39: Add `scale` property with deprecation warning?

wkcuber/wkcuber/_internal/utils.py
```diff
@@ -69,13 +69,14 @@ def open_knossos(info: KnossosDatasetInfo) -> KnossosDataset:
     return KnossosDataset.open(info.dataset_path, np.dtype(info.dtype))
 
 
-def add_scale_flag(parser: argparse.ArgumentParser, required: bool = True) -> None:
+def add_voxel_size_flag(parser: argparse.ArgumentParser, required: bool = True) -> None:
     parser.add_argument(
```
normanrz, line 73: Can we mark `--scale` as deprecated somehow?

wkcuber/wkcuber/export_wkw_as_tiff.py
```diff
@@ -103,7 +103,7 @@ def _make_tiff_name(name: str, slice_index: int) -> str:
 
 def _slice_to_image(data_slice: np.ndarray, downsample: int = 1) -> Image:
     if data_slice.shape[0] == 1:
-        # discard greyscale dimension
+        # discard greyvoxel_size dimension
```
normanrz, line 106: 😆

normanrz (commented):
I think `webknossos/webknossos/properties.py` needs changes as well?

wkcuber/wkcuber/export_wkw_as_tiff.py
jstriebel: :see_no_evil: 
```suggestion
        # discard greyscale dimension
```

webknossos/webknossos/skeleton/skeleton.py
jstriebel: This is slightly complicated, since the scale currently is non-optional. I'll add a proposal and we'll see if it's worth it :+1:

webknossos/tests/test_annotation.py
jstriebel: I don't think so.

webknossos/webknossos/skeleton/skeleton.py
jstriebel: Sorry, this is just about the property, not about the init-argument, right? Good idea, I'll add this!

webknossos/webknossos/dataset/sampling_modes.py
```diff
@@ -0,0 +1,4 @@
+class SamplingModes:
```
normanrz, line 1: enum?

wkcuber/wkcuber/_internal/utils.py
```diff
@@ -69,13 +69,14 @@ def open_knossos(info: KnossosDatasetInfo) -> KnossosDataset:
...
+        "--voxel_size",
         "--scale",
         "-s",
-        help="Scale of the dataset (e.g. 11.2,11.2,25). This is the size of one voxel in nm.",
+        help="Voxel size of the dataset in nm (e.g. 11.2,11.2,25). --voxel_size is deprecated",
```
normanrz, line 77:
```suggestion
        help="Voxel size of the dataset in nm (e.g. 11.2,11.2,25). --scale is deprecated",
```

normanrz (commented):
Please see my comment about `properties.py`

webknossos/webknossos/dataset/sampling_modes.py
jstriebel: I thought so too, not sure why it wasn't before.

jstriebel:
> I think `webknossos/webknossos/properties.py` needs changes as well, because we don't want to change the format of `datasource-properties.json`?

I don't think it needs to change, exactly because we want to leave everything in the `datasource-properties.json` as-is. I think that the `properties.py` is basically the schema for the `datasource-properties.json`, and the usages need to adapt, e.g. as [shown here](https://github.com/scalableminds/webknossos-libs/blob/stick-to-terminology/webknossos/webknossos/dataset/dataset.py#L178). We can also adapt it in the `properties.py` and translate this during (de-)serialization, but I'd prefer the current way.

jstriebel:
@normanrz Thanks for the feedback! I think I applied it everywhere now, please have another look :eyes:

normanrz (approved):

PR #707, opened 2022-04-25
typo in doc

erjel:


hotzenklotz:
Thanks for the fix. Much appreciated.

webknossos/webknossos/annotation/annotation.py
```diff
@@ -218,7 +218,7 @@ def download(
         """
         * `annotation_id_or_url` may be an annotation id or a full URL to an annotation, e.g.
           `https://webknossos.org/annotations/Explorational/6114d9410100009f0096c640`
-        * `annotation_type` must be supplied iff an annotation id was used in the previous argument
+        * `annotation_type` must be supplied if an annotation id was used in the previous argument
```
jstriebel, line 221: `iff` is an abbreviation for [if and only if](https://en.wikipedia.org/wiki/If_and_only_if), but the full version definetly is more clear:
```suggestion
        * `annotation_type` must be supplied if and only if an annotation id was used in the previous argument
```

jstriebel (approved):

Issue #257, opened 2020-12-17
Support Python 3.10


hotzenklotz:
Looks like Python 3.9 does not yet work with `wkcuber`. 

https://github.com/scalableminds/webknossos-cuber/runs/1571097340?check_suite_focus=true

erjel:
It also does not work with `webknossos`. With Python 3.9 I get:
```
TypeError: Invalid first argument to `register()`. typing.Literal['color', 'segmentation'] is not a class.
```
for a simple:
```python
import webknossos as wk
```
Please restrict the PyPi dependencies accordingly or run unit test also against 3.9 & 3.10. False badges in `README.md` do not help anyone ...

hotzenklotz:
Thanks for letting us know. 

> Please restrict the PyPi dependencies accordingly or run unit test also against 3.9 & 3.10. False badges in README.md do not help anyone ...

That's unfortunate. This are inferred automatically from the `pyproject.toml`. It seems newer versions are underspecified there. Will fix that.

jstriebel:
wk-libs supports Python 3.9 in the meantime, and the badges in the Readme are fixed. Python 3.10 is not tested yet and might not work.

PR #715, opened 2022-05-02
Restrict pyproject.toml to Python 3.7 and 3.8

hotzenklotz:
### Description:
- As noted in #https://github.com/scalableminds/webknossos-libs/issues/257, PyPi and the widget in our readme lists `webKnossos` as Py `3.9`, and `3.10` compatible, which is not the case.
- This PR restricts the supported versions in the `pyproject.toml`, to Python `3.7` and `3.8` as tested by the CI.

### Issues:
- related to #257 

### Todos:
 - [ ] Updated Changelog


jstriebel (approved):
:+1:

PR #716, opened 2022-05-02
Support for Python 3.9

hotzenklotz:
### Description:
- PR adds fixes to make wk-libs and wkCuber compatible with Py 3.9
  - update some dependencies for compatibility
  - added another weird `allowed_hosts` parameter to fix UNIX sockets for multiprocessing while running the CI test with disabled network
 - update GitHub action to include Py3.9  in test matrix

### Issues:
- partial fix for #257 

### Todos:
 - [ ] Updated Changelog


webknossos/webknossos/__init__.py
```diff
@@ -7,9 +7,9 @@
...
-| **`Skeleton`**   | [`Skeleton(​…)`](webknossos/skeleton/skeleton.html#Skeleton.__init__)         | [`Skeleton.​load(…)`](webknossos/skeleton/skeleton.html#Skeleton.load)                                                                                  | [`skeleton.​save(…)`](webknossos/skeleton/skeleton.html#Skeleton.save)         | *only in an annotation*                                                                           | *only in an annotation*                                                            |
+| **`Dataset`**    | [`Dataset(…)`](webknossos/dataset/dataset.html#Dataset.__init__)             | [`Dataset.open(…)`](webknossos/dataset/dataset.html#Dataset.open),<br/>[`Dataset.open_remote(…)`](webknossos/dataset/dataset.html#Dataset.open_remote) | *N/A*[^ds-save]                                                               | [`Dataset.download(…)`](webknossos/dataset/dataset.html#Dataset.download)                         | [`dataset.upload(…)`](webknossos/dataset/dataset.html#Dataset.upload)              |
```
hotzenklotz, line 10: I removed some zero-width whitespace detected by Pylint

webknossos/webknossos/annotation/annotation.py
```diff
@@ -58,8 +58,8 @@
 
 MAG_RE = r"((\d+-\d+-)?\d+)"
 SEP_RE = r"(\/|\\)"
-CUBE_RE = fr"z\d+{SEP_RE}y\d+{SEP_RE}x\d+\.wkw"
-ANNOTATION_WKW_PATH_RE = re.compile(fr"{MAG_RE}{SEP_RE}(header\.wkw|{CUBE_RE})")
+CUBE_RE = rf"z\d+{SEP_RE}y\d+{SEP_RE}x\d+\.wkw"
```
hotzenklotz, line 61: black formatting

jstriebel: :fr: :crying_cat_face:

webknossos/webknossos/dataset/properties.py
```diff
@@ -278,7 +281,7 @@ def __layer_properties_pre_structure(
                 mag_view_properties_pre_unstructure(m) for m in d["wkwResolutions"]
             ]
             del d["wkwResolutions"]
-        obj = converter_fn(d)
+        obj = converter_fn(d, _type)
```
philippotto, line 284: since `_type` is used now, I'd rename it to `type`.

wkcuber/Changelog.md
```diff
@@ -14,6 +14,7 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Added
 
 ### Changed
+- added Python 3.9 support to wkCuber [#716](https://github.com/scalableminds/webknossos-libs/pull/716)
```
philippotto, line 17:
```suggestion
- Added Python 3.9 support to wkcuber [#716](https://github.com/scalableminds/webknossos-libs/pull/716)
```

In the readme, we use `wkcuber` in all lower case.

philippotto (approved):
Great, lgtm :+1: Only see a few nitpicks.

Should we drop the 3.7 support at some point then? I'm mostly wondering, because the CI now runs 50% more code due to the added 3.9 matrix element.

webknossos/webknossos/dataset/properties.py
hotzenklotz: `type` is a keyword/builtin function in Python. We could rename it to `dtype`, but that typically implies `float` or `uInt8`or similar to me. #namingthings

Open to suggestions.

philippotto: ah, I see. maybe `current_type` or `type_value`. I don't have a strong opinion.

hotzenklotz:
> Great, lgtm 👍 Only see a few nitpicks.
> 
> Should we drop the 3.7 support at some point then? I'm mostly wondering, because the CI now runs 50% more code due to the added 3.9 matrix element.

I am not sure about the relevance of Py3.7. I tried to find out more about the Python version distribution in the wild, but without much success. There is this: https://w3techs.com/technologies/history_details/pl-python/3

My gut feeling is that 3.7 is still relevant, but I can't say why.

cc @normanrz

normanrz:
> Should we drop the 3.7 support at some point then? I'm mostly wondering, because the CI now runs 50% more code due to the added 3.9 matrix element.

Does that slow down the CI? We're not paying for minutes in this repo. I'd say we should keep 3.7 support.

philippotto:
> Does that slow down the CI?

No, I don't think so. Only had the cost in mind.

>  We're not paying for minutes in this repo.

Ok, then let's keep 3.7 :+1:

jstriebel:
RE Python 3.7 support: I think it's useful to keep support for the "official" Python releases that are not EOLed so far if it does not incur problems on our side. There's a good overview of the EOL dates of the Python releases:
https://endoflife.date/python

PR #720, opened 2022-05-03
Fix upsampling with constant z

### LINKED ISSUE #713: Upsampling logic for constant_z correct?
I would like to upsample a segmentation from 4-4-1 to 2-2-1. A minimal example would look similar to:
```python
import tempfile
import numpy as np
from webknossos import Mag, Dataset

def main():
    data = np.ones((3,4,5), dtype=np.uint8)
    scale = (128, 128, 35)
    from_mag = Mag('4-4-1')

    with tempfile.TemporaryDirectory() as path:
        layer = Dataset(path, scale=scale)\
            .get_or_add_layer('test_seg', 'segmentation', largest_segment_id=1)

        layer.get_or_add_mag(from_mag).write(data)

        layer.upsample(
            from_mag = from_mag,
            min_mag = Mag('2-2-1'),
            compress=False,
            sampling_mode = 'constant_z',
        )

if __name__ == '__main__':
    main()
```
I get the error message:
```
RuntimeError: The maximum mag 4-4-1 can
not be reached from 2-2-1 with scale (128.0, 128.0, 35.0)!
```
I do not see how 4-4-1 shouldn't be reachable from 2-2-1. (Note the switched target <-> source mag order due to the reuse of the downsampling code).  Together with the comment:
https://github.com/scalableminds/webknossos-libs/blob/7777ca18f8bab66a1f34631ba75fc6e2947d4566/webknossos/webknossos/dataset/downsampling_utils.py#L49-L51
I concluded that either the comment is wrong or there is a hidden logic error because:

https://github.com/scalableminds/webknossos-libs/blob/0713b80d3dbb9d0ce9575fbaf2ff47644eff9d64/webknossos/webknossos/dataset/layer.py#L919-L927
Since `voxel_size` equals `scale` in the presented case. 

Tested with webknossos `0.9.21`.

jstriebel:
### Description:
The previous upsampling code broke with constant z upsampling, when the voxel-size ratio was not constant, see #713.
`calculate_mags_to_upsample` / `calculate_mags_to_downsample` expects the `voxel_size` to be None in case of constant z, which is fixed in this PR.

### Issues:
- fixes #713

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -19,6 +19,7 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Changed
 
 ### Fixed
+- Fixed upsampling with constant z in certain anisotropic cases. [#720](https://github.com/scalableminds/webknossos-libs/pull/720)
 
 
 ## [0.9.23](https://github.com/scalableminds/webknossos-libs/releases/tag/v0.9.23) - 2022-05-03
--- a/webknossos/webknossos/dataset/layer.py
+++ b/webknossos/webknossos/dataset/layer.py
@@ -924,7 +924,7 @@ class Layer:
             finest_mag_with_fixed_z = finest_mag.to_list()
             finest_mag_with_fixed_z[2] = from_mag.to_list()[2]
             finest_mag = Mag(finest_mag_with_fixed_z)
-            voxel_size = self.dataset.voxel_size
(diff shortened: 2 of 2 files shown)


fm3 (approved):
LGTM :) Thanks for looking into this!

PR #721, opened 2022-05-03
Readme: update python version badge

jstriebel:
Updates the Python version badge.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,10 @@
 # webKnossos-libs
 [![PyPI version](https://img.shields.io/pypi/v/webknossos)](https://pypi.python.org/pypi/webknossos)
-[![Supported Python Versions](https://img.shields.io/pypi/pyversions/webknossos.svg)](https://pypi.python.org/pypi/webknossos)
+[![Supported Python Versions](https://img.shields.io/badge/Python-3.7%20%7C%203.8-blue)](https://pypi.python.org/pypi/webknossos)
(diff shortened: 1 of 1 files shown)


hotzenklotz:
This is odd, how Shield.io parses the version. From the code, I can see that they check the PyPi listing for the attribute `Classifiers :: Programming Language`, which currently is:

image

Hence, the `Py3.8` badge is correct. I wonder if we could correct the metadata on PyPi instead and hence fix the bade in the process. (e.g. see Rasa [repo badge](https://github.com/rasaHQ/rasa) and [PyPi](https://pypi.org/project/rasa/)

hotzenklotz:
The "Classifier" info is likely directly filled out by either `poetry publish` or inferred by `PyPi`. In any case it seems to have changed with my last PR on limiting the Python version in our `project.toml`.

webKnossos:
```
[tool.poetry.dependencies]
python = ">=3.7.1, <3.10"
```

Rasa:
```
[tool.poetry.dependencies]
python = ">=3.7,<3.10"
```

Note the lack of a space between a space and the lack of any patch version in their string. Perhaps, we try that.

jstriebel:
@hotzenklotz Yes, the reason 3.7 is missing is because of the `>=3.7.1` specification we have. However, this is needed, since some dependencies exclude `3.7.0` specifically, since it contained important bugs. That's why I wanted to adapt the badge manually. However, I just found out that we can manually add classifiers, I'll try that instead:
https://python-poetry.org/docs/pyproject/#classifiers

jstriebel:
Superseded by #722

PR #723, opened 2022-05-04
Remote Dataset

### LINKED ISSUE #630: changing allowed teams on a dataset from API
Automating the upload of multiple datasets I always need to change manually the allowed teams on each specific dataset.
It would be useful to be able to do that automatically from the `webknossos` package.

There's no documented API endpoint to do that on 
https://docs.webknossos.org/webknossos/rest_api.html
but the web interface seems to be sending a PATCH request to
```/datasets/<organization-id>/<dataset-name>/teams```

Raw request:
```
PATCH /api/datasets/<organization-id>/<dataset-name>/teams HTTP/1.1
Host: webknossos.org
X-Auth-Token: <token>
Content-Type: application/json
Content-Length: 28

["<team-id>"]
```

By sending this request with cURL I'm able to achieve ```200``` status and the allowed teams are actually changed
```bash
/bin/bash

curl -X PATCH https://webknossos.org/api/datasets/<organization-id>/<dataset-name>/teams -H "X-Auth-Token: <token>" -H "Content-Type: application/json" -d "[\"<team-id>\"]"
```

and also using Python ```requests```
 
```python
import requests
from requests.structures import CaseInsensitiveDict

base_url = "https://webknossos.org/api"
organization_id = <organization-id>
dataset_name = <dataset-name>

url = f"{base_url}/{organization_id}/{dataset_name}/teams"

headers = CaseInsensitiveDict()
headers["X-Auth-Token"] = <token>
headers["Content-Type"] = "application/json"

data = '["<team-id>"]'

resp = requests.patch(url, headers=headers, data=data)

print(resp.status_code)
```

[I'm not really sure why but yesterday the same code didn't work reliably and sometimes the requests were returning a 415 error. I suspect that sending a single array and not a JSON like ```{"teams" : [team_id,]}``` is somewhat nonstandard or maybe something changed on the server side?]

### LINKED ISSUE #688: Remote Datasets
The `webknossos` package should contain a class that allows streaming access to a dataset in webKnossos.
I think the zarr interface in webKnossos would be very useful for this https://github.com/scalableminds/webknossos/issues/6106.

With that a dataset could be open liked this:
```python
import webknossos as wk

# low-level zarr access
ds = wk.Dataset.open(UPath(
    "https://webknossos.org/data/zarr/scalable_minds/l4dense_motta_et_al_demo", 
    storage_options={ "headers": { "Authorization": "Bearer my_webknossos_token" } },
))

# or more convenient
with wk.webknossos_context(token="my_webknossos_token"):
    ds = wk.Dataset.remote_open(name="l4dense_motta_et_al_demo")

some_data = ds.get_layer("segmentation").get_mag("1").read(absolute_offset=(1024, 1024, 1024), size=(32, 32, 32))
```

jstriebel:
### Description:
- Added `Dataset.open_remote`, which returns an object of the new `RemoteDataset`. This can
  * read image data via the webknossos zarr interface (see https://github.com/scalableminds/webknossos/pull/6144)
    (added a new read-only flag at the Dataset)
  * read and change the following dataset metadata as properties: `display_name`, `description`, `tags`, `is_public`, `sharing_token`, `allowed_teams`
* Added docs & tests for the new features.
* Also updated the docs-packages and adapted the pdoc templates (sorry, that I squeezed this in here).
  I verified that the docs still look the same, I'll add a link here after the first CI build.
  This was mostly to be able to display the overview table with more width now:
  Screenshot from 2022-05-04 14-38-39
  PS: Should I combine the load & download columns into a single one, same for save & upload?

### Notes for the review:
* I left `RemoteDataset` in `dataset.py` for now, to avoid a separate page in the docs. Please let me know if it should have its own file.
* I did not adapt the examples or add new ones for the new API yet, I think this might be a good follow-up PR.
* If a dataset can not be opened via zarr (e.g. due to old datastore or wk-connect datastore or just temporary issues), a stub Dataset is returned, so that the metadata-properties can still be used. Do you think that's ok this way?
* Read-only datasets can still be mutated, but the write will not be performed, instead an error is thrown. Do you think that's fair? Should a dataset reset itself before throwing the error, so that internal state is not mutated? Or check read-only before each method that mutates the dataset? Or leave it as-is?

### Issues:
- fixes #688
- fixes #630

### Todos:
 - [ ] Updated Changelog


jstriebel:
The docs were pushed here, but it seems that the branch-detection in the CI job is outdated, other CIs might overwrite it. I'll have a look tomorrow how to fix this:
https://static.webknossos.org/docs/merge/api/webknossos/dataset/dataset.html#RemoteDataset

webknossos/webknossos/dataset/dataset.py
```diff
@@ -247,6 +263,125 @@ def open(cls, dataset_path: Union[str, PathLike]) -> "Dataset":
...
+        organization_id: Optional[str] = None,
+        sharing_token: Optional[str] = None,
+        webknossos_url: Optional[str] = None,
+    ) -> "RemoteDataset":
+        """Opens a remote webknossos dataset. Image data is accessed via network requests.
+        Dataset metadata such as allowed teams or the sharing token can be read and set
+        via the respective `RemoteDataset` properties.
+
+        * `dataset_name_or_url` may be a dataset name or a full URL to a dataset view, e.g.
+          `https://webknossos.org/datasets/scalable_minds/l4_sample_dev/view`
```
normanrz, line 339: /view should be optional in the URL

webknossos/webknossos/dataset/layer.py
```diff
@@ -168,12 +168,12 @@ def _get_sharding_parameters(
...
 class Layer:
     """
-    A `Layer` consists of multiple `webknossos.dataset.mag_view.MagView`s, which store the same data in different magnifications.
+    A `Layer` consists of multiple `MagView`s, which store the same data in different magnifications.
     """
 
     def __init__(self, dataset: "Dataset", properties: LayerProperties) -> None:
         """
-        Do not use this constructor manually. Instead use `webknossos.dataset.layer.Dataset.add_layer` to create a `Layer`.
+        Do not use this constructor manually. Instead use `Dataset``.add_layer()` to create a `Layer`.
```
normanrz, line 176:
```suggestion
        Do not use this constructor manually. Instead use `Dataset.add_layer()` to create a `Layer`.
```

webknossos/webknossos/client/_generated/api/datastore/data_layer_folder_contents.py
```diff
@@ -0,0 +1,92 @@
...
+
+def _get_kwargs(
+    organization_name: str,
+    data_set_name: str,
+    data_layer_name: str,
+    *,
+    client: Client,
+    token: Union[Unset, None, str] = UNSET,
+) -> Dict[str, Any]:
+    url = "{}/data/zarr/{organizationName}/{dataSetName}/{dataLayerName}".format(
```
normanrz, line 17: These zarr route functions are used anywhere, right?

normanrz (commented):
Very nice stuff! Looking forward to using this.

I wonder if Dataset.upload should return a RemoteDataset. Then users could set properties right away (eg allowed teams). The RemoteDataset should also have a property that provides the view url.

jstriebel:
> The docs were pushed here, but it seems that the branch-detection in the CI job is outdated, other CIs might overwrite it. I'll have a look how to fix this: https://static.webknossos.org/docs/merge/api/webknossos/dataset/dataset.html#RemoteDataset

Fixed now, docs now show up here:
https://static.webknossos.org/docs/remote_dataset/api/webknossos/dataset/dataset.html#RemoteDataset

webknossos/webknossos/dataset/layer.py
jstriebel: The first version has the benefit, that `Dataset` is linked, in the second nothing is linked. Not sure which to prefer.

webknossos/webknossos/client/_generated/api/datastore/data_layer_folder_contents.py
jstriebel: I guess you meant `aren't used anywhere else`? Exactly, that's the only point where the zarr url is constructed.

webknossos/webknossos/dataset/dataset.py
jstriebel: It is already, I added an example for this to the tests.

jstriebel:
> I wonder if Dataset.upload should return a RemoteDataset. Then users could set properties right away (eg allowed teams). The RemoteDataset should also have a property that provides the view url.

Good idea, I added this and adapted the changelog accordingly.

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1002,3 +1118,203 @@ def _initialize_layer_from_properties(self, properties: LayerProperties) -> Laye
...
+            warnings.warn(
+                f"Cannot open remote webknossos dataset {dataset_path} as zarr. "
+                + "Returning a stub dataset instead, accessing metadata properties might still work.",
+                RuntimeWarning,
+            )
+            self.path = None  # type: ignore[assignment]
+        self._dataset_name = dataset_name
+        self._organization_id = organization_id
+        self._sharing_token = sharing_token
+        self._context = context
```
normanrz, line 1154: Is the ContextManager serializable?

webknossos/webknossos/client/_generated/api/datastore/data_layer_folder_contents.py
jstriebel: I'd keep them, since they're auto-generated. It would be effort to remove them atm (and keep them removed).

webknossos/webknossos/dataset/dataset.py
jstriebel: Do you mean with pickle? Yes, just checked it. However, using a remote dataset with a webknossos-context (e.g. a different wk-url as in the normal context) didn't work before, since the context-manager wasn't [reentrant](https://docs.python.org/3/library/contextlib.html#reentrant-context-managers). I fixed this in the latest commit

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1002,3 +1118,203 @@ def _initialize_layer_from_properties(self, properties: LayerProperties) -> Laye
...
+
+    @classmethod
+    def open(cls, dataset_path: Union[str, PathLike]) -> "Dataset":
+        """Do not call manually, please use `Dataset.remote_open()` instead."""
+        raise RuntimeError("Please use Dataset.open() instead.")
```
valentin-pinkau, line 1159:
```suggestion
        raise RuntimeError("Please use Dataset.remote_open() instead.")
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1002,3 +1118,203 @@ def _initialize_layer_from_properties(self, properties: LayerProperties) -> Laye
...
+        from webknossos.client.context import _get_generated_client
+
+        @attr.define
+        class _TeamIdsBody:
+            """Helper class which is deserialized by the generated client via to_dict."""
+
+            teams: List[str]
+
+            def to_dict(self) -> List[str]:
+                return self.teams
```
valentin-pinkau, line 1307: Is there a better way to do this? Why does to_dict return a List?

valentin-pinkau (commented):
Pretty cool stuff!
Regarding your questions:
- If a dataset can not be opened via zarr (e.g. due to old datastore or wk-connect datastore or just temporary issues), a stub Dataset is returned, so that the metadata-properties can still be used. Do you think that's ok this way?

    Yes, I think this is a good fall back.

- Read-only datasets can still be mutated, but the write will not be performed, instead an error is thrown. Do you think that's fair? Should a dataset reset itself before throwing the error, so that internal state is not mutated? Or check read-only before each method that mutates the dataset? Or leave it as-is?
    
    I think the cleanest approach would be to check read-only before each mutating method.

webknossos/webknossos/administration/user.py
```diff
@@ -35,7 +35,7 @@ class User:
     last_name: str
     created: int
     last_activity: int
-    teams: List["Team"]
+    teams: Tuple["Team", ...]
```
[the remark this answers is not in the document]
jstriebel: This is for immutability: Any changes to the teams list would not be reflected back to webknossos, however, code doing that might suggest so. It's even more clear with tags, where writing is actually possible, but only via assignment (via the property-setter):
```
### assuming tags would be lists:
remote_ds.tags.append("my_favorite_tag")
# changed only the returned object, change is not reflected anywhere,
# using remote_ds.tags afterwards would only give the previous list

### with tuples:
remote_ds += ("my_favorite_tag", ) 
# forces the user to use assignment, which is reflected in the server
```

webknossos/webknossos/dataset/dataset.py
jstriebel: Unfortunately, I don't really think so atm. The origin of this problem is, that the swagger annotation in the backend for the request-payload is not properly typed as a list. I tried to type it this way, but it isn't picked up correctly:
https://github.com/scalableminds/webknossos/blob/fc80c512ee5dd5cba5587ad0e68ccdc6166f6a95/app/controllers/DataSetController.scala#L331-L334
https://webknossos.org/swagger.json

Somehow, this is translated to an empty class `DatasetUpdateTeamsJsonBody` by the client generator. Its instances are deserialized via `to_dict`, as it's assumed that the json body is a json dict. We want to improve the swagger annotations in the future, however, that might involve switching the framework we currently use and is out of scope for this PR.

jstriebel: Actually, I just found a much nicer way by modifying `__generate_client.py`, see commit 3c49b40.

jstriebel:
@normanrz @valentin-pinkau Thanks for the great feedback! I added the read-only checks at the beginning of all mutating methods as proposed, and other minor improvements. Could you have another look on the latest changes?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -964,7 +1088,11 @@ def get_or_create(
         return cls(dataset_path, voxel_size, name, exist_ok=True)
 
     def __repr__(self) -> str:
-        return repr("Dataset(%s)" % self.path)
+        return f"Dataset({repr(self.path)})"
+
+    def _ensure_writable(self) -> None:
+        if self._read_only:
```
normanrz, line 1094: `View` exposes a `read_only` property. Perhaps that would also be a good idea for `Dataset` and `Layer`.

normanrz (approved):
LGTM. See my comment about the `read_only` property.

webknossos/webknossos/dataset/dataset.py
jstriebel: Good idea, added it :+1:

PR #730, opened 2022-05-13
Add option to guide downsampling mags via an existing dataset

### LINKED ISSUE #535: Use existing magnifications of other layers as guide when downsampling
The code already exists for this in the private voxelytics repository. However, we need to figure out how the interface (CLI and programmatic one) should look like for this.

MichaelBuessemeyer:
### Description:
This PR moves [code from the vx repo](https://github.com/scalableminds/voxelytics/blob/024839faa2d564f48d3dbdabf1a648389e77486c/voxelytics/commons/downsampling_helper.py#L152-L152) to wklibs that builds a mag list depending on an existing dataset. 
This can then be used in the downsample function of a layer to downsample it to the mags given by a dataset. This feature is available via a new argument of the downsample function.

### Issues:
- fixes #535

### Todos:
 - [x] Updated Changelog
 - [x] Code an example and test it


MichaelBuessemeyer:
Script for testing
```
import numpy as np

import webknossos as wk
import shutil

# pylint: disable=unused-variable


def main() -> None:
    #####################
    # Opening a dataset #
    #####################

    input_dataset = wk.Dataset.open("testdata/simple_wkw_dataset")
    # Assuming that the dataset has a layer "color"
    input_layer = input_dataset.get_layer("color")

    ###############################################
    # Adding additional mags to the input dataset #
    ###############################################
    input_layer.get_or_add_mag("2-2-1")
    input_layer.get_or_add_mag("4-4-2")
    input_layer.get_or_add_mag("8-8-4")
    input_layer.redownsample()
    assert len(input_layer.mags) == 4
    # Use the mag with the best resolution
    finest_input_mag = input_layer.get_finest_mag()

    ######################
    # Creating a dataset #
    ######################

    output_dataset = wk.Dataset("testoutput/my_new_dataset", voxel_size=(1, 1, 1))
    output_layer = output_dataset.add_layer(
        layer_name="color", category="color", dtype_per_channel="uint8", num_channels=3
    )
    # Create the same mag in the new output dataset
    output_mag = output_layer.add_mag(finest_input_mag.mag)

    ########################################
    # Copying data into the output dataset #
    ########################################
    input_data = finest_input_mag.read(absolute_offset=(10, 20, 30), size=(512, 512, 32))
    output_mag.write(absolute_offset=(10, 20, 30), data=input_data)

    #########################################################################
    # Downsampling the layer to the magnification used in the input dataset #
    #########################################################################
    output_layer.downsample(from_mag=output_mag.mag, dataset_with_reference_resolutions=input_dataset)
    for mag in input_layer.mags:
        assert output_layer.get_mag(mag)

    shutil.rmtree("testoutput")


if __name__ == "__main__":
    main()

```

MichaelBuessemeyer:
@jstriebel This is my first wklibs pr 🎉 .

Do you have time to review it and tell me if I missed something / did not comply with the "style" of this repo?

webknossos/webknossos/dataset/_downsampling_utils.py
```diff
@@ -326,3 +328,80 @@ def downsample_cube_job(
...
+        starting_mag = layer.get_most_coarse_mag().mag
+        starting_index = mag_list_template.index(starting_mag)
+        missing_mags = mag_list_template[starting_index + 1 :]
+        return missing_mags
+    else:
+        # Since the generated mag list does not contain any entries
+        # which are derived from existing mags, we return an empty list
+        # to signal that the cuber's anisotropic downsampling feature
+        # should be used to derive the output mags.
+        return None
```
jstriebel, line 407: I think it's useful to change the overall structure here a bit. Since `calculate_mags_to_downsample` is already in place and we want it's results to be used to fill in missing mags, we can directly put in the logic there.
I think something along those lines might work:
```python
def calculate_mags_to_downsample(
    from_mag: Mag, max_mag: Mag, voxel_size: Optional[Tuple[float, float, float]], dataset_to_align_with: Optional[Dataset] = None,
) -> List[Mag]:
    if dataset_to_align_with is None:
        mags_to_align_with = set()
    else:
        mags_to_align_with = set(mag for layer in dataset_to_align_with.layers.values() for mag in layer.mags.keys())
    mags_to_align_with_by_max_dim = {mag.max_dim: mag for mag in mags_to_align_with}
    assert len(mags_to_align_with) == len(mags_to_align_with_by_max_dim), "TODO (some layers contain different mags for the same mag, this is not allowed)"
    # … previous setup
    while current_mag < max_mag:
        if current_mag.max_dim * 2 in mags_to_align_with_by_max_dim:
            current_mag = mags_to_align_with_by_max_dim[current_mag.max_dim * 2]
            if current_mag > max_mag:
                warnings.warn("TODO (the mag taken from another layer is larger in some dimensions than max_mag)")
       elif … # previous ifs
```

webknossos/webknossos/dataset/_downsampling_utils.py
```diff
@@ -326,3 +328,80 @@ def downsample_cube_job(
             f"Downsampling of target {target_view.bounding_box} failed with {exc}"
         )
         raise exc
+
+
+def find_smallest_mag_of_dataset(
```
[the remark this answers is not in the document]
jstriebel: (we might end up not using the new functions, let's do the larger refactorings first)

jstriebel (commented):
Nice, I think this is a great missing feature! Awesome, that you tackle your first PR here @MichaelBuessemeyer :tada:

I think we might want to incorporate the new feature more into the existing code we have already, and provide an easier interface to the users than we had internally. I added some comments about this, but it might be useful to have a call if you want :telephone_receiver: 

Besides the code changes, please also adapt the documentation (I think it's just the docstring in this case), as well as adding tests for this use-case.Let's also open an issue to add this feature to upsampling as well afterwards.

MichaelBuessemeyer:
@jstriebel I applied the feedback we discussed. Could you please check out the newest changes? 🔍

webknossos/webknossos/dataset/layer.py
```diff
@@ -634,21 +634,30 @@ def _assert_mag_does_not_exist_yet(
...
+        if type(align_with_other_layers) is bool:
+            return self.dataset if align_with_other_layers else None
+        else:
+            return align_with_other_layers
+
     def downsample(
         self,
         from_mag: Optional[Mag] = None,
-        max_mag: Optional[Mag] = None,
+        coarsest_mag: Optional[Mag] = None,
```
jstriebel, line 648: Please add this to the changelog, as this might break keyword arguments (no fallback with the old name necessary IMO).

webknossos/webknossos/dataset/layer.py
```diff
@@ -634,21 +634,30 @@ def _assert_mag_does_not_exist_yet(
...
         interpolation_mode: str = "default",
         compress: bool = True,
         sampling_mode: Union[str, SamplingModes] = SamplingModes.ANISOTROPIC,
+        align_with_other_layers: Union[bool, "Dataset"] = True,
```
jstriebel, line 652: Please also mention in the changelog, that the default behavior changed (probably can be considered a fix).

webknossos/webknossos/dataset/layer.py
```diff
@@ -634,21 +634,30 @@ def _assert_mag_does_not_exist_yet(
                 )
             )
 
+    def _get_dataset_from_align_with_other_layers(
+        self, align_with_other_layers: Union[bool, "Dataset"]
+    ) -> Optional["Dataset"]:
+        if type(align_with_other_layers) is bool:
```
jstriebel, line 640:
```suggestion
        if isinstance(align_with_other_layers, bool):
```

jstriebel (commented):
Awesome, thanks for the great refactoring and the new tests! I added some minor suggestions. I think the new tests take way too long (CI seemed also to timeout). Probably it's most easy to use smaller datasets or simply restrict the layer's bounding box before `downsample()`. Besides that this PR looks great :rocket:

MichaelBuessemeyer:
Thanks for the feedback :D.

> Probably it's most easy to use smaller datasets or simply restrict the layer's bounding box before downsample(). Besides that this PR looks great 🚀

Sadly, this seems not to fix the issue. The extent of the test datasets is already at `(24,24,24)` which should be small enough IMO.
Both tests run in under 3 min 10 sec each (with all three datasets).

Maybe there is something else is the issue why the CI gets canceled?

jstriebel:
> Sadly, this seems not to fix the issue. The extent of the test datasets is already at `(24,24,24)` which should be small enough IMO. Both tests run in under 3 min 10 sec each (with all three datasets).
> 
> Maybe there is something else is the issue why the CI gets canceled?

Hm, that's weird. I have two proposals to move forward with those tests:
* Instead of having a smaller bounding box, you could also use less mags.
* You can add `--durations=0` to the pytest calls to show all test times or `--durations=10` to show the 10 slowest. I think that only happens after those have run, so not sure if this works in CI. [docs](https://docs.pytest.org/en/6.2.x/usage.html#profiling-test-execution-duration)
* [pytest-xdist](https://github.com/pytest-dev/pytest-xdist) or [pytest-parallel](https://github.com/browsertron/pytest-parallel) might speed up the tests by running them parallel (not sure how many cores we have in CI though)
* [pytest-timeout](https://github.com/pytest-dev/pytest-timeout) cancels tests which take too long, might be useful to figure out where the problems are
* There are [even](https://github.com/koaning/pytest-duration-insights) more [plugins](https://github.com/skudriashev/pytest-timer)…

jstriebel (approved):
Thanks for speeding up the tests :racing_car:  LGTM :+1:

webknossos/Changelog.md
```diff
@@ -15,8 +15,11 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Breaking Changes
 
 ### Added
+- Added a new parameter to `Layer.downsample` & `layer.upsample` which takes an existing dataset and uses the magnifications of this dataset as guidance for the downsampling. [#730](https://github.com/scalableminds/webknossos-libs/pull/730)
 
 ### Changed
+- Changed the name of `max_mag` in `Layer.downsample` to `coarsest_mag`. [#730](https://github.com/scalableminds/webknossos-libs/pull/730)
+- The `Layer.downsample` and `Layer.upsample`function now automatically downsample according to magnifications already existing in the dataset. This behaviour can be turned off by setting the new parameter `align_with_other_layers` to `False`. [#730](https://github.com/scalableminds/webknossos-libs/pull/730)
```
philippotto, line 22: in my opinion, these changes are breaking changes, since the interface is not backwards-compatible when named arguments were used before. I think, we should change this in the changelog.

MichaelBuessemeyer: Oh 😅. Looks like I placed them wrong. Did / Do you correct this @philippotto?

philippotto: Just opened a PR for that :)

PR #733, opened 2022-05-18
webknossos_context: Guard against trailing slashes in URL

jstriebel:
### Description:
URLs for the webknossos-context (e.g. in the `WK_URL` env var or via `webknossos_context(url=…)`) may now contain `/` in the end and are sanitized. Before, requests would fail if the URL contained a final `/`.

### Todos:
 - [x] Updated Changelog


fm3 (approved):
:tada:

Issue #607, opened 2022-02-21
Add OmeDataset


normanrz:
The webknossos-libs feature the fantastic `Dataset`-API upon all of the wkcuber features and others build. Since, we want to support Ome-Zarr in the webKnossos ecosystem, I think it would be very handy to have a `OmeDataset` class. Ideally, the Ome implementation would be a drop-in replacement for `Dataset`.

PR #737, opened 2022-05-25
Exports ome-ngff metadata

normanrz:
### Description:
- If a dataset has a Zarr layer, the OME-NGFF metadata as well as `.zgroup` structure will be written out.

### Issues:
- contributes to #607

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests


normanrz:
> Maybe it makes sense to include a test with `https://github.com/ome/ome-zarr-py`?

I think that's too much trouble, because the libs brings many deps and doesn't easily work with our minio S3 etc. The OME format is fairly straight-forward. I added a validation based on the json schema.

jstriebel (approved):
Awesome :+1:

Issue #539, opened 2022-01-01
multiprocessing can lock downsampling


drewlinsley:
Thanks for all of the hard work on these tools, they are excellent. One issue I've found is that multiprocessing can lock up indefinitely during downsampling. I ran into this issue when downsampling a few layers sequentially. 

I can fix the problem by setting number of jobs to 1, disabling multiprocessing, but I'd like to use it in the future. Perhaps joblib is better suited to avoid hanging processes? Thanks in advance for your help!

philippotto:
Thank you for the report! Currently, we cannot reproduce the issue you are describing which is why some details would be helpful:

- How do you initiate the downsampling? E.g., with `python3 -m wkcuber.downsampling --layer_name color . --jobs 32` ?
- Which OS and which python version are you using?
- How many processes are used when you don't provide `--jobs 1`?
- If the deadlock occurs and you hit CTRL+C, you should get a stacktrace. Can you paste that stacktrace here?

philippotto:
Also, what kind of file system are you using (e.g., an NFS)?

philippotto:
Another thought: Could the available RAM be a problem? If the available RAM does not suffice, the OOM killer can easily kill a child process which makes the parent process wait forever. This also fits well to your workaround of using "jobs == 1" (which consumes less RAM  obviously).

philippotto:
When you notice that the process hangs, you can run `sudo egrep -i -r 'killed process' /var/log/` to check whether the OOM killer became active.

drewlinsley:
Hi Philip,

Sorry for the delay! 

- How do you initiate the downsampling? E.g., with python3 -m wkcuber.downsampling --layer_name color . --jobs 32 ?
I use the wk library in a script. I have several layers I downsample, one after the other.
```
<do stuff>
layer_membranes.downsample(compress=True)
<do stuff>
layer_segmentations.downsample(compress=True)
<etc>
```

- Which OS and which python version are you using?
No LSB modules are available.
Distributor ID:	Ubuntu
Description:	Ubuntu 20.04 LTS
Release:	20.04
Codename:	focal

python 3.8

- How many processes are used when you don't provide --jobs 1?
The code defaults to the number of cpus I have, which is 64.

- If the deadlock occurs and you hit CTRL+C, you should get a stacktrace. Can you paste that stacktrace here?

```
[...]

PR #739, opened 2022-05-30
Detect out-of-memory crashes for slurm

### LINKED ISSUE #552: Poll PIDs of child processes to detect OOM killing
When using multiprocessing (via clustertools), it can happen that the child processes are killed due to OOM. If the OOM killer decides to kill the children, the parent potentially waits forever (sometimes a broken process pool exception will be raised, but we've seen cases where this does not happen). A workaround could be to regularly poll for the child PIDs. If they are not running, a warning could be emitted. Maybe a hard timeout would also make sense (where the main process exists if the timeout is exceeded).

philippotto:
### Description:
- Extends the `ClusterExecutor` interface so that subclasses can implement a `investigate_failed_job` method.
- The `SlurmExecutor` implements `investigate_failed_job` to check whether a job violated the specified RAM resources. If this is the case, a special `RemoteOutOfMemoryException` is emitted which can be handled by callers.
- In the future, there might be more reasons than just "OOM" that `investigate_failed_job` returns, but for now the relatively simple implementation should suffice. We can extend this mechanism easily in the future.

### Issues:
- This PR started with the idea of tackling #552, however, in the end there is only little overlap. In the context of multiprocessing, I always got a `BrokenProcessPool` exception when provoking OOM. That exception can (and will be) handled by the resumable executor in the upcoming version of vx. So, there is little to do on that side, as far as I see.
- Therefore, I'd suggest to defer the polling until it becomes actual relevant. I think, older python versions sometimes were not able to catch the broken process pool (ending up in a hanging state), but I think I haven't seen this for quite a while (last time was https://github.com/scalableminds/webknossos-libs/issues/539, but I'm not sure which python version was used there).

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -207,15 +227,29 @@ def _completion(self, jobid, failed_early):
...
+            # - because python isn't installed or the cluster_tools couldn't be found
+            # - because the job was killed (e.g., by slurm due to RAM limit violations)
+            # We don't try to deserialize pickling output, because it won't exist.
             success = False
-            result = "Job submission/execution failed. Please look into the log file at {}".format(
-                self.format_log_file_path(self.cfut_dir, jobid)
+
+            reason = (  # pylint: disable=assignment-from-none
+                self.investigate_failed_job(jobid)
+            )
```
jstriebel, line 242: Maybe `investigate_failed_job` could directly return an Exception itself, so that we don't need to signal the exception-type implicitly via `use_oom_exception`. Possibly also other Exceptions might be raised.

jstriebel (approved):
Nice work! I like the pluggable mechanism! I just left one comment, everything else LGTM, approving already :+1:

cluster_tools/cluster_tools/schedulers/cluster_executor.py
philippotto: good idea! I incorporated that. could you have a final look? :) then, I'd merge the pr.

jstriebel: Sure! I probably would have returned the Exception object, already initialized with the content, and just added the outer context in the cluster-executor, but also fine as-is :+1:

PR #740, opened 2022-05-30
add aiohttp as main dependency

valentin-pinkau:
### Description:
- This PR makes aiohttp a main dependency.
- This is necessary as e.g. Dataset.open_remote() fails with a module not found error citing missing aiohttp.


jstriebel (approved):
Great catch! :fishing_pole_and_fish: 

:shipit:

Issue #719, opened 2022-05-03
Adapt to new datastore API (mag instead of zoomStep)


fm3:
In https://github.com/scalableminds/webknossos/pull/6159/ the webknossos datastore API changes to version 2.0, allowing directly requesting mags rather than using the ambiguous zoomStep paramter.
The datastore download in the python client should use this.

PR #753, opened 2022-06-23
[WIP] Rename user to owner

philippotto:
### Description:
- abc

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


jstriebel:
> @jstriebel There are still type errors which I assume to be due to a differing setup for the client generation. These are the errors:
> 
> ```
> webknossos/client/_download_dataset.py:84: error: "DatasetInfoResponse200DataSourceDataLayersItem" has no attribute "default_view_configuration"  [attr-defined]
> webknossos/client/_download_dataset.py:86: error: "DatasetInfoResponse200DataSourceDataLayersItem" has no attribute "default_view_configuration"  [attr-defined]
> webknossos/administration/project.py:117: error: Item "Unset" of "Union[Unset, ProjectInfoByIdResponse200Owner, ProjectInfoByNameResponse200Owner]" has no attribute "id"  [union-attr]
> ```
> 
> I saw that the client generation expects a dataset `e2006_knossos` to exist. I'm not sure where this is coming from. This dataset is not available on master.webknossos.xyz or webknossos.org. In our notion page, it was uploaded, but that one is in KNOSSOS which can't be right. Locally, I had a `e2006_wkw` which I renamed to `e2006_knossos`, but I assume that it is missing a `default_view_configuration` ?

This is indeed quite tricky. The `__generate_client.py` script uses `e2006_knossos`, because it's [in the test-database](https://github.com/scalableminds/webknossos/blob/master/test/db/dataSets.csv#L7).

When using the local docker-compose setup from the wk-libs repo (which is spawned automatically, but only if nothing is running on port 9000), then this dataset always exists, as it is [checked into this repo](https://github.com/scalableminds/webknossos-libs/tree/master/webknossos/tests/binaryData/Organization_X/e2006_knossos).

It's not quite ideal for a setup, but I'm also not sure, how to change it. Maybe it would be best to simply ignore possible local setups by default and use another port for the client generation? I'd be super happy about suggestions and improvements :-)

> However, I'm also confused by the type error regarding the project info.

This is a separate issue. Here, the generated client uses a sentinel object (basically a singleton) `UNSET` of type `Unset`. You only checked for `None` in the code, not `Unset`.

------------------

Since I was assigned issue #719, I fixed the same problems on PR #760 and also included the deprecations for the username etc. I'll close this PR therefore, but let's still discuss how to improve the setup.

PR #757, opened 2022-06-27
Make resumable upload more robust

philippotto:
### Description:
- The resumable upload should initiate retries if a request itself failed for some reason.

### Issues:
- no issue

### Todos:
 - [x] Updated Changelog


normanrz:
Is there a delay in between retries (or even exp. back-off)?

philippotto:
No, there's just a retry threshold (by default 100). Delays sound good. I can add an exponential back-off, however, the high retry threshold would easily lead to very long wait times (especially, annoying if the error is not temporary and retrying will be pointless). How about a static delay of 5 seconds? That would mean ~ 8 minutes of waiting time in case all attempts fail.

Alternatively, I'd suggest a maximum of 6 retries (instead of 100) in combination with exponential back-off so that the longest/last delay is not longer than 3 minutes. What do you think?

webknossos/webknossos/client/_resumable/chunk.py
```diff
@@ -74,11 +74,15 @@ def _send_chunk(
...
-        data=_build_query(file, chunk, config.additional_query_params),
-        files={"file": chunk.read()},
-    )
+    try:
+        response = client.post(
+            config.target,
+            data=_build_query(file, chunk, config.additional_query_params),
+            files={"file": chunk.read()},
+        )
+    except:
```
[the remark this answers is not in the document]
philippotto: Good point. I decided to catch `httpx.TransportError`. See here for the class hierarchy: https://www.python-httpx.org/exceptions/#the-exception-hierarchy.

webknossos/webknossos/client/_resumable/chunk.py
```diff
@@ -36,6 +39,7 @@ def resolve_chunk(
     if not exists_on_server:
         tries = 0
         while not _send_chunk(client, config, file, chunk):
+            time.sleep(RETRY_DELAY)
```
normanrz, line 42: I assume this is run on a separate thread, so the blocking doesn't prevent simultaneous requests?

normanrz (approved):

webknossos/webknossos/client/_resumable/chunk.py
philippotto: Yes :+1:

PR #759, opened 2022-06-29
Amend breaking change in changelog

philippotto:
### Description:
- Since the interface of `Layer.downsample` was changed, this is a breaking change which might be important when updating.

### Issues:
- follow-up to #730

### Todos:
 - [X] Updated Changelog


MichaelBuessemeyer (approved):
LGTM 👍 
Please ping me if you need help fixing the CI. Strange thing that it failed although you only changed the changelog...

PR #763, opened 2022-07-11
add Team.get_by_name & Dataset.get_remote_dataset_registry

### LINKED ISSUE #628: Add a feature to list all uploaded datasets
It would be nice to have a simple API call to list all available datasets.
At the moment I use the `/datasets` API endpoint and it works well, but it would be a nice feature to add to the `webknossos` python package.

```python
import requests
import pandas as pd

session = requests.session()
headers = {"X-Auth-Token": <webknossos_API_token>}
base_url = "http://webknossos/api"
resp = session.get(base_url + "/datasets", headers=headers)
df = pd.DataFrame(resp.json())
df = df[df.owningOrganization == <organization_id>]

```

### LINKED ISSUE #725: Create wk.Team by name
I don't think the webKnossos UI exposes a team's "ID" somewhere. I think the name is unique per organization. So, perhaps we could use the name to instantiate a `wk.Team` object to specify in `RemoteDataset.allowed_teams`? The constructor could resolve the ID internally with the wk API.

jstriebel:
### Description:
- updates the generated client
- adds `Team.get_by_name()`
- adds `Dataset.get_remote_datasets()`
- adds an example to the docs & tests

### Issues:
- fixes #725 
- fixes #628 

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/dataset.py
```diff
@@ -1355,6 +1366,31 @@ def _initialize_layer_from_properties(self, properties: LayerProperties) -> Laye
                 f"Failed to initialize layer: the specified category ({properties.category}) does not exist."
             )
 
+    @staticmethod
+    def get_remote_dataset_registry() -> Mapping[
+        str, RemoteOrganizationDatasetRegistry
+    ]:
```
normanrz, line 1372: How about?
```suggestion
    @staticmethod
    def list_remote_datasets(organization_id: Optional[str] = None) -> RemoteOrganizationDatasetRegistry:
```

webknossos/webknossos/dataset/remote_dataset_registry.py
```diff
@@ -0,0 +1,87 @@
...
+        )
+        self.by_tag = LazyReadOnlyDict(
+            entries=by_tag,
+            func=lambda names: [
+                Dataset.open_remote(name, organization_id) for name in names
+            ],
+        )
+
+
+class RemoteDatasetRegistry(
```
normanrz, line 48: I'm not sure, we want to make access to all public datasets that easy. I think some users make datasets public, but think they are "unlisted". This would expose all of them. We don't need to close down the API, but only provide library functions for accessing a specified organization (the user's organization by default).

normanrz: Does the API already have an organization filter? @fm3

jstriebel: I could also easily make the dict a function, so it's not possible to iterate over the keys anymore, e.g.
`list_remote_datasets(organiztation_id: Optional[str]=None)`, where the `None` default indicates to use the current organization.

webknossos/webknossos/dataset/dataset.py
normanrz: Well, in combination with [my other comment](https://github.com/scalableminds/webknossos-libs/pull/763#discussion_r918960895), it essentially becomes a list

webknossos/webknossos/dataset/remote_dataset_registry.py
normanrz: Ideally, we wouldn't even load the datasets from other orgs in this API call.

webknossos/webknossos/dataset/remote_dataset_registry.py
```diff
@@ -0,0 +1,87 @@
...
+
+K = TypeVar("K")  # key
+V = TypeVar("V")  # value
+C = TypeVar("C")  # cache
+
+
+class RemoteOrganizationDatasetRegistry(LazyReadOnlyDict[str, "RemoteDataset"]):
+    """Dict-like class mapping dataset names to `RemoteDataset` instances."""
+
+    by_display_name: Mapping[str, "RemoteDataset"]
```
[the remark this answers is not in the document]
jstriebel: Atm the uploader is not exposed. I added this to the route wishlist here: https://github.com/scalableminds/webknossos/issues/6329

webknossos/webknossos/dataset/remote_dataset_registry.py
jstriebel: > Ideally, we wouldn't even load the datasets from other orgs in this API call.

This is not possble atm, I added an issue for it: https://github.com/scalableminds/webknossos/issues/6329

I filtered by the current organization now by default in the client.

webknossos/webknossos/dataset/remote_dataset_registry.py
```diff
@@ -0,0 +1,48 @@
...
+    ) -> None:
+        from webknossos.administration.user import User
+        from webknossos.client._generated.api.default import dataset_list
+        from webknossos.client.context import _get_generated_client
+        from webknossos.dataset.dataset import Dataset
+
+        client = _get_generated_client(enforce_auth=True)
+
+        if organization_id is None:
+            organization_id = User.get_current_user().organization_id
```
normanrz, line 30: The code for filtering by organization is missing.

jstriebel: Good catch! :see_no_evil: Fixed it in the new commit.

normanrz (approved):

PR #764, opened 2022-07-12
update client, allow annotation urls without type

### LINKED ISSUE #749: Support wk URLs which don't have Explorational or Task in url
By default, wK omits the `Explorational` and `Task` part from URLs (see https://github.com/scalableminds/webknossos/pull/6208). This is not yet supported by wklibs.

jstriebel:
### Description:
- updates the generated client, allows better definition of optional fields
- allows to download annotations without specifying the type (either via URL or parameter)

### Issues:
- fixes #749
- depends on #763, since I branched from there due to changes in the generated client

### Todos:
 - [ ] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/annotation/annotation.py
+++ b/webknossos/webknossos/annotation/annotation.py
@@ -229,8 +229,8 @@ class Annotation:
     ) -> "Annotation":
         """
         * `annotation_id_or_url` may be an annotation id or a full URL to an annotation, e.g.
-          `https://webknossos.org/annotations/Explorational/6114d9410100009f0096c640`
-        * `annotation_type` must be supplied if and only if an annotation id was used in the previous argument
+          `https://webknossos.org/annotations/6114d9410100009f0096c640`
+        * `annotation_type` is not used anymore
         * `webknossos_url` may be supplied if an annotation id was used
           and allows to specifiy in which webknossos instance to search for the annotation.
           It defaults to the url from your current `webknossos_context`, using https://webknossos.org as a fallback.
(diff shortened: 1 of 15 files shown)


webknossos/webknossos/annotation/annotation.py
```diff
@@ -229,8 +229,8 @@ def download(
     ) -> "Annotation":
         """
         * `annotation_id_or_url` may be an annotation id or a full URL to an annotation, e.g.
-          `https://webknossos.org/annotations/Explorational/6114d9410100009f0096c640`
-        * `annotation_type` must be supplied if and only if an annotation id was used in the previous argument
+          `https://webknossos.org/annotations/6114d9410100009f0096c640`
+        * `annotation_type` is ignored
```
normanrz, line 233:
```suggestion
        * `annotation_type` is no longer required and therefore deprecated and ignored
```

webknossos/webknossos/annotation/annotation.py
```diff
@@ -246,20 +246,17 @@ def download(
...
-            assert annotation_type is not None, (
-                "When calling Annotation.download() with an id you must supply the argument annotation_type, "
-                + "e.g. 'Task' or 'Explorational'. Alternatively, you can use the full annotation url, "
-                + "e.g. Annotation.download('https://webknossos.org/annotations/Explorational/6114d9410100009f0096c640')."
-            )
             annotation_id = annotation_id_or_url
-        annotation_type = AnnotationType(annotation_type)
+        annotation_type = (
+            None if annotation_type is None else AnnotationType(annotation_type)
+        )
```
normanrz, line 259: Shouldn't this also be removed?
```suggestion
```

normanrz: same as the assertion below

jstriebel:
@normanrz Thanks for the review! I properly deprecated the argument now. Before it was still checked, that the annotation type is not one of the compound types, but I agree that this is not important anymore, the download will simply fail in this case now.

normanrz (approved):

PR #776, opened 2022-07-25
update universal-pathlib

normanrz:
update universal-pathlib


jstriebel (changes requested):
Many `metadata.files` are empty now in the `poetry.lock` files, this is probably due to an old poetry version or an outdated pypi cache. Please check that you use poetry `1.1.14` (`poetry --version`) and rerun the updated after doing `poetry cache clear pypi --all`.

jstriebel (approved):
LGTM :+1:

PR #778, opened 2022-07-29
Detect OOM if exit code is suspicious

### LINKED ISSUE #751: Also detect OOM if seff command does not look suspicious (but exit code does)
If `seff` outputs something like `State: FAILED (exit code 137)`, the chances of an OOM a very likely.

In the case of an array job, one **could** check the other jobs, but this might be quite some overhead (per failed job, all jobs of the array job would probably be needed to check).

philippotto:
### Description:
- Also detect OOM if seff command does not look suspicious (but exit code does)

### Testing:
- It's a bit hard to provoke this special scenario (this only happens if the seff output is not very accurate) which is why I simply used `scancel jobid` to cancel the running job and adapted the code temporarily to watch for the exit code which appears then (weirdly, it's `0`). I think, this should be enough to test the general logic.

@jstriebel should be easier to review by disabling whitespace diffing in github :)

### Issues:
- fixes #751

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/cluster_tools/cluster_tools/schedulers/slurm.py
+++ b/cluster_tools/cluster_tools/schedulers/slurm.py
@@ -332,36 +332,58 @@ class SlurmExecutor(ClusterExecutor):
         if exit_code != 0:
             return None
 
-        # Look for the relevant line.
+        # Parse stdout into a key-value object
+        properties = {}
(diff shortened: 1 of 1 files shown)


jstriebel (approved):
LGTM :+1:

PR #785, opened 2022-08-12
update generated client and snapshots

jstriebel:
### Description:
- Updates the generated client, making only necessary fields required, which makes the client more resilient to changing unused fields. (Also, zarr endpoints are filtered, just to reduce the number of unused generated code).
- Updates the test snapshots
- Adds a fix for `task.get_project()`
- Better error handling for `wk.Dataset.download`, when dataset is not correctly available.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/__generate_client.py
```diff
@@ -336,8 +420,9 @@ def set_response_schema_by_example(
         for path_method in path.values()
         if path_method["operationId"] == operation_id
     ][0]
-    make_properties_required(recorded_response_schema)
+    did_require = make_properties_required(recorded_response_schema)
     request_schema["responses"]["200"]["content"] = recorded_response_schema
+    return did_require
```
fm3, line 425: the return type here is `None`, yet `did_require` is returned? Also I’m uncertain about the name, `did_require` sounds a bit like a boolean? maybe `required_keys` or similar?

webknossos/__generate_client.py
```diff
@@ -336,8 +420,9 @@ def set_response_schema_by_example(
         for path_method in path.values()
         if path_method["operationId"] == operation_id
     ][0]
-    make_properties_required(recorded_response_schema)
+    did_require = make_properties_required(recorded_response_schema)
```
fm3, line 423:
```suggestion
    did_require = make_properties_required_inplace(recorded_response_schema)
```
(just a suggestion, feel free to ignore. Coming from scala-land everything casually mutable scares me :D)

webknossos/webknossos/dataset/layer.py
```diff
@@ -1034,9 +1034,9 @@ def _setup_mag(self, mag: Mag) -> None:
                 info.compression_mode,
             )
             self._mags[mag]._read_only = self._dataset.read_only
-        except ArrayException as e:
-            logging.error(
-                f"Failed to setup magnification {mag_name}, which is specified in the datasource-properties.json. See {e}"
+        except ArrayException:
+            logging.exception(
+                f"Failed to setup magnification {mag_name}, which is specified in the datasource-properties.json:"
```
fm3, line 1039: The colon makes me think something comes afterwards, but the method ends here, is there some magic going on? :D

fm3 (commented):
Mostly LGTM, I added a few notes :)

webknossos/__generate_client.py
```diff
@@ -64,6 +64,21 @@ def add_api_prefix_for_non_data_paths(openapi_schema: Dict) -> None:
         paths[f"/api{path}"] = paths.pop(path)
 
 
+def remove_zarr_tagged_endpoints(openapi_schema: Dict) -> None:
+    """The current webKnossos backend does not include the
+    /api prefix into the different backend paths.
+    Howevery, the /data prefix for datastore paths is included.
+    This adds the missing /api prefixes in the openapi_schema (in-place)."""
```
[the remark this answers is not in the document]
jstriebel: Good catch, copy pasta :spaghetti:

webknossos/webknossos/dataset/layer.py
jstriebel: Yes :D Afterwards, the current exception is logged as well, some black magic :magic_wand: 
https://docs.python.org/3/library/logging.html#logging.exception

fm3: Ah, I missed the `.exception`. That’s not all that implicit then, fair enough 👍

webknossos/__generate_client.py
jstriebel: The I'd have to set it for most functions, I'll leave it as-is for now, I double-checked that it's mentioned in all doc-strings.

webknossos/__generate_client.py
jstriebel: Ah, this was left out from mypy so far, good catch! I added it to typecheck.sh and resolved the type errors.

jstriebel:
@fm3 Thanks for the great feedback! I think I applied all of it, please have another look :eyes:

webknossos/__generate_client.py
```diff
@@ -64,6 +64,18 @@ def add_api_prefix_for_non_data_paths(openapi_schema: Dict) -> None:
         paths[f"/api{path}"] = paths.pop(path)
 
 
+def remove_zarr_tagged_endpoints(openapi_schema: Dict) -> None:
+    """This removes all endpoints tagged with zarr-streaming, in-place."""
```
fm3, line 68:
```suggestion
    """This removes all endpoints tagged with zarr-streaming, in-place, to reduce the size of the generated client."""
```
(at least that’s how I understand it?)

fm3 (approved):
LGTM

webknossos/__generate_client.py
jstriebel: Sorry, automerge was too fast here, I'll include this in my next PR.

fm3: It’s not super important anyway. Sorry, I had not noticed that automerge was already active here

PR #786, opened 2022-08-19
make largest_segment_id optional

jstriebel:
### Description:
Adapts the largestSegmentId to be optional for segmentation layers. This is a minimal change that allows compatibility with webknossos, but still computes the largest segment id when possible as before.

### Issues:
- compare #6414

### Todos:
 - [x] Updated Changelog


fm3 (approved):
LGTM :)

PR #788, opened 2022-08-23
Add SLURM_MAX_ARRAY_SIZE env var for array jobs in cluster_tools

elhuhdron:
### Description:

In this branch / pull request I have added an environment variable for the slurm executor that I tried to pattern off the other env variables. This variable, if set, specifies the maximum number of simultaneously running slurm array jobs. I tested it for my use case on our cluster and it works, but as usual, did not add any test case in this pull request, sorry.

### Issues:

I've encountered scaling issues on our cluster that appear to relate to maximum number of running jobs. I think this is due to a problem with small file writes on our system. This means that many of the cluster_tools executor futures all relying on the `.cfut` directory causes some type of read/write swarm problem. In some cases the future writes get dropped entirely and this results in the future essentially hanging.

The features that have been added previously for controlling max submit and max array sizes worked very well for me. For this reason, could I please request an additional one that controls the slurm array maximum number of running jobs?


cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -118,6 +118,18 @@ def get_max_array_size():
             )
         return max_array_size
 
+    @staticmethod
+    @cache_in_production
+    def get_max_running():
+        max_running_size_env = os.environ.get("SLURM_MAX_RUNNING", None)
```
philippotto, line 124: I'd rename the method to `get_max_running_size` and the env var to `SLURM_MAX_RUNNING_SIZE` (this is also how the debug message mentions it).

philippotto (changes requested):
Thank you for your contribution! The PR looks good except for a minor naming nitpick :)

cluster_tools/cluster_tools/schedulers/slurm.py
elhuhdron: Cool. No problem; var name changes pushed just now.

elhuhdron: I also added a sentence to the README, in case you want to accept the pull request directly.

philippotto:
Awesome, thank you. Now, we only have to get the CI green. You need to run `./format.sh` to get the formatting right. Running `./lint.sh` might also be a good idea, but the CI will do so, too.

elhuhdron:
Ran both and re-pushed the format changes. Linter replied:
```
Your code has been rated at 10.00/10
```
:)

philippotto (approved):
Excellent, thank you :tada:

philippotto:
Hmm, the formatting is still not right. Since lines changed which were originally not touched by this PR, I assume that your installed version of `black` doesn't match the one which is supposed to be used (according to poetry.lock). Could you please doublecheck that you have `black` with version 21.12b0 installed and run `./format.sh` again?

elhuhdron:
K, pushed again. Sorry, I did not realize that the poetry requirements were different between the different webknossos-libs subprojects. I had a new version of black because I setup the environment by running `poetry install` in each of the subfolders, but in the order `cluster_tools`, `webknossos`, `wkcuber`.

PR #790, opened 2022-08-30
coherent fs-copy and data-streaming in dataset copy methods

### LINKED ISSUE #784: copy layer and dataset should use file-copy locally and streaming remote ds
## Detailed Description

At the moment copying datasets, layers and mags sometimes uses array reading and writing and sometimes filecopy. This should be cleaned up.

Overview of the current state:
* `copy_dataset` reads and writes the data, allowing re-chunking and -sharding
* `ds.add_copy_layer()` uses filecopy
* `layer.add_copy_mag()` uses filecopy

filecopy does not work well with remote datasets due to directory listing limitations. We should use array read/write in those cases. filecopy has the benefit of also copying extra data such as meshes or mappings, and should be faster for compressed datasets. Therefore we probably want to use filecopy for local datasets, and read/write for remote ones. When adapting this, symlinking should also be looked at again, to avoid broken symlinks to remote datasets.

## Use Cases & Context

This will resolve #781 and similar issues, cleaning up our copy API.

jstriebel:
### Description:
* renamed `dataset.add_copy_layer()` to `dataset.add_fs_copy_layer()`
* renamed `layer.add_copy_mag()` to `layer.add_fs_copy_mag()`
* `dataset.add_copy_layer()` and `layer.add_copy_mag()` now read and write the image data
* `dataset.copy_dataset` uses `dataset.add_copy_layer()` internally, continuing to read/write the data

An overview of the different copy and symlink methods after this PR:
```
dataset.copy_dataset(to): data r/w
dataset.shallow_copy_dataset(to): fs symlinking of first-level files & folders (e.g. layers)

dataset.add_copy_layer(from): data r/w
dataset.add_symlink_layer(from): fs symlinking
dataset.add_fs_copy_layer(from): fs copy

layer.add_copy_mag(from): data r/w
layer.add_symlink_mag(from): fs symlinking
layer.add_fs_copy_mag(from): fs copy
```

Should I also add `dataset.fs_copy_dataset()`? Probably not for now, since we've never needed it.

### Issues:
- fixes #784

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/tests/test_dataset.py
+++ b/webknossos/tests/test_dataset.py
@@ -1562,7 +1562,56 @@ def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
     assert tuple(layer.bounding_box.topleft) == (6, 6, 6)
     assert tuple(layer.bounding_box.size) == (10, 20, 30)
 
-    copy_mag = layer.add_copy_mag(original_mag_2)
+    copy_mag = layer.add_copy_mag(original_mag_2, extend_layer_bounding_box=False)
+
+    assert (symlink_path / "color" / "1").exists()
+    assert len(layer._properties.mags) == 2
+
+    assert tuple(layer.bounding_box.topleft) == (6, 6, 6)
+    assert tuple(layer.bounding_box.size) == (10, 20, 30)
+
+    # Write data in copied layer
+    write_data = (np.random.rand(5, 5, 5) * 255).astype(np.uint8)
+    copy_mag.write(absolute_offset=(0, 0, 0), data=write_data)
+
+    assert np.array_equal(
+        copy_mag.read(absolute_offset=(0, 0, 0), size=(10, 10, 10))[0], write_data
+    )
+    assert np.array_equal(original_layer.get_mag(2).read()[0], original_data)
+
+    assure_exported_properties(ds)
+    assure_exported_properties(original_ds)
+
+
+@pytest.mark.parametrize("data_format,output_path", DATA_FORMATS_AND_OUTPUT_PATHS)
+def test_add_fs_copy_mag(data_format: DataFormat, output_path: Path) -> None:
+    ds_path = prepare_dataset_path(data_format, output_path, "original")
+    symlink_path = prepare_dataset_path(data_format, output_path, "with_symlink")
+
+    original_ds = Dataset(ds_path, voxel_size=(1, 1, 1))
+    original_layer = original_ds.add_layer(
+        "color", COLOR_CATEGORY, dtype_per_channel="uint8", data_format=data_format
+    )
+    original_layer.add_mag(1).write(
+        data=(np.random.rand(10, 20, 30) * 255).astype(np.uint8)
+    )
+    original_data = (np.random.rand(5, 10, 15) * 255).astype(np.uint8)
+    original_mag_2 = original_layer.add_mag(2)
+    original_mag_2.write(data=original_data)
+
+    ds = Dataset(symlink_path, voxel_size=(1, 1, 1))
+    layer = ds.add_layer(
+        "color", COLOR_CATEGORY, dtype_per_channel="uint8", data_format=data_format
+    )
+    layer.add_mag(1).write(
+        absolute_offset=(6, 6, 6),
+        data=(np.random.rand(10, 20, 30) * 255).astype(np.uint8),
+    )
+
+    assert tuple(layer.bounding_box.topleft) == (6, 6, 6)
+    assert tuple(layer.bounding_box.size) == (10, 20, 30)
+
+    copy_mag = layer.add_fs_copy_mag(original_mag_2)
 
     assert (symlink_path / "color" / "1").exists()
     assert len(layer._properties.mags) == 2
--- a/webknossos/tests/test_dataset.py
+++ b/webknossos/tests/test_dataset.py
@@ -1536,100 +1536,78 @@ def test_remote_add_symlink_mag() -> None:
 
 @pytest.mark.parametrize("data_format,output_path", DATA_FORMATS_AND_OUTPUT_PATHS)
 def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
-    ds_path = prepare_dataset_path(data_format, output_path, "original")
-    symlink_path = prepare_dataset_path(data_format, output_path, "with_symlink")
+    original_ds_path = prepare_dataset_path(data_format, output_path, "original")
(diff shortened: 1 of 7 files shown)


webknossos/tests/test_dataset.py
```diff
@@ -1562,7 +1562,56 @@ def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
...
+
+    original_ds = Dataset(ds_path, voxel_size=(1, 1, 1))
+    original_layer = original_ds.add_layer(
+        "color", COLOR_CATEGORY, dtype_per_channel="uint8", data_format=data_format
+    )
+    original_layer.add_mag(1).write(
+        data=(np.random.rand(10, 20, 30) * 255).astype(np.uint8)
+    )
+    original_data = (np.random.rand(5, 10, 15) * 255).astype(np.uint8)
+    original_mag_2 = original_layer.add_mag(2)
```
philippotto, line 1599: I think use can use this variable in line 1629 in place of `original_layer.get_mag(2)`.

webknossos/tests/test_dataset.py
```diff
@@ -1562,7 +1562,56 @@ def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
     assert tuple(layer.bounding_box.topleft) == (6, 6, 6)
     assert tuple(layer.bounding_box.size) == (10, 20, 30)
 
-    copy_mag = layer.add_copy_mag(original_mag_2)
+    copy_mag = layer.add_copy_mag(original_mag_2, extend_layer_bounding_box=False)
+
+    assert (symlink_path / "color" / "1").exists()
```
philippotto, line 1567: I think the variable name `symlink_path` doesn't make sense, anymore, since there's no symlinking done in this test, right?

webknossos/tests/test_dataset.py
```diff
@@ -1562,7 +1562,56 @@ def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
...
+    assert np.array_equal(original_layer.get_mag(2).read()[0], original_data)
+
+    assure_exported_properties(ds)
+    assure_exported_properties(original_ds)
+
+
+@pytest.mark.parametrize("data_format,output_path", DATA_FORMATS_AND_OUTPUT_PATHS)
+def test_add_fs_copy_mag(data_format: DataFormat, output_path: Path) -> None:
+    ds_path = prepare_dataset_path(data_format, output_path, "original")
+    symlink_path = prepare_dataset_path(data_format, output_path, "with_symlink")
```
philippotto, line 1589: same here

webknossos/tests/test_dataset.py
```diff
@@ -1562,7 +1562,56 @@ def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
...
+    )
+    layer.add_mag(1).write(
+        absolute_offset=(6, 6, 6),
+        data=(np.random.rand(10, 20, 30) * 255).astype(np.uint8),
+    )
+
+    assert tuple(layer.bounding_box.topleft) == (6, 6, 6)
+    assert tuple(layer.bounding_box.size) == (10, 20, 30)
+
+    copy_mag = layer.add_fs_copy_mag(original_mag_2)
```
philippotto, line 1614: I'd name this `copy_mag_2`, since this will be mag 2, as well. Initially, I thought it would be mag 1 and was confused about what the test is comparing.. :see_no_evil:

philippotto: is there a specific reason why the test is dealing with mag 2? I think, the test could be a bit simpler if it was just dealing with mag 1. or at least don't write mag 1 if it's not read, anyway.

webknossos/tests/test_dataset.py
```diff
@@ -1562,7 +1562,56 @@ def test_add_copy_mag(data_format: DataFormat, output_path: Path) -> None:
...
+        copy_mag.read(absolute_offset=(0, 0, 0), size=(10, 10, 10))[0], write_data
+    )
+    assert np.array_equal(original_layer.get_mag(2).read()[0], original_data)
+
+    assure_exported_properties(ds)
+    assure_exported_properties(original_ds)
+
+
+@pytest.mark.parametrize("data_format,output_path", DATA_FORMATS_AND_OUTPUT_PATHS)
+def test_add_fs_copy_mag(data_format: DataFormat, output_path: Path) -> None:
```
philippotto, line 1587: there's no test for the symlinking method, right?

philippotto (changes requested):
awesome :+1: I've left some smaller remarks, but the overall state is looking very good :100:

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1054,59 +1089,48 @@ def add_symlink_layer(
...
 
-    def add_copy_layer(
+    def add_fs_copy_layer(
         self,
         foreign_layer: Union[str, Path, Layer],
         new_layer_name: Optional[str] = None,
     ) -> Layer:
         """
-        Copies the data at `foreign_layer` which belongs to another dataset to the current dataset.
+        Copies the files at `foreign_layer` which belongs to another dataset to the current dataset.
```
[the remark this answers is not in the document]
jstriebel: It might also work for remote datasets where dirlisting and writing work, e.g. webdav. I just added `via the filesystem` to make it more clear.

webknossos/tests/test_dataset.py
jstriebel: Yes, there's `test_add_symlink_mag` (where the other tests seem to be copied from without much adaption).

webknossos/tests/test_dataset.py
jstriebel: Tbh I just copied and adapted the previous test. At some point we'll have to refactor `test_dataset.py`, it has grown wild a bit^^ I tried to simplify the two touched test cases for now.

webknossos/tests/test_dataset.py
jstriebel: I think having a different bounding box before running the add_copy_mag makes it a bit more interesting, but also this is not super useful IMO. I'll reduce this to a more ordinary test case.

jstriebel:
@philippotto Thanks for the review! I applied the feedback, could you please have another look? I think the `test_dataset.py` needs some more love in general, I improved now the tests I've changed so far.

philippotto (approved):
perfect, thank you!

PR #794, opened 2022-09-09
Annotation streaming: annotation.get_remote_annotation_dataset()

### LINKED ISSUE #623: Annotation: improve volume layer handling
* [ ] Once https://github.com/scalableminds/webknossos/issues/6127, #688 and https://github.com/scalableminds/webknossos/issues/6193 are in place, we should provide methods to retrieve the (possibly merged) volume layers as a dataset / layer.
* [ ] Add shortcut to create an annotation with a fallback layer directly from a `Layer`.

---------

**Old issue description for reference**

Handling volume layers in annotations should be improved. Some points, from high to lower priority
* allow to merge volume layer into the respective fallback layer
* add shortcut to create an annotation with a fallback layer directly from a `Layer`
* Allow to merge multiple volume annotations into a single one
* Consider a dataset-less layer class to represent and interact with volume annotations directly, but the tmp copy seems fine for this, too.

PS: We should also add tests for annotation download & upload, e.g. a roundtrip test.

jstriebel:
### Description:
- adds `annotation.get_remote_annotation_dataset()` to get a streamed annotation dataset
  (I also went a bit into the direction of `Annotation` subclassing `Dataset`, but that was quite a mess, I'd rather keep it separated for now)
- updates the generated client

### Issues:
- fixes #623

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests
 - [x] Considered adding this to the Examples (@philippotto It would probably be great to update the learned segmenter example to use the new method, I just didn't have the time for it yet)


webknossos/webknossos/annotation/annotation.py
```diff
@@ -434,6 +436,52 @@ def _write_to_zip(self, zipfile: ZipFile) -> None:
...
+            token = context.required_token
+            organization_id = context.organization_id
+        else:
+            token = context.token
+            organization_id = self.organization_id
+            # avoid requiring authentication
+            if token is not None:
+                if organization_id != context.organization_id:
+                    warnings.warn(
+                        "The annotation used in with get_annotation_dataset "
```
philippotto, line 460:
```suggestion
                        "The annotation used with get_annotation_dataset "
```

webknossos/webknossos/annotation/annotation.py
```diff
@@ -434,6 +436,52 @@ def _write_to_zip(self, zipfile: ZipFile) -> None:
...
+        dataset_info_response = dataset_info.sync_detailed(
+            organization_name=organization_id,
+            data_set_name=self.dataset_name,
+            client=context.generated_client,
+        )
+        assert dataset_info_response.status_code == 200, dataset_info_response
+        parsed = dataset_info_response.parsed
+        assert parsed is not None
+
+        # token = sharing_token or wk_context.datastore_token
```
philippotto, line 475: can this be removed? would datastore tokens be necessary in certain scenarios? @fm3 can you answer this? I'm asking since the code currently does not deal with datastore tokens (at least, I don't see it).

webknossos/webknossos/annotation/annotation.py
```diff
@@ -434,6 +436,52 @@ def _write_to_zip(self, zipfile: ZipFile) -> None:
...
+
+        datastore_url = parsed.data_store.url
+
+        zarr_path = UPath(
+            f"{datastore_url}/data/annotations/zarr/{self.annotation_id}/",
```
philippotto, line 480: the entry point is always at the datastore and never at tracing store, right? @fm3 can you confirm this?

webknossos/webknossos/annotation/annotation.py
```diff
@@ -434,6 +436,52 @@ def _write_to_zip(self, zipfile: ZipFile) -> None:
                 layer_content,
             )
 
+    def get_annotation_dataset(self) -> Dataset:
+        """Returns a streamed dataset, incorporating fallback layers and mappings."""
```
philippotto, line 440: how are mappings incorporated? this only applies to agglomerate mappings which are pinned due to volume editing, right?

philippotto (commented):
looks good in general :) I have some small questions, though. @fm3 maybe you could have a look while @jstriebel is on vacation, since you know the zarr internals of the backend :)

webknossos/webknossos/annotation/annotation.py
fm3: All datastore routes now also take the user token, if sent via X-Auth-Token or as get param. The expiring datastore tokens still work as well (and can be sent in the same way) but I discussed with @jstriebel that it’s fine if the python client uses the non-expiring user token, since the user will probably need to insert that into their python code anyway and should be aware of the required secrecy.

webknossos/webknossos/annotation/annotation.py
fm3: Correct. JSON mappings and merger mode are not currently supported for this annotation streaming. May be good to mention that in the comment. It may change in the future though. I do think server-side json mappings may soon be deprecated anyway, though, compare https://scm.slack.com/archives/C5AKLAV0B/p1656524453854519

webknossos/webknossos/annotation/annotation.py
philippotto: ok, then I'll remove that line.

webknossos/webknossos/annotation/annotation.py
philippotto: Thanks for the information, I adapted the docstring now :writing_hand:

philippotto:
> * 
> * [x]  Considered adding this to the Examples (@philippotto It would probably be great to update the learned segmenter example to use the new method, I just didn't have the time for it yet)

I adapted the example now. @fm3 Could you have a look at the latest commit to doublecheck that it makes sense? :pray: 

If I see it correctly, the volume data of an annotation is always downloaded when calling `Annotation.download(url)`, but with the new `get_annotation_dataset` method it wouldn't be necessary to download the actual volume data, right? Are there any plans to avoid this redundant download in the future? (not blocking this PR obviously)

philippotto:
> > * [x]   Considered adding this to the Examples (@philippotto It would probably be great to update the learned segmenter example to use the new method, I just didn't have the time for it yet)
> 
> I adapted the example now. @fm3 Could you have a look at the latest commit to doublecheck that it makes sense? :pray:

Wait a second, the upload in the example doesn't work currently.

fm3:
> Are there any plans to avoid this redundant download in the future?

Note that the wk annotation download route already takes an optional boolean parameter skipVolumeData. This could be useful here

philippotto:
Welcome back from your vacation, @jstriebel! I see that you are already reading here ;) I just noticed a few things which would probably be good to discuss next week (so, enjoy your weekend first):

- Currently, the new method `get_annotation_dataset` sounds (and its docstring reads) like it would work for all annotations. However, the method only works if this annotation stems from an online annotation (maybe we need the concept of an `RemoteAnnotation`?)
- If I download an annotation today (as a zip), open that with wklibs in 3 weeks and use `get_annotation_dataset` to read the volume data, I'll get the newest data. However, maybe I want to read the "old" data? Ideally, the method name should make this clearer.
- It's not clear to me why `get_annotation_dataset` doesn't return a `RemoteDataset`, but instead a `Dataset`. Due to that difference, the upload in the learned segmenter example doesn't work right now (`upload()` expects layers of a `RemoteDataset` in `layers_to_link`).

jstriebel:
@philippotto Thanks a lot for taking care of this PR! I just pushed a small fix for the example, which makes it work now (both in CI and also locally, with the linked layer). I'll try to answer on the different threads:

> If I see it correctly, the volume data of an annotation is always downloaded when calling Annotation.download(url), but with the new get_annotation_dataset method it wouldn't be necessary to download the actual volume data, right? Are there any plans to avoid this redundant download in the future? (not blocking this PR obviously)

Good point, I was wondering if this should be changed as well. I'd say not for now, since the previous methods to get the volume data still exist. I think we might deprecated them, and then switch to download only the NML.

> * Currently, the new method `get_annotation_dataset` sounds (and its docstring reads) like it would work for all annotations. However, the method only works if this annotation stems from an online annotation (maybe we need the concept of an `RemoteAnnotation`?)

Fair point. So far I was assuming that annotations are mostly created online. That might not be the case for skeletons, but I think it holds true for volume annotations, where this would be used mostly. I added a note about it in the docstring, and renamed the method to `get_remote_annotation_dataset`, which makes this slightly more clear.

> * If I download an annotation today (as a zip), open that with wklibs in 3 weeks and use `get_annotation_dataset` to read the volume data, I'll get the newest data. However, maybe I want to read the "old" data? Ideally, the method name should make this clearer.

Good point, how about `get_remote_annotation_dataset`? I think this helps a bit. Also, we might consider to pin the version of an annotation in the future, but I wouldn't add that yet.

> * It's not clear to me why `get_annotation_dataset` doesn't return a `RemoteDataset`, but instead a `Dataset`. Due to that difference, the upload in the learned segmenter example doesn't work right now (`upload()` expects layers of a `RemoteDataset` in `layers_to_link`).

Good question, that's something I was wondering, too. The `RemoteDataset` assumes not only, that the dataset is streamed, but also that it is a proper wk dataset (with tags, owner, etc), which is different from the streamed annotation. A normal dataset can however simply be streamed, by using the `UPath` class instead of a local Path. `layers_to_link` could not work with an annotation layer atm, so it's fine in this case. I fixed it, by simply using `get_remote_base_dataset` for now.

------------

I agree that we should consider something like a `RemoteAnnotation`, but possibly not in this PR yet. With current NML/zip/streamed-layers mix it is not entirely clear, what data is local at which point, and what data would update, when changed in webknossos online. I added a separate issue for it, #803.

philippotto (approved):
Nice one :+1:

PR #796, opened 2022-09-12
Allow --pad arg in auto conversion entry point

### LINKED ISSUE #729: auto-detect conversion should allow --pad
the --pad option is available for convert_image_stack (I think?) but not for the auto-detect converter. Thus, it is not available for the wk-worker

philippotto:
### Description:
- Allow --pad arg in auto conversion entry point

I tested it by running `python -m wkcuber --voxel_size 1,1,1 --name test-pad-out --max_mag 1 --pad test-pad test-pad-out` with a `test-pad` folder as input.

### Issues:
- fixes #729 

### Todos:
 - [x] Updated Changelog


fm3 (approved):
Thanks! Works for me :)

PR #798, opened 2022-09-13
upgrades zarr and numcodecs

normanrz:
### Description:
`zarr` is actually only compatible with `numcodecs>=0.10`, because `numcodecs.compat.ensure_ndarray_like` has been introduced there is is imported by `zarr`. Therefore, we also need to upgrade `numcodecs`.

### Todos:
 - [x] Updated Changelog


fm3:
should we open an issue/pr in the zarr repository for this?

normanrz:
https://github.com/zarr-developers/zarr-python/pull/1135

fm3 (approved):
LGTM :) not sure if this deserves a changelog entry, what do you think?

Issue #748, opened 2022-06-13
Expand add_layer_from_images to cubing script


jstriebel:
After having #632 (PR #741) in place, `add_layer_from_images` is a good start to replace/update the current cubing.py and related scripts.

The following functionality is lacking in comparison to the current cubing.py script:
* z offset (start_z)
* skip_first_z_slices
* allow smaller batch-sizes if not compressing data
* channel_index
* sample_index
* target_mag (downsampling initial layer)
* manual dtype selection,
* missing formats (monkey-patching extensions, adding czi, dm3 and dm4 readers, see below)

Other functionality, that might get into it:
* use typer, add docs-generation
* general offset
* #340
* xy-grid (see `tile_cubing.py`), might also fix #44
* allow multiple layers / image stacks, might fix #285, #426
* add xy-batching (fixes #442)
* default to chunk-size 1 in z when converting 2D images to zarr (#322)

More issues which might get fixed:
#729, #273 (via [pims_nd2](https://github.com/soft-matter/pims_nd2)), #171 (via pims), #51, #39, #289 (via bioformats), #427, #282, #281, #280 (via bioformats), large step for #188

### Missing formats
```python
# current formats we support
cuber = {"tif", "tiff", "jpg", "jpeg", "jpe", "jfif", "jif", "jp2", "j2k", "jpf", "jpm", "jpg2", "j2c", "jpc", "jpx", "mj2", "png", "bmp", "dm3", "dm4", "czi"}
# pims provides supported extensions for the readers
pims_all = {"koala", "g3", "mpeg", "pfm", "ico", "targa", "psd", "ras", "dm3", "seq", "tif", "lsm", "tiff", "pgm", "lei", "pbm", "raw", "ffmpeg", "pgmraw", "cine", "spe", "avbin", "mp4", "jp2", "wbmp", "wmv", "ipw", "fits", "gif", "cut", "exr", "j2k", "pbmraw", "ppm", "npz", "avi", "gdal", "jpg", "iff", "jng", "webp", "ics", "fi_tiff", "xpm", "dicom", "dds", "stk", "itk", "bmp", "mkv", "hdr", "ts", "ipl", "mov", "ids", "jpeg", "lif", "pcx", "pcd", "sgi", "dummy", "swf", "ppmraw", "png", "nd2", "xbm", "pict"}
# this is excluding the bioformats reader extensions:
[...]

PR #808, opened 2022-10-10
Add Dataset.from_images

jstriebel:
### Description:
Adds `Dataset.from_images`, which converts images to a Dataset, possibly consisting of multiple layers. This might supersede the related wkcuber functions in the future.

### Issues:
- contributes to #748

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests
 - [x] Considered adding this to the Examples (not yet, might come later)


webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
...
+        batch_size: Optional[int] = None,
+        executor: Optional[Executor] = None,
+    ) -> "Dataset":
+        """
+        This method imports image data in a folder as a webKnossos dataset. The
+        image data can be 3D images (such as multipage tiffs) or stacks of 2D
+        images. In case of multiple 3D images or image stacks, those are mapped
+        to different layers. The exact mapping is handled by the argument
+        `map_filepath_to_layer_name`, which can be a pre-defined strategy from
+        the enum `ConversionLayerMapping`, or a custom callable which must take
```
fm3, line 539:
```suggestion
        the enum `ConversionLayerMapping`, or a custom callable, taking
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
...
+        a path of an image file and return the corresponding layer name. All
```
fm3, line 540:
```suggestion
        a path of an image file and returning the corresponding layer name. All
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
                 exist_ok=exist_ok,
             )
 
+    @classmethod
+    def from_images(
+        cls,
+        input_path: Union[str, PathLike],
+        output_path: Union[str, PathLike],
+        voxel_size: Optional[Tuple[float, float, float]] = None,
```
fm3, line 513: I did not expect voxel_size to be optional, how does the Dataset class handle this?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
...
+        elif isinstance(map_filepath_to_layer_name, Dataset.ConversionLayerMapping):
+            raise ValueError(
+                f"Got unexpected ConversionLayerMapping value for map_filepath_to_layer_name: {map_filepath_to_layer_name}"
+            )
+
+        ds = cls(output_path, voxel_size=voxel_size, name=name)
+
+        filepaths_per_layer: Dict[str, List[Path]] = {}
+        for input_file in input_files:
+            layername = map_filepath_to_layer_name(input_file)
```
fm3, line 645:
```suggestion
            layer_name = map_filepath_to_layer_name(input_file)
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
...
+                    lambda p: input_upath.name if p.parent == Path() else str(p.parent)
+                )
+        elif isinstance(map_filepath_to_layer_name, Dataset.ConversionLayerMapping):
+            raise ValueError(
+                f"Got unexpected ConversionLayerMapping value for map_filepath_to_layer_name: {map_filepath_to_layer_name}"
+            )
+
+        ds = cls(output_path, voxel_size=voxel_size, name=name)
+
+        filepaths_per_layer: Dict[str, List[Path]] = {}
```
fm3, line 643:
```suggestion
        file_paths_per_layer: Dict[str, List[Path]] = {}
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -121,6 +151,40 @@ class Dataset:
     When using `Dataset.open_remote()` an instance of the `RemoteDataset` subclass is returned.
     """
 
+    @unique
+    class ConversionLayerMapping(Enum):
+        """This specifies mapping strategies which can be used in
```
fm3, line 156:
```suggestion
        """Strategies for mapping file paths to layers, for use in
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -121,6 +151,40 @@ class Dataset:
...
+        `Dataset.from_images` for the `map_filepath_to_layer_name` argument.
+
+        If none of the strategies fit, the mapping can also be specified by a callable.
+        """
+
+        INSPECT_SINGLE_FILE = "inspect_single_file"
+        """The first found image file is opened. If it seems like
+        a 2D image, `ENFORCE_LAYER_PER_FOLDER` is used,
+        if it seems like a 3D image, `ENFORCE_LAYER_PER_FILE`.
+        This is the default mapping."""
```
fm3, line 166:
```suggestion
        """The first found image file is opened. If it appears to be
        a 2D image, `ENFORCE_LAYER_PER_FOLDER` is used,
        if it appears to be 3D, `ENFORCE_LAYER_PER_FILE` is used.
        This is the default mapping."""
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -121,6 +151,40 @@ class Dataset:
...
+
+        INSPECT_EVERY_FILE = "inspect_every_file"
+        """Similar to `INSPECT_SINGLE_FILE`, but the strategy
```
fm3, line 169:
```suggestion
        """Like `INSPECT_SINGLE_FILE`, but the strategy
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -121,6 +151,40 @@ class Dataset:
...
+        is determined for each image file separately."""
+
+        ENFORCE_LAYER_PER_FILE = "enforce_layer_per_file"
+        """Enforce a new layer per file. This is useful for 2D
+        images, which should be converted to 2D layers each."""
```
fm3, line 174:
```suggestion
        images that should be converted to 2D layers each."""
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -106,6 +109,33 @@ def _find_array_info(layer_path: Path) -> Optional[ArrayInfo]:
 )
 
 
+def _guess_if_segmentation_path(filepath: Path) -> bool:
+    return "segmentation" in str(filepath)
+
+
+def _has_image_z_dimension(
```
fm3, line 116: These two functions don’t seem to be too specific to the Dataset class, maybe they could be moved to utils? This file is starting to get pretty long 😅

webknossos/webknossos/dataset/_utils/pims_images.py
```diff
@@ -334,3 +348,23 @@ def dimwise_max(vectors: Sequence[T]) -> T:
...
+
+def _recursive_subclasses(cls: C) -> List[C]:
+    "Return all subclasses (and their subclasses, etc.)."
+    # Source: http://stackoverflow.com/a/3862957/1221924
+    return cls.__subclasses__() + [
+        g for s in cls.__subclasses__() for g in _recursive_subclasses(s)
+    ]
+
+
+def get_all_pims_handlers() -> Iterable[
```
fm3, line 364: Are the returned pims handlers used only for listing the extensions? Maybe return the supported extensions here directly and make this function private?

webknossos/webknossos/dataset/_utils/pims_images.py
```diff
@@ -334,3 +348,23 @@ def dimwise_max(vectors: Sequence[T]) -> T:
...
     else:
         return cast(T, tuple(map(max, *vectors)))
+
+
+C = TypeVar("C", bound=Type)
+
+
+def _recursive_subclasses(cls: C) -> List[C]:
+    "Return all subclasses (and their subclasses, etc.)."
+    # Source: http://stackoverflow.com/a/3862957/1221924
```
fm3, line 358: do we need to worry about licensing?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
...
+        *,
+        map_filepath_to_layer_name: Union[
+            ConversionLayerMapping, Callable[[Path], str]
+        ] = ConversionLayerMapping.INSPECT_SINGLE_FILE,
+        z_slices_sort_key: Callable[[Path], Any] = natsort_keygen(),
+        enforce_category: Optional[LayerCategoryType] = None,
+        data_format: Union[str, DataFormat] = DEFAULT_DATA_FORMAT,
+        chunk_shape: Optional[Union[Vec3IntLike, int]] = None,
+        chunks_per_shard: Optional[Union[int, Vec3IntLike]] = None,
+        compress: bool = False,
```
fm3, line 524: Is there a benefit to not compressing? I would have expected the default to be True

fm3 (changes requested):
Cool stuff! I think the concept of different mapping strategies from file names to layers works really well for this use case. Please see below for some small comments, mostly about naming things :)

Also, this might be something to include in a high-level guide documentation.

Are there file formats that the cuber supported but this doesn’t? I guess Knossos file hierarchy?

webknossos/webknossos/dataset/_utils/pims_images.py
jstriebel: I don't think so, I copied this from pims, which is also licensed permissively. Also, I'd assume this is too simple to be copyrightable :tm:

webknossos/webknossos/dataset/dataset.py
jstriebel: Compression would normally happen afterwards in a separate step, as directly writing compressed chunks requires to write whole shards, which defaults to reading 1024 slices at once instead of 32. If we add a CLI, this would probably run `from_images, compress, downsample` once after the other, but I'd rather not add a separate compression step in this function. Does this make sense?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -441,6 +505,178 @@ def download(
...
+        refer to its documentation.
+
+        For more fine-grained control, please create an empty dataset and use
+        `add_layer_from_images`.
+        """
+        try:
+            from ._utils.pims_images import get_all_pims_handlers
+        except ImportError as e:
+            raise RuntimeError(
+                "Cannot import pims, please install it e.g. using 'webknossos[all]'"
```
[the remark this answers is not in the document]
jstriebel: Yes, this happened twice. I've put the try-except into `pims_images.py` directly, which is more clean IMO :+1:

webknossos/webknossos/dataset/dataset.py
jstriebel: Good point, I think this was only possible in the constructor for backwards-compatibility, when opening existing datasets. Will change here.

webknossos/webknossos/dataset/dataset.py
jstriebel: Then I should also rename `map_filepath_to_layer_name` to `map_file_path_to_layer_name`, right? Also possible, not sure which one I like more…

jstriebel:
@fm3 Thanks a lot for the helpful review!

> Please see below for some small comments, mostly about naming things :)

I applied most of the changes or answered to comments. Please have a look through them, IMO the two open comments are the ones that might need more discussion.

> Also, this might be something to include in a high-level guide documentation.

Good point, I added it to [this table](https://docs.webknossos.org/api/webknossos.html).

> Are there file formats that the cuber supported but this doesn’t? I guess Knossos file hierarchy?

Yes, there's quite a detailed overview in #748. I'll check knossos, czi, dm3 and dm4 support in a follow-up PR. They should probably all be supported via python-only readers, not only via bioformats.

webknossos/webknossos/dataset/dataset.py
fm3: Alright I’ll take that back, apparently there is an ongoing process of including the word filepath into the english language, let’s leave it as you wrote it, then

fm3 (approved):

webknossos/webknossos/dataset/dataset.py
jstriebel: In the latest commit 25d11cb22564e8f3639c00b3f7ae9b68ff5d8287 I added post-hoc compression and downsampling, defaulting to true for both. I also added a small logic to compress directly when converting if the number of slices is small (`<= z-chunk-size`). I'm not sure if that's over-engineered, tbh., and if that actually saves space (I think so, though).
cc @normanrz, What do you think about this?

normanrz: Hmm. I'd err on the side of simplicity: Keep post-compression and downsample out of the function and leave the old compress option in. For convenience, we could add `Dataset.compress` and `Dataset.downsample` to compress and downsample all layers/mags.

jstriebel: Sounds good! I reverted those changes in this PR, `ds.compress` and `ds.downsample` will come in another PR.

PR #814, opened 2022-10-19
Annotation: support empty volumes & make volume location optional

### LINKED ISSUE #810: annotation.temporary_volume_layer_copy() fails for empty volume annotations
## Context

- While consuming downloaded annotations of an annotation project, for some annotations (where the annotator never brushed), the call to `temporary_volume_layer_copy()` failed with a `TypeError`.

## Expected Behavior

The call should succeed, providing an empty segmentation layer. If this is not possible, there should be a method to check whether the layer is empty, before, or the exception should be more clear and documented.

## Current Behavior

A `TypeError: reduce() of empty sequence with no initial value` is thrown.

## Steps to Reproduce the bug

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

Using this annotation I obtained by creating an annotation, setting a node and then downloading it (without brushing):

[annotation_never_brushed.zip](https://github.com/scalableminds/webknossos-libs/files/9767252/annotation_never_brushed.zip)

```
>>> import webknossos as wk
>>> annotation = wk.Annotation.load("annotation_never_brushed.zip")
>>> with annotation.temporary_volume_layer_copy() as seg_layer:
...   print(seg_layer.path)
...
  File "<stdin>", line 1, in <module>
  File "/usr/lib/python3.8/contextlib.py", line 113, in __enter__
    return next(self.gen)
  File "/usr/local/lib/python3.8/dist-packages/webknossos/annotation/annotation.py", line 701, in temporary_volume_layer_copy
    input_annotation_layer = self.export_volume_layer_to_dataset(
  File "/usr/local/lib/python3.8/dist-packages/webknossos/annotation/annotation.py", line 662, in export_volume_layer_to_dataset
    dataset.add_layer_for_existing_files(
  File "/usr/local/lib/python3.8/dist-packages/webknossos/dataset/dataset.py", line 768, in add_layer_for_existing_files
    layer.bounding_box = infer_bounding_box_existing_files(finest_mag_view)
  File "/usr/local/lib/python3.8/dist-packages/webknossos/dataset/_utils/infer_bounding_box_existing_files.py", line 12, in infer_bounding_box_existing_files
    return reduce(
TypeError: reduce() of empty sequence with no initial value
```

## Your Environment for bug

- Operating System and version: Ubuntu 20.04, Python 3.8
- Version of webKnossos-libs (Release or Commit): 0.10.18

jstriebel:
### Description:
This PR
* adds support for empty volumes in temporary_volume_layer_copy() (bug reported in #810)
* make volume locations optional (preparation for https://github.com/scalableminds/webknossos/pull/6566)

### Issues:
- fixes #810

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


fm3 (approved):
LGTM; did not test myself. I assume the get_remote_dataset is now also possible for annotations saved with skip_volume_data? (Note that currently, those are zipped, despite only being a single nml file. In the future, we might opt to not zip those in the backend. would this still be compatible here?)

What is the benefit of changing the BoundingBox staticmethods to class methods? As far as I see it, `cls` is not used in the methods?

jstriebel:
> LGTM; did not test myself. I assume the get_remote_dataset is now also possible for annotations saved with skip_volume_data?

I tested this with the `nml_with_volumes.zip` and `.nml`, this is downloaded from the actual wk PR, and also included in the tests.

> (Note that currently, those are zipped, despite only being a single nml file. In the future, we might opt to not zip those in the backend. would this still be compatible here?)

Both would work now.

> What is the benefit of changing the BoundingBox staticmethods to class methods? As far as I see it, `cls` is not used in the methods?

I should use `cls` for the constructor everywhere, changing this now. This ensures that when calling the method on subclasses (which we don't have atm), an instance of the subclass is created.

PR #815, opened 2022-10-26
Relax tifffile dependency version

fm3:
webknossos and wkcuber had conflicting tifffile dependencies, so installing both with [all] lead to errors.

I edited the tomls and then ran `poetry update --lock tifffile` in both subdirectories. Not sure why the lockfile diff is so huge, I would have expected it to only change the parts concerning tifffile. Do you know what might have happened here?


webknossos/pyproject.toml
```diff
@@ -59,7 +59,7 @@ zipp = "^3.5.0"
 imagecodecs = { version = "2021.11.20", optional = true }
```
jstriebel, line 59: Could you adapt this directly too?

jstriebel (commented):
I assumed that the `imagecodecs` version constraint in wkcuber would also need to be relaxed, I'm wondering how this could even be resolved… Do you also think that this should be changed to ">=2020.5.30"?

fm3:
You’re right, I didn’t notice that wkcuber also depends on imagecodecs. I had not re-resolved the wkcuber deps.
Do you know if the numpy upgrade in the lockfile is ok?

wkcuber/poetry.lock
```diff
@@ -487,17 +487,17 @@ python-versions = ">=3.5"
...
-version = "2020.12.24"
+version = "2021.11.20"
 description = "Image transformation, compression, and decompression codecs"
 category = "main"
 optional = false
 python-versions = ">=3.7"
 
 [package.dependencies]
-numpy = ">=1.15.1"
+numpy = ">=1.16.5"
```
jstriebel, line 497: This just specifies that the selected imagecodecs version has this requirement and is fine :+1:

jstriebel (approved):
LGTM :+1: Thanks for taking care of this :pray: 

> Do you know if the numpy upgrade in the lockfile is ok?

Yes, that's not actually an upgrade of numpy, just the meta-information about it's dependencies changed with the new imagecodecs version.

PR #818, opened 2022-10-28
ds.add_layer_from_images: added topleft and dtype kwargs

jstriebel:
### Description:
- `ds.add_layer_from_images`:
  * Turned some arguments into keyword-only arguments, only affecting positional arguments after the first 8 arguments
  * added topleft and dtype kw-only arguments
- cleaned up the dtype type annotations, using `np.typing.DTypeLike` now if available.

### Issues:
- part of #748

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests


fm3 (commented):
I’d prefer renaming `enforce_dtype` to `dtype`. I think the semantics is clear that if a dtype is passed, it is used, and otherwise it defaults to auto-detect (as do various other things). The name `enforce_dtype` makes me think of a boolean, deciding if something should be enforced or not.
Otherwise, LGTM!

fm3 (approved):
:tada:

PR #819, opened 2022-11-08
Update cattrs to 22.2.0

daniel-wer:
### Description:
- Update cattrs to 22.2.0 to enable and fix `Literal` structuring support in voxelytics. Also needed to make cattrs work for newer python versions, see [changelog](https://cattrs.readthedocs.io/en/stable/history.html).

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/pyproject.toml
+++ b/webknossos/pyproject.toml
@@ -33,7 +33,7 @@ python = ">=3.7.1,<3.10"
 aiohttp = "^3.8.1"
 attrs = "^21.1.0"
 boltons = "~21.0.0"
-cattrs = "1.9.0"
+cattrs = "22.2.0"
 cluster_tools = { path = "../cluster_tools/", develop = true }
 fsspec = "^2022.2.0"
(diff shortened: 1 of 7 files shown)


daniel-wer:
@jstriebel Or would it be better to still allow the older versions? :thinking:

webknossos/pyproject.toml
```diff
@@ -33,7 +33,7 @@ python = ">=3.7.1,<3.10"
 aiohttp = "^3.8.1"
 attrs = "^21.1.0"
 boltons = "~21.0.0"
-cattrs = "22.2.0"
+cattrs = ">=1.7.1,<=22.2.0"
```
jstriebel, line 36: I think specifying the lower end is enough:
```suggestion
cattrs = ">=1.7.1"
```

daniel-wer: Wouldn't that be susceptible to untested new versions and potential breaking changes, though?

jstriebel (approved):
:shipit:

PR #823, opened 2022-11-18
allow wkcuber.tile_cubing to use zero indexed tile locations when using the --input_path_pattern option

thopp-tudelft:
### Description:

removes an assert that failed only when an index was zero

importing with zero indexed paths where eg. `{z}/{x}_{y}.tiff` is in the range of `[1..99]/[0..8]_[0..7].tiff` would fail as the assert would trigger

importing a zero indexed stack of images works fine with this assert removed, there is no reason to require this to not be zero

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/wkcuber/wkcuber/tile_cubing.py
+++ b/wkcuber/wkcuber/tile_cubing.py
@@ -145,7 +145,6 @@ def detect_interval_for_dimensions(
                             idx : idx + current_decimal_length[current_dimension]
                         ]
                         coordinate_value = int(coordinate_value_str)
-                        assert coordinate_value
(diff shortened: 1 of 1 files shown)


jstriebel:
@thopp-tudelft Thank you very much for your contribution! The change looks good. If you sign the CLA (see comment above) we can merge the PR :+1:

jstriebel (approved):

thopp-tudelft:
I need to sign a cla to delete one line? um, sure I guess?

PR #836, opened 2022-12-06
persist bbox ids when serializing NML

### LINKED ISSUE #835: Annotation upload does not preserve user bounding boxes
## Context

While uploading annotations, I noticed that user bounding boxes are not preserved over a round-trip via `webknossos.org`.
I.e. the following script `test.py`:
```python
import webknossos as wk

REMOTE_DATASET_URL = 'https://webknossos.org/datasets/Bao_Lab_Sloan_Kettering_Institute/c_elegans_embryo_475min'

WK_USER = '<User Name>'
WK_TOKEN = '<Auth Token>'
WK_URL = 'https://webknossos.org'


ds = wk.Dataset.open_remote(REMOTE_DATASET_URL)

annotation_up = wk.Annotation(
    name = 'test_bbbox_upload',
    dataset_name = ds.name,
    voxel_size = ds.voxel_size,
    owner_name = WK_USER,
)
# Add data to prevent uploading empty content
group = annotation_up.skeleton.add_group("a group")
tree = group.add_tree("a tree")
node_1 = tree.add_node(position=(0, 0, 0), comment="node 1")

# Create a test bounding box
annotation_up.user_bounding_boxes = [wk.BoundingBox.from_tuple6((1024, 512, 128, 64, 64, 64))]

with wk.webknossos_context(url=WK_URL, token=WK_TOKEN):
    annotation_url = annotation_up.upload()
    annotation_down = wk.Annotation.download(annotation_url)

assert annotation_down.user_bounding_boxes == annotation_up.user_bounding_boxes, \
    f'\n{annotation_down.user_bounding_boxes=}' \
    f'\n{annotation_up.user_bounding_boxes=}'


```
results in assertion error:
```Python
  File "test.py", line 27, in <module>
    assert annotation_down.user_bounding_boxes == annotation_up.user_bounding_boxes, \
AssertionError:
annotation_down.user_bounding_boxes=[]
annotation_up.user_bounding_boxes=[BoundingBox(topleft=(1024, 512, 128), size=(64, 64, 64))]
```

## Expected Behavior

I would expect preserved user bounding boxes for a round-trip via `webknossos.org`.

## Current Behavior
Instead I loose the hard-coded bounding boxes.

## Steps to Reproduce the bug
```cmd
conda env create webknossos python=3.9
conda activate webknossos
pip install webknossos
python .\test.py
```

## Your Environment for bug

- Operating System and version: Windows 10
- Version of webKnossos-libs (Release or Commit): webknossos==0.10.25

jstriebel:
### Description:
Bounding Box ids are optional in the code, but need to be persisted with NMLs so that the bboxes are parsed correctly when being uploaded. This adds default ids when persisting nmls, and removes id from the BoundingBox class.

### Issues:
- fixes #835

### Todos:
 - [x] Updated Changelog
 - [x] Added Tests

### DIFF
--- a/webknossos/webknossos/_nml/parameters.py
+++ b/webknossos/webknossos/_nml/parameters.py
@@ -26,72 +26,54 @@ class Parameters(NamedTuple):
     taskBoundingBox: Optional[BoundingBox] = None
     userBoundingBoxes: Optional[List[BoundingBox]] = None
 
-    def _get_max_bbox_id(self) -> int:
-        bboxes = []
-        if self.taskBoundingBox is not None:
-            bboxes.append(self.taskBoundingBox)
-        if self.userBoundingBoxes is not None:
-            bboxes += self.userBoundingBoxes
-        ids = []
-        for bbox in bboxes:
-            if bbox.id is not None:
-                try:
-                    ids.append(int(bbox.id))
-                except ValueError:
-                    pass
-        return max(ids, default=0)
-
     def _dump_bounding_box(
         self,
         xf: XmlWriter,
         bounding_box: BoundingBox,
         tag_name: Text,
-        next_bbox_id: int,
+        id: Optional[int],
     ) -> None:
 
         color = bounding_box.color or DEFAULT_BOUNDING_BOX_COLOR
 
+        attributes = {
+            "name": _DEFAULT_BBOX_NAME
+            if bounding_box.name is None
+            else str(bounding_box.name),
+            "isVisible": "true" if bounding_box.is_visible else "false",
+            "color.r": str(color[0]),
+            "color.g": str(color[1]),
+            "color.b": str(color[2]),
+            "color.a": str(color[3]),
+            "topLeftX": str(bounding_box.topleft.x),
+            "topLeftY": str(bounding_box.topleft.y),
+            "topLeftZ": str(bounding_box.topleft.z),
+            "width": str(bounding_box.size.x),
+            "height": str(bounding_box.size.y),
+            "depth": str(bounding_box.size.z),
+        }
+        if id is not None:
+            attributes["id"] = str(id)
+
         xf.tag(
             tag_name,
-            {
-                "id": next_bbox_id if bounding_box.id is None else str(bounding_box.id),
-                "name": _DEFAULT_BBOX_NAME
-                if bounding_box.name is None
-                else str(bounding_box.name),
-                "isVisible": "true" if bounding_box.is_visible else "false",
-                "color.r": str(color[0]),
-                "color.g": str(color[1]),
-                "color.b": str(color[2]),
-                "color.a": str(color[3]),
-                "topLeftX": str(bounding_box.topleft.x),
-                "topLeftY": str(bounding_box.topleft.y),
-                "topLeftZ": str(bounding_box.topleft.z),
-                "width": str(bounding_box.size.x),
-                "height": str(bounding_box.size.y),
-                "depth": str(bounding_box.size.z),
-            },
+            attributes,
         )
 
     def _dump_task_bounding_box(self, xf: XmlWriter) -> None:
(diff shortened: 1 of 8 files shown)


webknossos/webknossos/geometry/bounding_box.py
```diff
@@ -31,7 +33,7 @@ class BoundingBox:
     topleft: Vec3Int = attr.field(converter=Vec3Int)
     size: Vec3Int = attr.field(converter=Vec3Int)
     bottomright: Vec3Int = attr.field(init=False)
-    name: Optional[str] = "Unnamed Bounding Box"
+    name: Optional[str] = _DEFAULT_BBOX_NAME
     is_visible: bool = True
     id: Optional[str] = None
```
jstriebel, line 38: Maybe the id should rather be an int?

webknossos/webknossos/_nml/parameters.py
```diff
@@ -53,15 +75,23 @@ def _dump_bounding_box(
     def _dump_task_bounding_box(self, xf: XmlWriter) -> None:
         task_bounding_box = getattr(self, "taskBoundingBox")
         if task_bounding_box is not None:
-            self._dump_bounding_box(xf, task_bounding_box, "taskBoundingBox")
+            next_bbox_id = self._get_max_bbox_id() + 1
+            self._dump_bounding_box(
+                xf, task_bounding_box, "taskBoundingBox", next_bbox_id
+            )
```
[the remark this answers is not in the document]
jstriebel: Do you know if the id is used for the order of the bounding boxes, or rather the order in the xml file? I'm not quite sure why we have them anyways :shrug:

webknossos/webknossos/geometry/bounding_box.py
fm3: It’s int32 in the proto definition so int sounds right to me

webknossos/webknossos/_nml/parameters.py
```diff
@@ -53,15 +75,23 @@ def _dump_bounding_box(
...
 
     def _dump_user_bounding_boxes(self, xf: XmlWriter) -> None:
         user_bounding_boxes = getattr(self, "userBoundingBoxes")
 
         if user_bounding_boxes is not None:
+            next_bbox_id = self._get_max_bbox_id() + 2
```
philippotto, line 87: why + 2 and not + 1?

webknossos/webknossos/_nml/parameters.py
```diff
@@ -25,17 +26,38 @@ class Parameters(NamedTuple):
     taskBoundingBox: Optional[BoundingBox] = None
     userBoundingBoxes: Optional[List[BoundingBox]] = None
 
+    def _get_max_bbox_id(self) -> int:
```
philippotto, line 29: Could you add a docstring which explains what the maximum bbox id is (e.g, that only integer ids are considered), as the semantics weren't immediately clear to me because the ids can be strings too apparently.

webknossos/webknossos/_nml/parameters.py
```diff
@@ -25,17 +26,38 @@ class Parameters(NamedTuple):
...
+        bboxes = []
+        if self.taskBoundingBox is not None:
+            bboxes.append(self.taskBoundingBox)
+        if self.userBoundingBoxes is not None:
+            bboxes += self.userBoundingBoxes
```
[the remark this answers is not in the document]
jstriebel: They are not, `+=` adds another list (the inplace equivalent would be `bboxes.update(self.userBoundingBoxes)`, whereas `append` only adds a single element.

jstriebel:
@fm3 @philippotto I removed the `id` parameter for the `BoundingBox` class as discussed, and how name, color and is_visible are handled through methods. Also added a test for the upload-download roundtrip.

PR #838, opened 2022-12-06
Cancel cluster jobs on shutdown

daniel-wer:
### Description:
- Try to cancel the submitted cluster jobs when receiving the SIGINT signal so if the user decided to abort the execution
- Also don't interfere with the shutdown procedure of the main process by ignoring SIGTERM or calling sys.exit

### Issues:
- contributes to https://github.com/scalableminds/voxelytics/issues/3127

### Todos:
 - [x] Updated Changelog


cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -200,6 +200,25 @@ def submit_text(cls, job, cfut_dir):
     def handle_kill(self, *args, **kwargs):
         for submit_thread in self.submit_threads:
             submit_thread.stop()
+
+        # Jobs with a "pending" state have not been submitted to the cluster yet
+        scheduled_job_ids = [
+            job_id for job_id, job_state in self.jobs.items() if job_state != "pending"
```
philippotto, line 206: the pending-check covers this case here, right?

```
                # Register the job in the jobs array, although the jobid is not known yet.
                # Otherwise it might happen that self.jobs becomes empty, but some of the jobs were
                # not even submitted yet.
                self.jobs[workerid_with_index] = "pending"
```

can you rename "pending" in both places to something else so that it cannot be confused with slurm's pending state. this caught me off guard for a moment. how about "unsubmitted" or "not_yet_submitted"? could be in all caps, too.

cluster_tools/cluster_tools/schedulers/pbs.py
```diff
@@ -53,6 +53,29 @@ def format_log_file_name(job_id_with_index, suffix=".stdout"):
     def get_job_id_string(cls):
         return cls.get_current_job_id()
 
+    def handle_kill(self, *args, **kwargs):
```
philippotto, line 56: did you test this (pbs) method?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -93,7 +93,6 @@ def __init__(
         os.makedirs(self.cfut_dir, exist_ok=True)
 
         signal.signal(signal.SIGINT, self.handle_kill)
-        signal.signal(signal.SIGTERM, self.handle_kill)
```
philippotto, line 96: maybe add a comment explaining why sigint is handled and sigterm isn't?

philippotto (commented):
Cool stuff :) Looks pretty good already. Only left some small comments.

cluster_tools/cluster_tools/schedulers/pbs.py
daniel-wer: No, I only read the (very good!) docs. Do you know where this could be tested?

cluster_tools/cluster_tools/schedulers/slurm.py
daniel-wer: Yes, that's right, Good point, I'll rename to `NOT_YET_SUBMITTED`.

cluster_tools/cluster_tools/schedulers/pbs.py
philippotto: I just looked up how I originally tested the PBS support when I integrated it. Apparently, I tested it on a production instance of a client for which this was developed. Theoretically, you could try to set up PBS with docker (like it's done for slurm), but I don't think it's necessary. The PBS support is untested now for a long time and we might need to invest some effort into it regardless of this feature once we decide to use it again. So, I'd defer the effort until it's really needed :)

daniel-wer: Agreed, thanks for looking up these details!

philippotto (approved):
Awesome, works great :+1:

daniel-wer:
One more thing just came to my mind. I don't think this will work for jobs that schedule jobs themselves.

What could work for that case would be to send a SIGINT to the respective jobs, first (possible using [scancel](https://slurm.schedmd.com/scancel.html) -s) and then execute scancel normally, shortly afterwards.

philippotto:
Very good point!

> What could work for that case would be to send a SIGINT to the respective jobs, first (possible using [scancel](https://slurm.schedmd.com/scancel.html) -s) and then execute scancel normally, shortly afterwards.

Yes, this sounds like a good option. Do you think it's necessary two do 2 passes (friendly termination request with SIGINT and shortly afterwards the bazooka)? Since the previous clustertools didn't terminate those jobs, I think, it could be acceptable to do only the SIGINT way. But maybe doing both isn't too much work.

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -503,7 +511,7 @@ def shutdown(self, wait=True):
         self.was_requested_to_shutdown = True
         if wait:
             with self.jobs_lock:
-                if self.jobs:
+                if self.jobs and self.wait_thread.is_alive():
```
philippotto, line 514: nice find!

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -210,10 +210,14 @@ def inner_handle_kill(self, *args, **kwargs):
...
-            # Array jobs (whose id looks like `<job_id>_<array_index>`) don't need to be canceled individually,
-            # but can be canceled together using the job_id
+            # Array jobs (whose id looks like `<job_id>_<array_index>`) don't need to be signaled individually,
+            # but can be canceled together using the job_id.
             unique_job_ids = set(map(lambda x: str(x).split("_")[0], scheduled_job_ids))
-            _stdout, stderr, exit_code = call(f"scancel {' '.join(unique_job_ids)}")
+            # Send SIGINT signal to running jobs instead of terminating the jobs right away. This way, the jobs can
+            # react to the signal, safely shutdown and signal (cancel) jobs they possibly scheduled, recursively.
+            _stdout, stderr, exit_code = call(
+                f"scancel --state=PENDING {' '.join(unique_job_ids)}; scancel -s SIGINT --state=RUNNING {' '.join(unique_job_ids)}"
```
philippotto, line 219: According to the docs of [`scancel`](https://slurm.schedmd.com/scancel.html) there can also be `SUSPENDED` jobs. Maybe also send the default signal to these jobs?

philippotto: We discussed this already for the not-recursive case, but I want to doublecheck again: If a job doesn't react to SIGINT, there is no follow-up action taken here, right? Since slurm jobs were not canceled automatically at all in the past, it's probably okay to go with this behavior (let's see whether it occurs at all), right?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -96,9 +96,11 @@ def __init__(
         os.makedirs(self.cfut_dir, exist_ok=True)
 
         # Clean up if a SIGINT signal is received. However, do not interfere with the
+        # existing signal handler of the process or the
         # shutdown of the main process which sends SIGTERM signals to terminate all
         # child processes.
-        signal.signal(signal.SIGINT, self.handle_kill)
+        existing_sigint_handler = signal.getsignal(signal.SIGINT)
+        signal.signal(signal.SIGINT, partial(self.handle_kill, existing_sigint_handler))
```
philippotto, line 103: good solution :+1:

philippotto (approved):
great! I only have one tidbit regarding `suspended` jobs.

cluster_tools/cluster_tools/schedulers/slurm.py
daniel-wer: I'd say yes to both. Generally, slurm jobs are like separate voxelytics instances. So if the shutdown procedure works for the sequential and multiprocessing strategies, and the SIGINT signal propagation to the slurm job works, the shutdown procedure for slurm jobs should work as well.

daniel-wer:
I spent quite a while debugging the failing test and encountered very flaky behavior of `scancel` in combination with the `--state` flag. Often, pending tasks were not canceled, although they were marked as pending before and after the scancel command when checking with `squeue`. The same sometimes happened for running tasks. I did not find any mention of this issue online.

Finally, I checked the history of the scancel code in the public git repo and found this: https://github.com/SchedMD/slurm/commit/93b8f2395a09db5b8124f02e779e1182ea930da0

which seems to refer to a private bug report. I assume, before this fix, the state supplied with `--state` was compared to one of the many detailed job states (as contained in `SLURM_STATES`, for example CONFIGURING) whereas after the fix, it is correctly compared to the job's base state (which hopefully always is one of PENDING, RUNNING, SUSPENDED). Indeed, the version of slurm in the docker container is `19.05.1-2` while the other Slurm cluster where I tested this code successfully has version `21.08.8-2`. I'll update the version of slurm in the docker container and retest, tomorrow.

daniel-wer:
I updated the slurm version which fixed the issue :) Still, in the unit tests I discovered one remaining issue where jobs are not properly canceled if they just switched from pending to running. I assume this might be, because Python is not started yet and, therefore, cannot react to the SIGINT (otherwise the SIGINT would cause a `KeyboardInterrupt`, terminating the python program, even if our custom handler was not registered yet). Since I can only think of complicated solutions for this issue, it should be rather rare, and before this PR slurm jobs were not canceled at all, I would propose to leave it at that for now and monitor whether this occurs often in production.

philippotto (approved):
Thank you for pulling this to the finish line :100: The code looks good, I only left a nitpick which you can ignore if you want.

> I assume this might be, because Python is not started yet and, therefore, cannot react to the SIGINT (otherwise the SIGINT would cause a KeyboardInterrupt, terminating the python program, even if our custom handler was not registered yet). Since I can only think of complicated solutions for this issue, it should be rather rare, and before this PR slurm jobs were not canceled at all, I would propose to leave it at that for now and monitor whether this occurs often in production.

Your theory sounds plausible. I think a SIGINT retry could make sense for that scenario, but I agree that we should monitor this in production first.

daniel-wer:
> I think a SIGINT retry could make sense for that scenario, but I agree that we should monitor this in production first.

That's a nice idea I haven't thought of. I'll keep that in mind for when we encounter the issue in production.

PR #858, opened 2023-01-26
[WIP] Typed cluster tools

jstriebel:
### Description:
- abc

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/cluster_tools/cluster_tools/__init__.py
+++ b/cluster_tools/cluster_tools/__init__.py
@@ -10,292 +10,18 @@ from shutil import rmtree
 from typing import Union
 
 from . import pickling
-from .multiprocessing_logging_handler import get_multiprocessing_logging_setup_fn
+from .executors.debug_sequential import DebugSequentialExecutor
+from .executors.multiprocessing import MultiprocessingExecutor
+from .executors.sequential import SequentialExecutor
+from .executors.pickle import PickleExecutor
 from .schedulers.cluster_executor import ClusterExecutor, RemoteOutOfMemoryException
 from .schedulers.kube import KubernetesExecutor
 from .schedulers.pbs import PBSExecutor
 from .schedulers.slurm import SlurmExecutor
 from .util import enrich_future_with_uncaught_warning
 
-
-def get_existent_kwargs_subset(whitelist, kwargs):
-    new_kwargs = {}
-    for arg_name in whitelist:
-        if arg_name in kwargs:
-            new_kwargs[arg_name] = kwargs[arg_name]
-
-    return new_kwargs
-
-
-PROCESS_POOL_KWARGS_WHITELIST = ["max_workers", "initializer", "initargs"]
-
-
-class WrappedProcessPoolExecutor(ProcessPoolExecutor):
-    """
-    Wraps the ProcessPoolExecutor to add various features:
-    - map_to_futures and map_unordered method
-    - pickling of job's output (see output_pickle_path_getter and output_pickle_path)
-    - job submission via pickling to circumvent bug in python < 3.8 (see MULTIPROCESSING_VIA_IO_TMP_DIR)
-    """
-
-    def __init__(self, **kwargs):
-        assert (not "start_method" in kwargs or kwargs["start_method"] is None) or (
-            not "mp_context" in kwargs
-        ), "Cannot use both `start_method` and `mp_context` kwargs."
-
-        new_kwargs = get_existent_kwargs_subset(PROCESS_POOL_KWARGS_WHITELIST, kwargs)
-
-        mp_context = None
-
-        if "mp_context" in kwargs:
-            mp_context = kwargs["mp_context"]
-        elif "start_method" in kwargs and kwargs["start_method"] is not None:
-            mp_context = multiprocessing.get_context(kwargs["start_method"])
-        elif "MULTIPROCESSING_DEFAULT_START_METHOD" in os.environ:
-            mp_context = multiprocessing.get_context(
-                os.environ["MULTIPROCESSING_DEFAULT_START_METHOD"]
-            )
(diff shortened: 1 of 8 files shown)


cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -111,7 +151,7 @@ def get_max_array_size():
             "scontrol show config | sed -n '/^MaxArraySize/s/.*= *//p'"
         )
         if exit_code == 0:
-            max_array_size = int(stdout.decode("utf8"))
```
philippotto, line 114: this wasn't necessary?

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -483,4 +525,4 @@ def run(self):
                     return
 
             job_id = SlurmExecutor.submit_text(script, self.cfut_dir)
-            future.set_result(job_id)
+            future.set_result(str(job_id))
```
philippotto, line 528: How about changing the signature of `submit_text` to return the job id as a string? String based job IDs generalize better over different cluster environments, anyway.

cluster_tools/cluster_tools/_utils/string.py
```diff
@@ -0,0 +1,22 @@
+import os
+import random
+import string
+
+
+def local_filename(filename: str = "") -> str:
+    return os.path.join(os.getenv("CFUT_DIR", ".cfut"), filename)
+
+
+# Instantiate a dedicate generator to avoid being dependent on
```
philippotto, line 10:
```suggestion
# Instantiate a dedicated generator to avoid being dependent on
```

cluster_tools/cluster_tools/_utils/meta.py
```diff
@@ -0,0 +1,34 @@
+import os
```
philippotto, line 1: why `meta.py`? I think, I'd find `misc.py` less confusing.

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -492,8 +545,7 @@ def register_jobs(
                 with_preliminary_postfix(output_path), jobid_with_index
             )
 
-            fut.cluster_jobid = jobid
-            fut.cluster_jobindex = array_index
```
philippotto, line 496: I think this was there for debugging purposes? Probably won't hurt to keep it?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -525,50 +581,57 @@ def shutdown(self, wait=True):
                 pass
         self.files_to_clean_up = []
 
-    def map(self, func, args, timeout=None, chunksize=None):
+    # TODO: args should be *iterables, this would be a breaking change!
```
philippotto, line 584: todo. whats your plan here? :)

philippotto (changes requested):
Awesome, looks already very good :+1:

cluster_tools/cluster_tools/schedulers/slurm.py
jstriebel: I refactored the `call` function above to do this instead, since it was decoded for all usages.

cluster_tools/cluster_tools/_utils/meta.py
jstriebel: I meant `meta` metaprogramming. Might also call it `metaprogramming` therefore, or do you prefer misc?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
jstriebel: Sure, I just got type errors for it and didn't find any usages. I'll re-add it, type-ignore it and add some comments.

cluster_tools/cluster_tools/schedulers/cluster_executor.py
jstriebel: This is part of the follow-up points mentioned in the description of this PR. I'd still leave the TODO in the code if you agree.

cluster_tools/cluster_tools/_utils/meta.py
philippotto: Ah, I see, yes, metaprogramming would be clearer. However, I think, I'd prefer `reflection` in this case, as it fits even, (I think?) :)

philippotto:
@jstriebel Did you push your latest changes? :)

jstriebel:
> @jstriebel Did you push your latest changes? :)

Needed to fetch first, now :see_no_evil:

philippotto (approved):
Awesome :+1: Code looks good :+1:

Two things to clarify, though:
1. Did you test it within vx? I think it would be important to test the integration with the vx executor.
2. Are the follow-up todos from the PR description in an issue somewhere?

jstriebel:
> * Did you test it within vx? I think it would be important to test the integration with the vx executor.

Nope, this would be easier with an actual release. I'd rather merge this, create a release and directly test this with vx afterwards myself, doing another round if necessary. Ok?

> * Are the follow-up todos from the PR description in an issue somewhere?

Just created #859.

philippotto:
> > * Did you test it within vx? I think it would be important to test the integration with the vx executor.
> 
> Nope, this would be easier with an actual release. I'd rather merge this, create a release and directly test this with vx afterwards myself, doing another round if necessary. Ok?

Ok, makes sense :+1: 

> > * Are the follow-up todos from the PR description in an issue somewhere?
> 
> Just created #859.

Great :100:

PR #862, opened 2023-02-07
adds example for downloading a tiff stack

normanrz:
### Description:
- Adds an example for how to download data as a TIFF stack

### Todos:
 - [ ] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests


webknossos/examples/download_tiff_stack.py
```diff
@@ -0,0 +1,30 @@
...
+MAG = wk.Mag("4-4-1")
+
+
+def main() -> None:
+    dataset = wk.Dataset.open_remote(DATASET_NAME, organization_id="scalable_minds")
+    mag_view = dataset.get_layer(LAYER_NAME).get_mag(MAG)
+
+    z = mag_view.layer.bounding_box.topleft.z
+    with mag_view.get_buffered_slice_reader() as reader:
+        for slice_data in reader:
```
jstriebel, line 20: I'd rather have a progress bar instead of the print statement. E.g. using `from rich.progress import track`:

```suggestion
        for slice_data in track(reader):
```

jstriebel (approved):
LGTM, just added some minor comments :+1:

webknossos/examples/download_tiff_stack.py
normanrz: I'd rather like to keep it simple. These examples are meant for Python novices to adapt. If we add `track`, I think that is contra-productive. It would be different if it were part of the library.

PR #864, opened 2023-02-08
Fix slurm execution

daniel-wer:
### Description:
- In https://github.com/scalableminds/webknossos-libs/pull/858 an assertion was added to the `get_job_array_id` method of the slurm executor, asserting that the `SLURM_ARRAY_JOB_ID` environment variable is always set. However, it is not set for non-array jobs which triggers the assertion.

### Issues:
- I successfully tested this PR using the workflow which I ran when discovering the bug.
- I'm not sure why this wasn't triggered in the slurm tests :thinking: 

### Todos:
 - [x] Updated Changelog
 - [ ] Added / Updated Tests

### DIFF
--- a/cluster_tools/Changelog.md
+++ b/cluster_tools/Changelog.md
@@ -16,6 +16,7 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Changed
 
 ### Fixed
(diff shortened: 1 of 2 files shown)


jstriebel (approved):
Thanks :+1:

PR #865, opened 2023-02-13
Pims normalize extensions for assertion

fm3:
I got
`The following suffixes are supported: ['.czi', '.dm3', '.dm4', 'avbin', 'avi', 'bmp', 'cine', 'cut', 'dds',…]` and noticed that in that line, `'.czi'` still has a leading `.`, breaking the assertion despite my data having the right extension.

 - [x] Updated Changelog

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -20,6 +20,8 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 
 ### Fixed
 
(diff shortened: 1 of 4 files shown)


fm3:
@jstriebel do you know if this might cause trouble in other places as well?

jstriebel (approved):
Thanks for investigating this, @fm3! I directly fixed this as well where the strings come from, but the additional normalization LGTM :+1:

PR #867, opened 2023-02-14
No equal sign in generated layer names

fm3:
### Description:
- Remove the equal sign from generated layer names for better url safety

### Todos:
 - [x] Updated Changelog


jstriebel (approved):
LGTM

PR #876, opened 2023-02-21
Handle axis offsets in czi files

fm3:
### Description:
- we got some czi files where the axis start was not at 0
- this pr adapts the czi reader to read the axis offset from the metadata, and adds this offset when reading at a position.

### Todos:
 - [x] Updated Changelog


jstriebel (approved):
:tada:

PR #877, opened 2023-02-21
add reader for imagej tiffs and auto-conversion for big endian dtypes

### LINKED ISSUE #869: add_layer_from_images: autoconvert big endian dtypes
## Detailed Description
When importing big endian images, currently one has to manually set the dtype (for `add_layer_from_images`, it's not possible for `Dataset.from_images`). Instead, it could be autodetected that a conversion is necessary when nothing is set manually. This would also be helpful for the wk-worker imports. See [this dataset](https://scm.slack.com/archives/C02Q35Q28P5/p1676472393336519?thread_ts=1676468774.644139&cid=C02Q35Q28P5) for an example (needs `use_bioformats=True` atm).

### LINKED ISSUE #870: pims_reader for imagej tifffiles with z-slices
## Detailed Description
ImageJ writes tifffiles where z-slices are not represented as pages, but via a different memory-layout. The current [`TiffStack_tifffile` reader](https://github.com/soft-matter/pims/blob/main/pims/tiff_stack.py) handles those only as a single 2D image (as other tools do as well). However, they can be recognized and read by using:
```python
import tifffile

path = "…"
a = tifffile.TiffFile(path)
a.is_imagej  # True
a.imagej_metadata  # contains metadata

tifffile.memmap(path).shape  # correct shape, memmapped np array
```

Therefore we could add a custom reader for ImageJ tiff-stacks. Alternatively, we could recognize those and fall back to the bioformats reader, which supports those already.

jstriebel:
### Description:
- added a reader for imagej tifffiles, supporting hyperstacks
- added auto-conversion from big endian dtype to little endian if nothing is specified manually

I tested this locally with [those files](https://scm.slack.com/archives/C5AKLAV0B/p1676539393839729?thread_ts=1676475091.980849&cid=C5AKLAV0B), which resulted in [this dataset](https://webknossos.org/datasets/scalable_minds/test-ds). I couldn't find public test data.

I'll also verify this with [this dataset](https://scm.slack.com/archives/C02Q35Q28P5/p1676472393336519?thread_ts=1676468774.644139&cid=C02Q35Q28P5) when deployed, to be sure.

### Issues:
- fixes #869
- fixes #870

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/_utils/pims_imagej_tiff_reader.py
+++ b/webknossos/webknossos/dataset/_utils/pims_imagej_tiff_reader.py
@@ -1,0 +1,57 @@
+from os import PathLike
+from pathlib import Path
+from typing import Set
+
+import numpy as np
+
+try:
+    from pims import FramesSequenceND
+except ImportError as e:
+    raise RuntimeError(
+        "Cannot import pims, please install it e.g. using 'webknossos[all]'"
+    ) from e
+
+try:
+    import tifffile
+except ImportError as e:
+    raise RuntimeError(
+        "Cannot import tifffile, please install it e.g. using 'webknossos[tifffile]'"
+    ) from e
+
+
+class PimsImagejTiffReader(FramesSequenceND):
+    @classmethod
+    def class_exts(cls) -> Set[str]:
+        return {"tif", "tiff"}
+
+    # class_priority is used in pims to pick the reader with the highest priority.
+    # Default is 10, and bioformats priority is 2.
+    # See http://soft-matter.github.io/pims/v0.6.1/custom_readers.html#plugging-into-pims-s-open-function
+    class_priority = 20
+
+    def __init__(self, path: PathLike) -> None:
+        super().__init__()
+        path = Path(path)
+        tiff = tifffile.TiffFile(path)
+        assert tiff.is_imagej, f"{path} is not an ImageJ Tiff"
+        channels = tiff.imagej_metadata["channels"]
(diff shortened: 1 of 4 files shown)


webknossos/webknossos/dataset/_utils/pims_imagej_tiff_reader.py
```diff
@@ -0,0 +1,57 @@
...
+    # class_priority is used in pims to pick the reader with the highest priority.
+    # Default is 10, and bioformats priority is 2.
+    # See http://soft-matter.github.io/pims/v0.6.1/custom_readers.html#plugging-into-pims-s-open-function
+    class_priority = 20
+
+    def __init__(self, path: PathLike) -> None:
+        super().__init__()
+        path = Path(path)
+        tiff = tifffile.TiffFile(path)
+        assert tiff.is_imagej, f"{path} is not an ImageJ Tiff"
```
normanrz, line 36: Will pims use another reader, if this assertion fails?

normanrz (approved):

webknossos/webknossos/dataset/_utils/pims_imagej_tiff_reader.py
jstriebel: Yes, as long as `__init__` fails, the next compatible reader is used.
See
https://github.com/soft-matter/pims/blob/main/pims/api.py#L199-L205

PR #884, opened 2023-03-01
Fix skeleton doc example

erjel:
### Description:
- The "simple example" in the skeleton documentation is broken. I copy/pasted it to the tests, fixed the problem and updated the doc accordingly.

### Todos:

 - [x] Updated Documentation
 - [x] Added / Updated Tests


normanrz (approved):
Thanks!

PR #891, opened 2023-03-28
[WIP] Add remove_tree and overload add_tree

philippotto:
### Description:
- abc

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/skeleton/group.py
+++ b/webknossos/webknossos/skeleton/group.py
@@ -1,3 +1,4 @@
+import copy
 from typing import TYPE_CHECKING, Iterator, Optional, Set, Tuple, Union, cast
 
 import attr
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/skeleton/tree.py
```diff
@@ -136,8 +136,8 @@ def __init__(
         self.node_dict_factory = _NodeDict
         # The lambda works because self._node is set before self._adj in networkx.Graph.__init__
         # and because the lambda is evaluated lazily.
-        self.adjlist_outer_dict_factory = lambda: _AdjDict(node_dict=self._node)
-        self.adjlist_inner_dict_factory = lambda: _AdjDict(node_dict=self._node)
+        # self.adjlist_outer_dict_factory = lambda: _AdjDict(node_dict=self._node)
+        # self.adjlist_inner_dict_factory = lambda: _AdjDict(node_dict=self._node)
```
philippotto, line 140: remove?

webknossos/webknossos/skeleton/group.py
```diff
@@ -63,25 +64,43 @@ def id(self) -> int:
 
     def add_tree(
         self,
-        name: str,
+        name_or_tree: Union[str, Tree],
```
philippotto, line 67: strictly speaking this is a breaking change, because code that used `add_tree(name=...)` won't work now, anymore. However, I couldn't find such code in the SCM organization (since it's the first parameter this seems unusual, anyway). So, I'd think, it's fine to do this change here. Do you agree?

daniel-wer (approved):
LGTM :+1:

PR #905, opened 2023-05-23
Added new WK tutorials to navbar and replaced video embedding plugin for docs

hotzenklotz:
### Description:
- PR adds new WK tutorials to navbar and replaced video embedding plugin for docs

# ONLY MERGE TOGETHER WITH https://github.com/scalableminds/webknossos/pull/7095

### Issues:
- None

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


docs/poetry.lock
daniel-wer, line 1: I'm not sure whether it's a good idea to use poetry 1.4.2, here. So far, I think we have stuck to the version pinned in voxelytics which is 1.1.14. For example, I'm having trouble generating the docs ("Upgrade Poetry to ensure the lock file is read properly"). I could imagine different poetry versions could also lead to problems when developing webknossos-libs in combination with voxelytics, locally.
Unless you are sure about what you're doing, I'd propose to revert the changes to the lock file and regenerate it again with poetry 1.1.14 :thinking:

webknossos/poetry.lock
daniel-wer, line 1: Looks like these changes are not needed for this PR and were included by accident? If so, I'd revert them.

hotzenklotz: I reverted and re-ran with Poetry 1.1.14

docs/poetry.lock
daniel-wer: Thanks, I'm not sure why poetry removed all the links to the wheels :shrug:

daniel-wer (approved):
LGTM :+1:

PR #915, opened 2023-06-19
fixes a bug with upsampling a dataset that has a non-aligned bounding…

normanrz:
### Description:
- Fixes a bug where upsampling of a layer would fail, if the layer had a bounding box that doesn't align with the `from_mag` mag.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/tests/dataset/test_upsampling.py
+++ b/webknossos/tests/dataset/test_upsampling.py
@@ -2,7 +2,14 @@ from pathlib import Path
 
 import numpy as np
 
-from webknossos import COLOR_CATEGORY, Dataset, Mag, Vec3Int
+from webknossos import (
+    COLOR_CATEGORY,
+    SEGMENTATION_CATEGORY,
+    BoundingBox,
+    Dataset,
+    Mag,
+    Vec3Int,
+)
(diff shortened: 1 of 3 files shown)


webknossos/tests/dataset/test_upsampling.py
```diff
@@ -125,3 +132,19 @@ def test_upsample_multi_channel(tmp_path: Path) -> None:
...
+    )
+    l.bounding_box = BoundingBox(topleft=(0, 0, 0), size=(8409, 10267, 5271))
+    l.add_mag(32)
+
+    l.upsample(
+        from_mag=Mag(32),
+        finest_mag=Mag(8),
+        sampling_mode=SamplingModes.ISOTROPIC,
+        compress=True,
+    )
```
philippotto, line 150: how about asserting that the original bbox stays intact?

normanrz:
> do you know why this is only necessary when upsampling?

This makes sure that the 2 bboxes are aligned: https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/layer.py#L857-L869
In these cases the `source_view` is actually larger than the MagView is. But that only works in `read_only` mode.

PR #920, opened 2023-06-26
Fix compressed from images oom

### LINKED ISSUE #900: compressed add_layer_from_images uses too much memory
## Context
converting image data to compressed wkw files leads possibly to an OOM issue

## Expected Behavior
add_layer_from images should chunk file before writing it to hard drive

## Current Behavior
When add_layer_from_images is called with compression enabled, it calls _handle_compressed_write on the entire file instead of a chunked file. The allocation of the np array can lead to OOM issue. 

## Steps to Reproduce the bug

1. Call add_layer_from_images on the Dataset class with big jpg image and set the argument `compress` to `True`
2. Tha call _handle_compressed_write leads to ArrayMemoryError

markbader:
### Description:
- Compressed add_layer_from_images used too much memory. This is now avoided by chunking the data before it is written.

### Issues:
- fixes #900

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -19,6 +19,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Changed
 
 ### Fixed
+- Fixed a bug where compression in add_layer_from_images uses too much memory [#900](https://github.com/scalableminds/webknossos-libs/issues/900)
 
 
 ## [0.13.0](https://github.com/scalableminds/webknossos-libs/releases/tag/v0.13.0) - 2023-06-21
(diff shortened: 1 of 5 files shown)


philippotto (commented):
PR already looks pretty good :+1: I have two small suggestions:

- rename the _handle_compressed_write* methods to _prepare_compressed_write because the methods don't do the actual writing, but instead return bbox and ndarray (this was already suboptimally named prior to this PR)
- add two comments to the two _prepare_compressed_write* methods to explain what's going on. i.e. something like this:

  a) This method takes an arbitrary sized chunk of data with an accompanying bbox, divides these into chunks which are not larger than the file shards and delegates the preparation to ...
  b) This method takes an arbitrary sized chunk of data with an accompanying bbox (ideally not larger than a shard) and enlarges that chunk to fit the shard it resides in (by reading the entire shard data and writing the passed data ndarray into the specified volume). That way, the returned data can be written as a whole shard which is a requirement for compressed writes).

philippotto (approved):
Excellent :+1:

PR #922, opened 2023-07-11
Fix parallel buffered slice writer bug

### LINKED ISSUE #919: webknossos new CLI JSON error with multiprocessing futures
## Context
- webknossos CLI version 0.13.0
- cubing tiff series of 5714 images, images require padding, total size 121G
- the same series cubes without error in an older wkcuber version (0.9.21)

## Expected Behavior
- cubing should proceed and create mag1 wkw dataset without errors

## Current Behavior

I ran from command line using this command:
```
webknossos convert --jobs 72 --voxel-size 256,256,35 --compress --name Retina_macaque_MEA-align2-regions /gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Retina_macaque_MEA/meta/webknossos/Retina_macaque_MEA-align2-regions-tiffs /gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Retina_macaque_MEA/meta/webknossos/Retina_macaque_MEA-align2-regions
```

It results in this error and terminates before any cubing occurs:
```
ERROR:root:A future crashed with an exception: Expecting value: line 1 column 1 (char 0). Future: <Future at 0x2b6ccbd4a730 state=finished raised JSONDecodeError>
╭─────────────────────────────── Traceback (most recent call last) ────────────────────────────────╮
│                                                                                                  │
│ /gpfs/soma_fs/home/watkins/mambaforge/envs/wkw-new/lib/python3.9/site-packages/webknossos/cli/co │
│ nvert.py:95 in main                                                                              │
│                                                                                                  │
│    92 │   )                                                                                      │
│    93 │                                                                                          │
│    94 │   with get_executor_for_args(args=executor_args) as executor:                            │
│ ❱  95 │   │   dataset = Dataset.from_images(                                                     │
│    96 │   │   │   source,                                                                        │
│    97 │   │   │   target,                                                                        │
│    98 │   │   │   voxel_size,                                                                    │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │              compress = True                                                                 │ │
│ │           data_format = <DataFormat.WKW: 'wkw'>                                              │ │
│ │ distribution_strategy = <DistributionStrategy.MULTIPROCESSING: 'multiprocessing'>            │ │
│ │              executor = <cluster_tools.executors.multiprocessing.MultiprocessingExecutor     │ │
│ │                         object at 0x2b6cbb79c4c0>                                            │ │
│ │         executor_args = Namespace(jobs=72,                                                   │ │
│ │                         distribution_strategy=<DistributionStrategy.MULTIPROCESSING:         │ │
│ │                         'multiprocessing'>, job_resources=None)                              │ │
│ │         job_resources = None                                                                 │ │
│ │                  jobs = 72                                                                   │ │
│ │                  name = 'Retina_macaque_MEA-align2-regions'                                  │ │
│ │                source = PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Retina_ma… │ │
│ │                target = PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Retina_ma… │ │
│ │            voxel_size = VoxelSize(x=256.0, y=256.0, z=35.0)                                  │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│ /gpfs/soma_fs/home/watkins/mambaforge/envs/wkw-new/lib/python3.9/site-packages/webknossos/datase │
│ t/dataset.py:640 in from_images                                                                  │
│                                                                                                  │
│    637 │   │   │   │   )                                                                         │
│    638 │   │   │   else:                                                                         │
│    639 │   │   │   │   category = layer_category                                                 │
│ ❱  640 │   │   │   ds.add_layer_from_images(                                                     │
│    641 │   │   │   │   filepaths[0] if len(filepaths) == 1 else filepaths,                       │
│    642 │   │   │   │   layer_name,                                                               │
│    643 │   │   │   │   category=category,                                                        │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │                 batch_size = None                                                            │ │
│ │                   category = 'color'                                                         │ │
│ │                chunk_shape = None                                                            │ │
│ │           chunks_per_shard = None                                                            │ │
│ │                        cls = <class 'webknossos.dataset.dataset.Dataset'>                    │ │
│ │                   compress = False                                                           │ │
│ │                data_format = <DataFormat.WKW: 'wkw'>                                         │ │
│ │                         ds = Dataset(PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/brigg… │ │
│ │                   executor = <cluster_tools.executors.multiprocessing.MultiprocessingExecut… │ │
│ │                              object at 0x2b6cbb79c4c0>                                       │ │
│ │                  filepaths = [                                                               │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   ... +5704                                                   │ │
│ │                              ]                                                               │ │
│ │        filepaths_per_layer = {                                                               │ │
│ │                              │   'Retina_macaque_MEA-align2-regions-tiffs': [                │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │                                                           │ │
│ │                              PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                              │   │   ... +5704                                               │ │
│ │                              │   ]                                                           │ │
│ │                              }                                                               │ │
│ │                     flip_x = False                                                           │ │
│ │                     flip_y = False                                                           │ │
│ │                     flip_z = False                                                           │ │
│ │    get_valid_pims_suffixes = <function get_valid_pims_suffixes at 0x2b6ccbbbd3a0>            │ │
│ │                 input_file = PosixPath('wafer03_manifest01329_1330_S1330R1330_stitched.tiff… │ │
│ │                input_files = [                                                               │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer03_manifest00684_0685_S685R685_stitched.tiff'), │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer02_manifest00371_0372_S372R372_stitched.tiff'), │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer01_manifest01034_1035_S1035R1035_stitched.tiff… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer01_manifest01487_1488_S1488R1488_stitched.tiff… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer03_manifest00411_0412_S412R412_stitched.tiff'), │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer01_manifest01511_1512_S1512R1512_stitched.tiff… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer02_manifest01139_1140_S1140R1140_stitched.tiff… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer01_manifest00815_0816_S816R816_stitched.tiff'), │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer03_manifest01307_1308_S1308R1308_stitched.tiff… │ │
│ │                              │                                                               │ │
│ │                              PosixPath('wafer03_manifest00894_0895_S895R895_stitched.tiff'), │ │
│ │                              │   ... +5704                                                   │ │
│ │                              ]                                                               │ │
│ │                 input_path = PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                input_upath = PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │             layer_category = None                                                            │ │
│ │                 layer_name = 'Retina_macaque_MEA-align2-regions-tiffs'                       │ │
│ │ map_filepath_to_layer_name = <function                                                       │ │
│ │                              Dataset.ConversionLayerMapping._to_callable.<locals>.<lambda>   │ │
│ │                              at 0x2b6ccbbbd670>                                              │ │
│ │                 max_layers = 20                                                              │ │
│ │                       name = 'Retina_macaque_MEA-align2-regions'                             │ │
│ │                output_path = PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Reti… │ │
│ │                    swap_xy = False                                                           │ │
│ │             use_bioformats = None                                                            │ │
│ │             valid_suffixes = {                                                               │ │
│ │                              │   'gdal',                                                     │ │
│ │                              │   'jpeg',                                                     │ │
│ │                              │   'gif',                                                      │ │
│ │                              │   'ppmraw',                                                   │ │
│ │                              │   'pgm',                                                      │ │
│ │                              │   'ics',                                                      │ │
│ │                              │   'ipl',                                                      │ │
│ │                              │   'stk',                                                      │ │
│ │                              │   'mp4',                                                      │ │
│ │                              │   'avi',                                                      │ │
│ │                              │   ... +59                                                     │ │
│ │                              }                                                               │ │
│ │                 voxel_size = VoxelSize(x=256.0, y=256.0, z=35.0)                             │ │
│ │          z_slices_sort_key = functools.partial(<function natsort_key at 0x2b6e9d47cd30>,     │ │
│ │                              key=None, string_func=<function                                 │ │
│ │                              parse_string_factory.<locals>.func at 0x2b6f26e68310>,          │ │
│ │                              bytes_func=<function parse_bytes_factory.<locals>.<lambda> at   │ │
│ │                              0x2b6f26e683a0>, num_func=<function                             │ │
│ │                              parse_number_factory.<locals>.func at 0x2b6f26e68430>)          │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /gpfs/soma_fs/home/watkins/mambaforge/envs/wkw-new/lib/python3.9/site-packages/webknossos/datase │
│ t/dataset.py:1234 in add_layer_from_images                                                       │
│                                                                                                  │
│   1231 │   │   │   │   │   module="webknossos",                                                  │
│   1232 │   │   │   │   )                                                                         │
│   1233 │   │   │   │   with get_executor_for_args(None, executor) as executor:                   │
│ ❱ 1234 │   │   │   │   │   shapes_and_max_ids = wait_and_ensure_success(                         │
│   1235 │   │   │   │   │   │   executor.map_to_futures(func_per_chunk, args),                    │
│   1236 │   │   │   │   │   │   progress_desc="Creating layer from images",                       │
│   1237 │   │   │   │   │   )                                                                     │
│                                                                                                  │
│ ╭─────────────────────────────────────────── locals ───────────────────────────────────────────╮ │
│ │                                      _ = 0                                                   │ │
│ │                       add_layer_kwargs = {}                                                  │ │
│ │                  allow_multiple_layers = True                                                │ │
│ │                                   args = [                                                   │ │
│ │                                          │   (0, 32),                                        │ │
│ │                                          │   (32, 64),                                       │ │
│ │                                          │   (64, 96),                                       │ │
│ │                                          │   (96, 128),                                      │ │
│ │                                          │   (128, 160),                                     │ │
│ │                                          │   (160, 192),                                     │ │
│ │                                          │   (192, 224),                                     │ │
│ │                                          │   (224, 256),                                     │ │
│ │                                          │   (256, 288),                                     │ │
│ │                                          │   (288, 320),                                     │ │
│ │                                          │   ... +169                                        │ │
│ │                                          ]                                                   │ │
│ │                             batch_size = 32                                                  │ │
│ │                               category = 'color'                                             │ │
│ │                                channel = None                                                │ │
│ │                            chunk_shape = None                                                │ │
│ │                             chunk_size = None                                                │ │
│ │                       chunks_per_shard = None                                                │ │
│ │                               compress = False                                               │ │
│ │                          current_dtype = dtype('uint8')                                      │ │
│ │                            czi_channel = None                                                │ │
│ │                            data_format = <DataFormat.WKW: 'wkw'>                             │ │
│ │                            dimwise_max = <function dimwise_max at 0x2b6cbb7b83a0>            │ │
│ │                                  dtype = None                                                │ │
│ │                               executor = <cluster_tools.executors.multiprocessing.Multiproc… │ │
│ │                                          object at 0x2b6cbb79c4c0>                           │ │
│ │                            first_layer = None                                                │ │
│ │                                 flip_x = False                                               │ │
│ │                                 flip_y = False                                               │ │
│ │                                 flip_z = False                                               │ │
│ │                         func_per_chunk = functools.partial(<bound method                     │ │
│ │                                          PimsImages.copy_to_view of                          │ │
│ │                                          <webknossos.dataset._utils.pims_images.PimsImages   │ │
│ │                                          object at 0x2b6cab443e50>>,                         │ │
│ │                                          mag_view=MagView(name='1',                          │ │
│ │                                          bounding_box=BoundingBox(topleft=(0, 0, 0),         │ │
│ │                                          size=(4956, 5459, 5714))), is_segmentation=False,   │ │
│ │                                          dtype=dtype('uint8'))                               │ │
│ │                                 images = [                                                   │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │                                                   │ │
│ │                                          PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/b… │ │
│ │                                          │   ... +5704                                       │ │
│ │                                          ]                                                   │ │
│ │                                  layer = Layer('Retina_macaque_MEA-align2-regions-tiffs',    │ │
│ │                                          dtype_per_channel=uint8, num_channels=1)            │ │
│ │                             layer_name = 'Retina_macaque_MEA-align2-regions-tiffs'           │ │
│ │                      layer_name_suffix = ''                                                  │ │
│ │                                    mag = Mag(_mag=Vec3Int(1,1,1))                            │ │
│ │                               mag_view = MagView(name='1',                                   │ │
│ │                                          bounding_box=BoundingBox(topleft=(0, 0, 0),         │ │
│ │                                          size=(4956, 5459, 5714)))                           │ │
│ │                             max_layers = 20                                                  │ │
│ │                            pims_images = <webknossos.dataset._utils.pims_images.PimsImages   │ │
│ │                                          object at 0x2b6cab443e50>                           │ │
│ │                       pims_open_kwargs = {}                                                  │ │
│ │                             PimsImages = <class                                              │ │
│ │                                          'webknossos.dataset._utils.pims_images.PimsImages'> │ │
│ │                        possible_layers = None                                                │ │
│ │                                   self = Dataset(PosixPath('/gpfs/soma_fs/cne-mSEM/mSEM-pro… │ │
│ │ suffix_with_pims_open_kwargs_per_layer = {'': {}}                                            │ │
│ │                                swap_xy = False                                               │ │
│ │                              timepoint = None                                                │ │
│ │                                topleft = Vec3Int(0,0,0)                                      │ │
│ │                   truncate_rgba_to_rgb = True                                                │ │
│ │                         use_bioformats = None                                                │ │
│ │                                  z_end = 5714                                                │ │
│ │                                z_start = 5696                                                │ │
│ ╰──────────────────────────────────────────────────────────────────────────────────────────────╯ │
│                                                                                                  │
│ /gpfs/soma_fs/home/watkins/mambaforge/envs/wkw-new/lib/python3.9/site-packages/webknossos/utils. │
│ py:137 in wait_and_ensure_success                                                                │
│                                                                                                  │
│   134 │   │   with get_rich_progress() as progress:                                              │
│   135 │   │   │   task = progress.add_task(progress_desc, total=len(futures))                    │
│   136 │   │   │   for fut in as_completed(futures):                                              │
│ ❱ 137 │   │   │   │   results.append(fut.result())                                               │
│   138 │   │   │   │   progress.update(task, advance=1)                                           │
│   139 │   return results                                                                         │
│   140                                                                                            │
│                                                                                                  │
│ ╭───────────────────────────────────── locals ─────────────────────────────────────╮             │
│ │           fut = <Future at 0x2b6ccbd01730 state=finished raised JSONDecodeError> │             │
│ │       futures = [                                                                │             │
│ │                 │   <Future at 0x2b6cd3e123d0 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6ccbbda1f0 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3e12460 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3e12910 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3e12be0 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3e129a0 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3eb8e50 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3eb8370 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3eb8f70 state=finished returned tuple>,    │             │
│ │                 │   <Future at 0x2b6cd3eb8c40 state=finished returned tuple>,    │             │
│ │                 │   ... +169                                                     │             │
│ │                 ]                                                                │             │
│ │      progress = <rich.progress.Progress object at 0x2b6cd3d5d850>                │             │
│ │ progress_desc = 'Creating layer from images'                                     │             │
│ │       results = []                                                               │             │
│ │          task = 0                                                                │             │
│ ╰──────────────────────────────────────────────────────────────────────────────────╯             │
│                                                                                                  │
│ /gpfs/soma_fs/home/watkins/mambaforge/envs/wkw-new/lib/python3.9/concurrent/futures/_base.py:439 │
│ in result                                                                                        │
│                                                                                                  │
│   436 │   │   │   │   if self._state in [CANCELLED, CANCELLED_AND_NOTIFIED]:                     │
│   437 │   │   │   │   │   raise CancelledError()                                                 │
│   438 │   │   │   │   elif self._state == FINISHED:                                              │
│ ❱ 439 │   │   │   │   │   return self.__get_result()                                             │
│   440 │   │   │   │                                                                              │
│   441 │   │   │   │   self._condition.wait(timeout)                                              │
│   442                                                                                            │
│                                                                                                  │
│ ╭──── locals ────╮                                                                               │
│ │    self = None │                                                                               │
│ │ timeout = None │                                                                               │
│ ╰────────────────╯                                                                               │
│                                                                                                  │
│ /gpfs/soma_fs/home/watkins/mambaforge/envs/wkw-new/lib/python3.9/concurrent/futures/_base.py:391 │
│ in __get_result                                                                                  │
│                                                                                                  │
│   388 │   def __get_result(self):                                                                │
│   389 │   │   if self._exception:                                                                │
│   390 │   │   │   try:                                                                           │
│ ❱ 391 │   │   │   │   raise self._exception                                                      │
│   392 │   │   │   finally:                                                                       │
│   393 │   │   │   │   # Break a reference cycle with the exception in self._exception            │
│   394 │   │   │   │   self = None                                                                │
│                                                                                                  │
│ ╭── locals ───╮                                                                                  │
│ │ self = None │                                                                                  │
│ ╰─────────────╯                                                                                  │
╰──────────────────────────────────────────────────────────────────────────────────────────────────╯
JSONDecodeError: Expecting value: line 1 column 1 (char 0)
```

I ran the cubing to completion using an older version of the wkcuber CLI (0.9.21), which I modified so that it supports the  `--pad` argument with this command line:
```
python -m wkcuber --jobs 72 --scale 256,256,35 --max_mag 32 --sampling_mode constant_z --pad --name Retina_macaque_MEA-align2-regions /gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Retina_macaque_MEA/meta/webknossos/Retina_macaque_MEA-align2-regions-tiffs /gpfs/soma_fs/cne-mSEM/mSEM-proc/2021/briggman/Retina_macaque_MEA/meta/webknossos/Retina_macaque_MEA-align2-regions
```

## Steps to Reproduce the bug

It's not clear to me the source of the issue, but unless it's something specific about my environment or some specific combination of options, it seems to me that just running `webknossos convert` using multiprocessing on a tiff series should reproduce the error.

## Your Environment for bug

```
(wkw-new) [watkins@somalogin01 cubing]$ python --version
Python 3.9.16
(wkw-new) [watkins@somalogin01 cubing]$ pip list installed | grep webknossos
webknossos                0.13.0
(wkw-new) [watkins@somalogin01 cubing]$ uname -a
Linux somalogin01 3.10.0-1160.el7.x86_64 #1 SMP Mon Oct 19 16:18:59 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux
```

webknossos CLI was installed with:
```
pip install "webknossos[all]"
```

markbader:
### Description:
- Hot fix implementation of bbox extension to avoid parallel access to properties JSON during Dataset from_images

### Issues:
- fixes #919 

### Todos:
 - [ ] Updated Changelog


markbader:
I added a boolean argument `update_bbox` to the named partial `copy_to_view` and propagated it to the actual write of the view or the mag view. This makes sure that the bounding box is not updated by multiple processes. Therefore, the corruption of the properties JSON file should be avoided.

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -34,6 +34,7 @@ def __init__(
         self,
         view: "View",
         offset: Optional[Vec3IntLike] = None,
+        update_bbox: bool = False,
```
fm3, line 37: I’d say the default should be `True`, and there could be a comment explaining what it does and to encourage setting it to false IF parallel access is intended.

webknossos/webknossos/dataset/view.py
```diff
@@ -192,6 +192,7 @@ def write(
         self,
         data: np.ndarray,
         offset: Optional[Vec3IntLike] = None,  # deprecated, relative, in current mag
+        update_bbox: bool = True,
```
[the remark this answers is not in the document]
fm3: Also, did you test what happens if there is a write outside of the existing box? does it just work?

fm3:
@philippotto if you have time, could you also have a look at these changes? I’m not super sure about the semantics with view vs magview and I think another opinion would make sense

webknossos/webknossos/dataset/view.py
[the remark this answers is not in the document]
markbader: 1. I agree, that the name `allow_write_outside_view_bbox` fits better for `View.write()` but it seems a bit odd to have different signatures for View and MagView.

2. Yes, it just worked for the files that I used to test it.

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -124,12 +129,22 @@ def _write_buffer(self) -> None:
...
+                    update_bbox=self.update_bbox,
+                )
+            else:
+                self.view.write(
+                    data,
+                    offset=buffer_start.add_or_none(self.offset),
+                    relative_offset=buffer_start_mag1.add_or_none(self.relative_offset),
+                    absolute_offset=buffer_start_mag1.add_or_none(self.absolute_offset),
+                    allow_write_outside_bbox=not self.update_bbox,
+                )
```
philippotto, line 147: This can be DRYed with something like this, I think:

```
kwargs = None
if isinstance(self.view, MagView):
    kwargs = {"update_bbox": self.update_bbox}
else:
    kwargs = {"allow_write_outside_bbox": not self.update_bbox}

self.view.write(
    data,
    offset=buffer_start.add_or_none(self.offset),
    relative_offset=buffer_start_mag1.add_or_none(self.relative_offset),
    absolute_offset=buffer_start_mag1.add_or_none(self.absolute_offset),
    **kwargs
)
```

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -124,12 +129,22 @@ def _write_buffer(self) -> None:
...
+                    absolute_offset=buffer_start_mag1.add_or_none(self.absolute_offset),
+                    update_bbox=self.update_bbox,
+                )
+            else:
+                self.view.write(
+                    data,
+                    offset=buffer_start.add_or_none(self.offset),
+                    relative_offset=buffer_start_mag1.add_or_none(self.relative_offset),
+                    absolute_offset=buffer_start_mag1.add_or_none(self.absolute_offset),
+                    allow_write_outside_bbox=not self.update_bbox,
```
philippotto, line 146: I don't understand this logic. If the bbox must not be updated, writing outside of the bbox is allowed? Shouldn't this be inverted?

philippotto (commented):
As a bandaid hotfix this seems okay to me (however, shouldn't the default for `update_bbox` be false everywhere to avoid the runtime errors?).

Is there a plan for a proper fix? The PR would cause bounding boxes to be not set correctly potentially, if I understand it correctly?

fm3:
> Is there a plan for a proper fix? The PR would cause bounding boxes to be not set correctly potentially, if I understand it correctly?

I’m unsure how to go about this. The idea here was to move the responsibility of setting the full bbox from the subjobs (which do the write call) to the main process (afterwards). The benefit would be that the subjobs don’t run into this parallel editing of the json problem, and yet, the final bbox would be correct (as the caller, in this case add_layer_from_images, can then set it).

The idea was to have the defaults be True, which is the intuitive behavior (everything updates), and only set them to False explicitly in the cases where concurrent usages are intended (creating a need to set the bbox in the caller code afterwards).

A problem that arose from this is that the write calls must now be able to write outside of the (previous, smaller) bbox, because it will be set only at the end. That is why there is the `allow_write_outside_bbox` param in this case.

I agree that this is all a bit hard to comprehend from the code alone. I am not super happy with this solution, but I also don’t have an idea for a “proper fix” – if you have ideas, please share them!

Another idea was to set the bbox to a huge one (millions of voxel per dimensions) before calling the subjobs. In that case, the subjobs would not touch the json as nothing is bigger than expected. That would alleviate the need for passing these booleans. However, itwould introduce magic numbers for this huge bbox, and also not feel like a proper fix.

We could probably also implement a locking mechanism for the json, but I’m not sure how to do that (in a way that does not impede performance)

philippotto:
I see two possible solutions for this problem (I'm preferring (2)):

1) Defer the mutation of the JSON. This is similar to what the PR is currently doing, but I'd like to see an automatic fix-up of the JSON, so that the users of this library don't need to take care of this. Also, they shouldn't need to pass special parameters if they use multiprocessing as this is complicated. The automated fix-up could be done by remembering all the written bboxes and passing them to the parent job in the end.

2) Use a locking mechanism as you also suggested. We have something similar in [voxelytics](https://github.com/scalableminds/voxelytics/blob/6d3ef9ac5fc77e3710c00751d6a2b4616373f13d/voxelytics/commons/workflow/workflow.py#L372-L372) and it looks relatively simple (however, I did not author this, maybe @normanrz wants to chime in?). It's file system based which should be okay since the JSON itself is FS-based, too. The lock file could be named `datasource-properties.json.lock`. Since the update of the JSON should take very little time (in comparison to writing image data), I don't think that performance will be a problem.
We should take extra care to reduce the chance that the lock file is not cleaned up. That is, the locking phase should be as short as possible (i.e., only when updating the JSON and only when the JSON really changes). If the JSON update raises an exception for some reason, the lock file should also be cleaned up again. Then, the lock file might only be a problem when the python process is terminated during the update of the JSON. In that case, a following run would probably dead-lock. A warning could be emitted if the lock is awaited for more than X seconds with the hint how it can be deleted manually.

I think, I'm in strong favor of (2) since it seems way simpler to me.

fm3:
I would have assumed that file locks are not guaranteed to be multiprocessing-safe, but I may be wrong about that. Do we use it in voxelytics in the same way without issue? Then I’m ok with that approach. Maybe you could give @markbader an introduction on how it is done in voxelytics so that he can adapt it into here?

philippotto:
> I would have assumed that file locks are not guaranteed to be multiprocessing-safe, but I may be wrong about that.

I think it should be safe. The [package](https://py-filelock.readthedocs.io/en/latest/index.html#) describes the locking mechanism as a "a simple way of inter-process communication".

> Do we use it in voxelytics in the same way without issue? Then I’m ok with that approach.

In vx we use the locking to avoid that the same workflow task is run multiple times (e.g., due to user error). So, I think, that this should work for the libs use case, too.

> Maybe you could give @markbader an introduction on how it is done in voxelytics so that he can adapt it into here?

@markbader See [this code](https://github.com/scalableminds/voxelytics/blob/6d3ef9ac5fc77e3710c00751d6a2b4616373f13d/voxelytics/commons/workflow/workflow.py#L372-L372). I think, it's as simple as importing the module and adding the `with SoftFileLock(lock_path, timeout=3):` around the update of the JSON file. Let me know if you have questions :)

markbader:
@philippotto Thanks for your feedback and implementation ideas. I tried to implement the `SoftFileLock` approach but ran into some issues:

1. The `filelock` python library does not support `UPaths` right now, and therefore is not usable for remote datasets.
2. I adapted the `filelock` library to support UPaths and replaced the `os.open` calls with `pathlib.Path(..).open('x')` as it is the python representation for atomic write a file and throw an Exception if it already exists. Sadly, the `s3fs` library does not support the mode "x" for open. I think it is out of scope to change the `s3fs` library as well to implement a file lock. 

I just discussed with @daniel-wer that under the given circumstances, the implementation of your first proposed solution would be better. @philippotto do you have another idea for an implementation of a locking mechanism, or would you agree that I should implement solution 1?

philippotto:
> I just discussed with @daniel-wer that under the given circumstances, the implementation of your first proposed solution would be better. @philippotto do you have another idea for an implementation of a locking mechanism, or would you agree that I should implement solution 1?

I would hope that a filelock-implementation would be possible with S3, but I also don't see an easy way to achieve it without changing the entire approach that is used by the file-lock library. Alternatively, one could use a file lock somewhere else (e.g. in the current working directory) that is shared between all workers. For multiprocessing this would be easy (the only downside is that this wouldn't catch independent actors writing to the same DS). However, I don't see how this would be generalized to slurm etc.

Therefore, I'm fine with doing (1) :)

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1248,9 +1248,11 @@ def add_layer_from_images(
                     .offset(topleft)
                 )
             if pims_images.expected_shape != actual_size:
-                warnings.warn(
+                logger.info(
```
philippotto, line 1251: why did you switch from warnings.warn to logger.info here?

webknossos/webknossos/dataset/_utils/pims_images.py
```diff
@@ -496,6 +498,7 @@ def copy_to_view(
             with mag_view.get_buffered_slice_writer(
                 relative_offset=(0, 0, z_start * mag_view.mag.z),
                 buffer_size=mag_view.info.chunk_shape.z,
+                json_update_allowed=False,
```
philippotto, line 501: please add a comment that `copy_to_view` is typically used in a multiprocessing-context and therefore needs this flag.

philippotto (commented):
Great, looks good to me :) Two things:

- The documentation should state somewhere explicitly which restrictions exist for the parallel access of wk datasets. Namely:
  - when writing shards in parallel, json_update_allowed should be set to False to disable the automatic update of the bounding box meta data. otherwise, race conditions may happen. the user is responsible for updating the bounding box manually.
  - when writing to chunks in shards, one chunk may only be written to by one actor at any time
  - when writing to compressed shards, one shard may only be written to by one actor at any time
  - reading in parallel should be fine.
- see my one review comment

@normanrz could you double check my restrictions from above?

normanrz:
>   * when writing to chunks in shards, one chunk may only be written to by one actor at any time
>   * when writing to compressed shards, one shard may only be written to by one actor at any time

Ok for now. With Zarr, parallel writes to a shard will no longer be allowed at all. Maybe we already put such wording in. 

>   * reading in parallel should be fine.

reading in parallel without concurrent writes is ok.

webknossos/webknossos/dataset/dataset.py
markbader: I think I experimented a bit and forget to change it back. It is now `warning.warn` again.

webknossos/webknossos/dataset/_utils/pims_images.py
markbader: Is done with commit ea7dc34.

markbader:
@philippotto Thanks for your review! I added a paragraph about parallel access of Datasets in the Dataset Usage part of the docs.

philippotto (approved):
Excellent, looks good to me :)

fm3:
> With Zarr, parallel writes to a shard will no longer be allowed at all.

That sounds like trouble for some of our applications that do slice-wise operations. What would be the right way to deal with that? Shards with very small z size so that we can do the parallelization? Or ditch the fine-grained parallelization to write to the large chunk files sequentially each? Or re-chunking in a second pass? All seem to have drawbacks compared to now.

philippotto:
> > With Zarr, parallel writes to a shard will no longer be allowed at all.
> 
> That sounds like trouble for some of our applications that do slice-wise operations. What would be the right way to deal with that? Shards with very small z size so that we can do the parallelization? Or ditch the fine-grained parallelization to write to the large chunk files sequentially each? Or re-chunking in a second pass? All seem to have drawbacks compared to now.

For the vx align case, picking a small shard-z dimension should be a good option imo. Especially, because zarr supports non-cubic shards (so, x and y can still have larger sizes). After materialization, a recubing would be necessary (not strictly, but good for WK) which is a drawback, but on the other hand a low shard-z-size would allow higher parallelization.

PR #924, opened 2023-07-13
No slashes in layer name

### LINKED ISSUE #918: Dataset.from_images should not add slashes in layer names
## Context

Some datasets are imported with subfolders. The layer names should not contain slashes. It should be enforced that layers can only have valid names. Ask @fm3 for the regex.

markbader:
### Description:
- Slashes in layer names resulted in bugs, because of the unconventional file naming. The conversion layer mapping was adapted to ensure, that layer names do not contain slashes.

### Issues:
- fixes #918 

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -17,6 +17,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Added
 
 ### Changed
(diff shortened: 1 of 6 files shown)


normanrz (commented):
Code looks good. Would it be possible to add a test?

markbader:
I added a test that checks for all ConversionLayerMappings if the layer name includes a slash.

normanrz (approved):
Thanks!

PR #925, opened 2023-07-13
make largest_segment_id optional

### LINKED ISSUE #917: Make largest_segment_id optional
## Detailed Description

When creating a new segmentation layer, the largest_segment_id should not be required. Webknossos doesn't require it anymore.
We could add a function to determine and set the max_id using an executor.

markbader:
### Description:
- Largest segment id was required in webknossos libs, as it raises an Assertion Error if absence. As it is no longer required in webknossos it is optional in webknossos libs now.

### Issues:
- fixes #917 

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/layer.py
+++ b/webknossos/webknossos/dataset/layer.py
@@ -1162,11 +1162,11 @@ class SegmentationLayer(Layer):
     def _get_largest_segment_id(self, view: View) -> int:
         return np.max(view.read(), initial=0)
 
-    def update_largest_segment_id(
+    def refresh_largest_segment_id(
         self, chunk_shape: Optional[Vec3Int] = None, executor: Optional[Executor] = None
     ) -> None:
         """Sets the largest segment id to the highest value in the data.
-        largest_segment_id is set to `None` if data is undefined."""
+        largest_segment_id is set to `None` if the data is empty."""
 
         try:
             chunk_results = self.get_finest_mag().map_chunk(
--- a/webknossos/webknossos/dataset/layer.py
+++ b/webknossos/webknossos/dataset/layer.py
@@ -1134,8 +1134,11 @@ class SegmentationLayer(Layer):
     _properties: SegmentationLayerProperties
 
     @property
-    def largest_segment_id(self) -> Optional[int]:
-        return self._properties.largest_segment_id
+    def largest_segment_id(self) -> int:
+        if max_id := self._properties.largest_segment_id:
+            return max_id
+
+        return self._calc_largest_segment_id()
 
     @largest_segment_id.setter
     def largest_segment_id(self, largest_segment_id: Optional[int]) -> None:
@@ -1155,3 +1158,8 @@ class SegmentationLayer(Layer):
 
     def _get_largest_segment_id_maybe(self) -> Optional[int]:
         return self.largest_segment_id
+
+    def _calc_largest_segment_id(self) -> int:
+        if data := self.get_finest_mag().read().flatten():
+            return data.max()
+        return 0
--- a/webknossos/webknossos/dataset/layer.py
+++ b/webknossos/webknossos/dataset/layer.py
@@ -14,7 +14,7 @@ from numpy.typing import DTypeLike
 from upath import UPath
 
 from webknossos.dataset.sampling_modes import SamplingModes
-from webknossos.dataset.view import _copy_job
+from webknossos.dataset.view import View, _copy_job
 from webknossos.geometry import BoundingBox, Mag, Vec3Int, Vec3IntLike
 
 from ._array import ArrayException, BaseArray, DataFormat
@@ -46,6 +46,7 @@ from ..utils import (
     is_fs_path,
     named_partial,
     rmtree,
+    wait_and_ensure_success,
     warn_deprecated,
 )
 from .defaults import (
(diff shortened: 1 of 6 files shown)


markbader:
Most of this issue was already solved in #786. This PR removes an assertion that prevents the creation of a segmentation layer without the largest segment id and adapts the test to check the expected behavior. I assumed that a segmentation layer that contains no data has `None` as `largest_segment_id`.

fm3 (commented):
@normanrz do you have an opinion concerning None for newly created empty segmentation layers? I guess zero could also be a valid value in that case, but if the data write functions don’t update it, it may make more sense to leave it at None so that it is more explicitly unset?

Other than this question, the changes look good to me.

normanrz:
> @normanrz do you have an opinion concerning None for newly created empty segmentation layers? I guess zero could also be a valid value in that case, but if the data write functions don’t update it, it may make more sense to leave it at None so that it is more explicitly unset?

I'd leave it at `None`

normanrz:
Please also add a method to determine+save the max_id.

fm3:
> Please also add a method to determine+save the max_id.

You can probably take that (or inspiration) from https://github.com/scalableminds/voxelytics/blob/master/voxelytics/worker/jobs/find_largest_segment_id.py

markbader:
I added a function `update_largest_segment_id` that determines and updates the largest segment id. I thought, that parallelization might be useful for bigger datasets. 
Otherwise it might be done with a one-liner like:
```python
def _find_largest_segment_id(self):
    self.get_finest_mag().read().flatten().max()
```
Do you have thought on the current implementation? @normanrz @fm3

fm3:
I agree that parallelization is good to have for that :) Maybe you could make use of `MagView.for_each_chunk` to shorten the code?

I think it would be good to avoid `flatten`, as it creates a copy of the array. I used `np.max(data, initial=0)` in my implementation in voxelytics, which should work for nd-arrays over all axes. Also, the initial=0 prevents exceptions for empty arrays.

normanrz:
> I agree that parallelization is good to have for that :) Maybe you could make use of `MagView.for_each_chunk` to shorten the code?

Definitely needs parallelization. These segmentations can be 100s of TVx.

Perhaps, we should add a `View.map_chunk` function because `for_each_chunk` doesn't collect return values of the functions.

markbader:
I refactored the `update_largest_segment_id` method as @fm3 suggested and implemented a `View.map_chunk` method. Additionally, I added a test for the `update_largest_segment_id` method. 

Is it reasonable to assume that `update_largest_segment_id` is called by a user of the library if the segmentation data is changed manually, or should the segment id be updated automatically when data of a segmentation layer is edited?

webknossos/webknossos/dataset/layer.py
```diff
@@ -1158,3 +1158,23 @@ def category(self) -> LayerCategoryType:
 
     def _get_largest_segment_id_maybe(self) -> Optional[int]:
         return self.largest_segment_id
+
+    def _get_largest_segment_id(self, view: View) -> int:
+        return np.max(view.read(), initial=0)
+
+    def update_largest_segment_id(
```
fm3, line 1165:
```suggestion
    def refresh_largest_segment_id(
```

With “update” I would expect a simple setter, maybe refresh is the better name, signaling the semantics and that this is actually an expensive operation?

webknossos/webknossos/dataset/layer.py
```diff
@@ -1158,3 +1158,23 @@ def category(self) -> LayerCategoryType:
...
+        self, chunk_shape: Optional[Vec3Int] = None, executor: Optional[Executor] = None
+    ) -> None:
+        """Sets the largest segment id to the highest value in the data.
+        largest_segment_id is set to `None` if data is undefined."""
```
fm3, line 1169:
```suggestion
        largest_segment_id is set to `None` if the data is empty."""
```

fm3 (commented):
Cool stuff, looks good to me!

I’m not certain about auto-updating the largest_segment_id. It would be a nice feature, but it may be expensive, since we have to call max on all written data. Also, when opening an existing dataset, we can never be certain that the largest_segment_id in its metadata is actually correct, so we might not want to give the impression that we can guarantee it being correct and up to date.

Do you have opinions about this @philippotto @normanrz ?

philippotto:
> I’m not certain about auto-updating the largest_segment_id. It would be a nice feature, but it may be expensive, since we have to call max on all written data. [...]

I'm leaning towards auto-updating the LSI, since it simplifies data handling for the user. However, it shouldn't add an additional read-pass across all written data chunks (instead, during each write, it should compute the maximum of the input data which is already in-memory). 

> Also, when opening an existing dataset, we can never be certain that the largest_segment_id in its metadata is actually correct, so we might not want to give the impression that we can guarantee it being correct and up to date.

If a dataset already has an LSI, that value should be trusted in my opinion (if existing metadata cant be trusted the user should be responsible for fixing it). If the LSI is None (and bbox > 0), an automatic update should not happen, as partial writes wouldn't be able to know the new LSI properly.

Does this make sense?

fm3:
Sounds fair, let’s wait if norman has an opinion as well :)

Also note that this may also need a file lock for the json during in parallel writes.

normanrz:
I don't think an auto-update is necessary. It would bring the same concurrency issues as the bbox changes.

In the future, Webknossos should have an "optimize" worker job for segmentations that
- rechunks the data, if necessary
- calculates largest_segment_id
- calculates a segment index file

That job could run directly on import or user-triggered.

markbader:
Do you think an entry in the documentation or an example would be useful?

Otherwise, since there is no longer a need for an auto-update mechanism, I think the PR is in a state that it can be merged.

normanrz (approved):
Yes, I think some documentation on the `largest_segment_id` would be nice in the `write` method.

webknossos/webknossos/dataset/view.py
```diff
@@ -203,6 +204,25 @@ def write(
         This parameter used to be relative for `View` and absolute for `MagView`,
         and specified in the mag of the respective view.
 
+        Writing data to a segmentation layer manually does not automatically update the largest_segment_id. To set
+        the largest segment id properly run the `refresh_largest_segment_id` method on your layer.
```
normanrz, line 208:
```suggestion
        the largest segment id properly run the `refresh_largest_segment_id` method on your layer or set the `largest_segment_id` manually.
```

PR #929, opened 2023-07-20
Add detection for exceeded time limits for slurm jobs

### LINKED ISSUE #921: Cluster tools should recognize that cluster time limits were reached
## Detailed Description
- Affected library: cluster_tools

The slurm executor automatically recognizes if a job was killed due to excessive RAM usage and advises users on how to adapt the config to fix that. It would be great if the executor would also recognize that a job was killed due to the configured time limit and similarly would advise users on how to adapt the config. It can also happen that both of these conditions apply in which case users should be advised to fix both.

Since this is currently done via exceptions, I'm not sure about the best way to accomplish that. It might make sense to generalize the `RemoteOutOfMemoryException` to a `RemoteOutOfResourcesException` which can notify about memory and time limits at once, but maybe there are other options too.

## Use Cases & Context

Enables users to self-help in case of configuration issues and eases support burden on our side.

daniel-wer:
### Description:
- Also, add tests to check that job kills due to memory or time are detected by the cluster_tools.
Use a new version of the slurm docker image to be able to use the `seff` tool which is usually included in slurm installations, but wasn't included in our docker image. `seff` is used to detect OOM kills.

### TODO:
- [x] Skip time limit test, because it takes >1m due to Slurm's minute-resolution
- [x] Change the tag in the docker-compose to use the newest slurm-docker-cluster version

### Issues:
- fixes #921 

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests


cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -411,69 +411,110 @@ def matches_states(slurm_states: List[str]) -> bool:
...
+            return None
+
+        time_limit = properties.get("TimeLimit", None)
+        run_time = properties.get("RunTime", None)
+        time_limit_note = f"Time Limit: {time_limit} Run Time: {run_time}"
+
+        reason = f"The job was probably terminated because it ran for too long ({time_limit_note})."
+        return (reason, RemoteResourceLimitException)
+
+    def _investigate_memory_consumption(
```
daniel-wer, line 475: This method was only moved, not modified

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -411,69 +411,110 @@ def matches_states(slurm_states: List[str]) -> bool:
...
+        except ValueError:
+            return None
+
+        if percentage < 100:
+            return None
+
+        reason = f"The job was probably terminated because it consumed too much memory ({efficiency_note})."
+        return (reason, RemoteResourceLimitException)
+
+    def _investigate_exit_code(
```
daniel-wer, line 502: This method was only moved, not modified

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -411,69 +413,110 @@ def matches_states(slurm_states: List[str]) -> bool:
...
+        # the job run time exceeded the specified time limit. If that is not the case, it
+        # checks whether the job used too much RAM. As a last resort, it checks the exit code
+        # of the job: If the job was killed with signal 9, it's very likely due to some
+        # resource limit.
+
+        def parse_key_value_pairs(
+            pair_delimiter: str, key_value_delimiter: str
+        ) -> Dict[str, str]:
+            properties = {}
+            for key_value_pair in stdout.split(pair_delimiter):
```
philippotto, line 426: I stumbled a bit here, because this function reads `stdout` which is defined below said function. how about passing it as an input parameter to the function?

cluster_tools/tests/test_slurm.py
```diff
@@ -284,11 +291,70 @@ def test_slurm_max_array_size() -> None:
...
     finally:
-        chcall(f"sed -i 's/{command}//g' /etc/slurm/slurm.conf && scontrol reconfigure")
+        chcall(
+            f"sed -ci 's/{command}//g' /etc/slurm/slurm.conf && scontrol reconfigure"
+        )
         reset_max_array_size = executor.get_max_array_size()
         assert reset_max_array_size == original_max_array_size
 
 
+# TODO: Comment back in after the test ran through in the CI
```
philippotto, line 301: todo

cluster_tools/tests/test_slurm.py
```diff
@@ -284,11 +291,70 @@ def test_slurm_max_array_size() -> None:
...
+    )
+
+    original_gather_frequency_config = "JobAcctGatherFrequency=30"  # from slurm.conf
+    new_gather_frequency_config = "JobAcctGatherFrequency=1"
+
+    try:
+        # Increase the frequency at which slurm checks whether a job uses too much memory
+        chcall(
+            f"sed -ci 's/{original_gather_frequency_config}/{new_gather_frequency_config}/g' /etc/slurm/slurm.conf && scontrol reconfigure"
+        )
```
philippotto, line 336: how about extracting this into a helper function, since it's also used in line 294?

philippotto: if you want to get extra fancy, the helper function could return an `undo` function. something like this:

```python
def change_config(old, new):
  chcall(f"sed {old} ... {new}")
  def undo():
    chcall(f"sed {new} ... {old}")
  return undo
```

philippotto (approved):
splendid stuff :+1: Only left a couple of nitpicks :)

cluster_tools/cluster_tools/schedulers/slurm.py
daniel-wer: Oh, yes that was by accident! I'll pass it in as a parameter.

daniel-wer:
@philippotto I found out and fixed the issue with the slurm setup in the CI. The slurm tests ran through successfully, already once, so I'd be hopeful that the CI will succeed now.

Would be great if you could have another look at the latest changes. Regarding the name-clashes with standard lib modules, it's a pity that pylint confuses those (this used to be a general issue in python 2.x, but should no longer be in python 3.x :/). Still, the general consensus online was to avoid naming modules the same as standard lib modules. I mitigated this by appending an underscore, but let me know if you've got a better idea :)

cluster_tools/cluster_tools/_utils/call.py
```diff
@@ -13,6 +13,8 @@ def call(command: str, stdin: Optional[str] = None) -> Tuple[str, str, int]:
         stdin_flag = None
     p = subprocess.run(
         command,
+        stdin=stdin_flag,
```
philippotto, line 16: wow, good catch :+1:

cluster_tools/dockered-slurm/cgroup.conf
```diff
@@ -0,0 +1,3 @@
+# autodetect which is the default detects cgroup/v2 in the github CI,
+# which fails during the initialization of the c1 and c2 nodes
+CgroupPlugin=cgroup/v1
```
philippotto, line 3: kudos! how did you find this? was there a hint about this in the logs?

philippotto (approved):
Awesome! Chapeau for fighting through this and also for the major clean ups regarding linting :muscle: :superhero: 

> @philippotto I found out and fixed the issue with the slurm setup in the CI. The slurm tests ran through successfully, already once, so I'd be hopeful that the CI will succeed now.

Great! Can you link the successful CI run? I couldn't find it auf anhieb.

> Would be great if you could have another look at the latest changes. Regarding the name-clashes with standard lib modules, it's a pity that pylint confuses those (this used to be a general issue in python 2.x, but should no longer be in python 3.x :/). Still, the general consensus online was to avoid naming modules the same as standard lib modules. I mitigated this by appending an underscore, but let me know if you've got a better idea :)

Since I don't have a better idea, I'm fine with the underscore :)

cluster_tools/dockered-slurm/cgroup.conf
daniel-wer: The initialization of the c1 and c2 nodes failed but there wasn't any hint on what went wrong. Therefore, I added

```
# Show log output for debugging
docker logs slurmctld
docker logs c1
docker logs c2
```

to the ci.yml which outputs the initialization log. In there I saw

```
slurmd: error: Couldn't find the specified plugin name for cgroup/v2 looking at all files
slurmd: error: cannot find cgroup plugin for cgroup/v2
slurmd: error: cannot create cgroup context for cgroup/v2
slurmd: error: Unable to initialize cgroup plugin
slurmd: error: slurmd initialization failed
```

and a quick local cross-check showed that at least on my laptop cgroup/v1 is used instead, successfully :)

I left the `docker logs ...` statements in there as they don't do any harm and might help again in the future.

philippotto: great find :) and :+1: for the added logs.

PR #930, opened 2023-07-24
Rename Open Tasks to Pending Tasks

fm3:
### Description:
- Open Tasks are now named Pending Tasks
- Refreshes the client with new version of openapi-python-client lib
- Note that that new version required a newer version of httpx, which has a breaking change (follow_redirects was renamed and now defaults to false)

### Issues:
- contributes to https://github.com/scalableminds/webknossos/issues/7218

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


philippotto (approved):
lgtm :+1:

webknossos/webknossos/administration/task.py
```diff
@@ -28,7 +28,7 @@
 
 @attr.frozen
 class TaskStatus:
-    open_instance_count: int
+    pending_instance_count: int
     active_instance_count: int
     finished_instance_count: int
```
normanrz, line 33:
```suggestion
    pending_instance_count: int
    active_instance_count: int
    finished_instance_count: int
    
    @property
    def open_instance_count(self) -> int:
        warn_deprecated("open_instance_count", "pending_instance_count")
        return self.pending_instance_count
```

PR #932, opened 2023-07-31
update zarrita, relaxes numpy requirement

normanrz:
- Updates zarrita, relaxes numpy requirement


webknossos/webknossos/dataset/_array.py
```diff
@@ -504,26 +505,30 @@ def info(self) -> ArrayInfo:
...
         zarray = self._zarray
         if isinstance(zarray, Array):
-            if len(zarray.codecs) == 1 and isinstance(zarray.codecs[0], ShardingCodec):
-                sharding_codec = zarray.codecs[0]
+            if len(zarray.codec_pipeline.codecs) == 1 and isinstance(
+                zarray.codec_pipeline.codecs[0], ShardingCodec
+            ):
+                sharding_codec = zarray.codec_pipeline.codecs[0]
+                shard_shape = zarray.metadata.chunk_grid.configuration.chunk_shape
+                chunk_shape = zarray.metadata.chunk_grid.configuration.chunk_shape
```
markbader, line 513: `shard_shape` and `chunk_shape` are both initialized with the same value.

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1173,6 +1174,10 @@ def add_layer_from_images(
                     chunk_shape = DEFAULT_CHUNK_SHAPE.with_z(1)
                 if chunks_per_shard is None:
                     chunks_per_shard = DEFAULT_CHUNKS_PER_SHARD_ZARR.with_z(1)
+
+            if chunks_per_shard is None and layer.data_format == DataFormat.Zarr3:
+                chunks_per_shard = DEFAULT_CHUNKS_PER_SHARD_FROM_IMAGES
+
```
markbader, line 1180: In my opinion, this should happen in `Layer.add_mag` method because the default values for `chunks_per_shard` for wkw and Zarr are set there.

webknossos/tests/dataset/test_dataset.py
```diff
@@ -867,6 +867,7 @@ def test_chunking_wk(data_format: DataFormat, output_path: Path) -> None:
     ds_path = prepare_dataset_path(data_format, output_path)
     ds = Dataset(ds_path, voxel_size=(2, 2, 1))
     chunk_shape, chunks_per_shard = default_chunk_config(data_format, 8)
+    shard_shape = chunk_shape * chunks_per_shard
```
markbader, line 870: I think the naming of `shard_shape` and `chunk_shape` is confusing. Why is the `chunk_shape` for initializing the mag different from the `chunk_shape` for iterating over the mag?

markbader (commented):
Looks already good to me. I think it would be nice to have more consistent use of `chunk_shape` and `shard_shape`.

webknossos/webknossos/dataset/_array.py
normanrz: good catch!

webknossos/webknossos/dataset/dataset.py
normanrz: This is a new constant specifically for datasets that are converted from image sequences. I think it should be set in `from_images`. Do you agree?

webknossos/tests/dataset/test_dataset.py
normanrz: Yes, that is a bit unfortunate. shard and chunk are the terms of the Zarr storage, but chunk is also used for the iterator functions. Do you have an idea how to resolve that?

webknossos/webknossos/dataset/dataset.py
markbader: Yes, I agree then. What is the reason for using (128,128,1) as `chunks_per_shard` for `from_images` and (32,32,32) for other Zarr3 datasets?

normanrz: To efficiently write Zarr data, shards need to be written in one go. `from_images` typically is used with image sequences where we convert in batches of 32 (our chunk shape). The default shard shape of 1024 \** 3 would, therefore, not be efficient. I came up with 128,128,1 because that 16k chunks in one shard, which is similar to the 32k chunk in our default shards.

markbader (approved):
Looks good to me. Regarding the naming issue with `chunk_shape` and `shard_shape` I don't have a better idea for the variable names.

PR #933, opened 2023-08-08
Remove dataformat enum.

markbader:
### Description:
- Removes unused dataformat enum class


normanrz (approved):

normanrz:
Did you check if other enums/classes are also unused?

markbader:
The other enums are actually used to enable typer to show the different options.

PR #934, opened 2023-08-08
Fix infinite loop in mag calculation during anisotropic downsampling

daniel-wer:
### Description:
- See title, it makes sense that the target mag cannot be reached while adhering to the anisotropic downsampling scheme, but it shouldn't lead to an infinite loop.
- I added respective tests which failed with a timeout error without my fix. Also, I restructured the mag calculation tests to use parametrized pytests instead of a single test with a loop, because that makes it much easier to spot which test exactly didn't pass.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/tests/dataset/test_downsampling.py
```diff
@@ -260,27 +256,75 @@ def test_anisotropic_mag_calculation() -> None:
...
-
-        assert list(reversed(sampling_scheme[:-1])) == calculate_mags_to_upsample(
-            from_mag, finest_mag, None, voxel_size
-        ), f"The calculated upsampling scheme of the {i+1}-th test case is wrong."
+    ],
+)
+def test_mag_calculation(
+    voxel_size: Optional[Tuple[float, float, float]],
+    from_mag_name: Tuple[float, float, float],
+    max_mag_name: Tuple[float, float, float],
```
MichaelBuessemeyer, line 264: why is there a `_name` at the end of the parameter names? For me it sounds like a `mag_name` would be something like `"mag_xyz"` and not a tuple of values to represent the mag? Is this a naming convention in wklibs?

webknossos/tests/dataset/test_downsampling.py
```diff
@@ -260,27 +256,75 @@ def test_anisotropic_mag_calculation() -> None:
...
+    # If a user does not specify a max_mag, then a default is calculated.
+    # Therefore, max_mag=None is not covered in this test case.
+    # The same applies for `voxel_size`:
+    # This is either extracted from the properties or set to comply with a specific sampling mode.
+
+    from_mag = Mag(from_mag_name)
+    max_mag = Mag(max_mag_name)
+
+    with pytest.raises(RuntimeError):
+        calculate_mags_to_downsample(from_mag, max_mag, None, voxel_size)
```
MichaelBuessemeyer, line 327: :rocket:

MichaelBuessemeyer (approved):
The changes look good to me :+1:. Before merging please note my question about the naming of the variables / parameters.

Do you want to take over @markbader or @philippotto as daniel is absent for now?

webknossos/tests/dataset/test_downsampling.py
markbader: I don't think there is a naming convention like this in wklibs. I changed the naming of the parameters and adapted the type specification.

daniel-wer:
@markbader Thank you for finishing this PR! :pray:

PR #904, opened 2023-05-23
Add axis order zarr

markbader:
### Description:
- Webknossos requires `axis_order` field for each mag. This field is now added by webknossos-libs.

### Issues:
- fixes #902 

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests
 - [ ] Considered adding this to the Examples

PR #935, opened 2023-08-08
add axis_order to zarr3 datasets as well

normanrz:
add axis_order to zarr3 datasets as well, follow-up to #904


markbader (approved):
LGTM

PR #937, opened 2023-08-10
Improve speed and RAM consumption of buffered slice writer

philippotto:
### Description:
- Improves RAM consumption by avoiding to concate all data sections at once. Instead, the given sections are traversed in chunks (using the shard size) and each chunk is filled so that it can be written to disk.
- As a byproduct, the speed also increased a bit. However, I'm not too sure whether this is a general win. Either way, `np.pad` is not necessary anymore, because a sufficiently sized buffer is created first into which the sections are written.
- Adds more tests for `BufferedSliceWriter`
- Adds warnings when the BufferedSliceWriter is used without aligning it to the dataset's chunk size.

As a performance benchmark I created a WKW dataset with chunk_shape=(32, 32, 32) and chunks_per_shard=(32, 32, 32). Then, I wrote random data with shape (4096, 4096, 32) to it (multiple times).

| Runs          | Avg. Time [s]      | Avg. RAM [MB]      |
| ------------- | ------------- | ------------- |
| New           | 4.76          | 1188          |
| Old           | 5.94          | 1713          |

So, the new approach only uses ~70% of the RAM for this benchmark. It should be close to 50%, but I measured the RAM with `time` which also measures the overall ram consumption of the entire python script and pytest environment in which I executed the benchmark.

### Issues:
- related to https://github.com/scalableminds/voxelytics/issues/3318

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests


normanrz:
> Runs	Avg. Time [s]	Avg. RAM [MB]
> Old	4.76	1188
> New	5.94	1713

Did you mix up old <> new?

philippotto:
> > Runs	Avg. Time [s]	Avg. RAM [MB]
> > Old	4.76	1188
> > New	5.94	1713
> 
> Did you mix up old <> new?

Oops, yes, I did. Corrected it now.

webknossos/tests/dataset/test_buffered_slice_utils.py
```diff
@@ -129,3 +136,117 @@ def test_buffered_slice_reader_along_different_axis(tmp_path: Path) -> None:
...
+            for z in range(0, shape[2]):
+                section = data[:, :, z]
+                writer.send(section)
+
+    written_data = mag1.read(absolute_offset=(0, 0, 0), size=shape)
+
+    assert np.all(data == written_data)
+
+
+def test_buffered_slice_writer_should_warn_about_unaligned_usage(
```
normanrz, line 167: So, in `test_buffered_slice_writer_along_different_axis` you aligned the offset and here are testing that unaligned screams?

normanrz (approved):
Looks good. I think the next iteration could include to read the input images in a chunked manner instead of having to read entire images at once.

webknossos/tests/dataset/test_buffered_slice_utils.py
philippotto: Yes, exactly.

philippotto:
> Looks good. I think the next iteration could include to read the input images in a chunked manner instead of having to read entire images at once.

I'm not sure if I understand correctly. The BufferedSliceWriter isn't responsible for reading images. Do you mean its interface should support sending chunks to it so that users of the Writer can read images in chunks?

normanrz:
> > Looks good. I think the next iteration could include to read the input images in a chunked manner instead of having to read entire images at once.
> 
> I'm not sure if I understand correctly. The BufferedSliceWriter isn't responsible for reading images. Do you mean its interface should support sending chunks to it so that users of the Writer can read images in chunks?

I know this isn't the responsibility of the BufferedSliceWriter. But when looking holistically at `Dataset.from_images` (and similar APIs), that would be my next angle of attack for optimization.

philippotto:
Ah ok, sorry for the misunderstanding. Sounds good :+1:

PR #943, opened 2023-09-14
DaskExecutor

normanrz:
### Description:
- Adds a DaskExecutor that can use an existing dask cluster + scheduler to distribute computation. We can use dask to easily schedule workloads across on-demand-created VMs
- Refactors `Executor` type to a protocol that should make the interface more explicit and therefore better enforceable by mypy
- Upgrades dependencies in cluster_tools and webknossos

### How to test
- Start a dask scheduler `python -m dask scheduler`
- Start at least one dask worker `python -m dask worker tcp://10.0.0.2:8786` (can be on the same machine)
- Open a Python shell and run:
```python
import webknossos as wk
from cluster_tools import get_executor

m = wk.Dataset.open("path/to/l4_sample").get_layer("color").get_mag("1")
with get_executor("dask", address="10.0.0.2:8786") as exc:
  m.map_chunk(lambda view: view.read().mean(), executor=exc)
```

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


cluster_tools/Changelog.md
```diff
@@ -10,10 +10,13 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 [Commits](https://github.com/scalableminds/webknossos-libs/compare/v0.13.6...HEAD)
 
 ### Breaking Changes
+- Dropped support for Python 3.7. [#943](https://github.com/scalableminds/webknossos-libs/pull/943)
```
philippotto, line 13: how about recommending `exc.as_completed` instead of `concurrent.futures.as_completed` since the latter won't work for dask futures, if I see this correctly?

philippotto (approved):
awesome stuff :+1: testing went also very smooth :1st_place_medal: 

there might be some potential to DRY things between the cluster executor and the dask executor, but I guess it's alright for now.

PR #950, opened 2023-10-09
Enable zarrita by default

normanrz:
### Description:
- Zarrita is always installed and now the default for Zarr and Zarr3 datasets
- The `WK_USE_ZARRITA` env variable is no longer required
- Writes out OME metadata for both Zarr and Zarr3 datasets.
- Reorganizes the imports to use relative imports in the `webknossos` package

### Todos:
 - [x] Updated Changelog


markbader (commented):
LGTM :)

markbader (approved):

PR #952, opened 2023-10-11
Zarr volume annotations

### LINKED ISSUE #951: Support Zarr3 Volume Annotation format
As of https://github.com/scalableminds/webknossos/pull/7288, webknossos will support zarr3 volume annotation downloads.

The libs should be able to read those.

Example annotation file (note that the webknossos PR is not yet reviewed so this may still slightly change):

[l4_sample__explorational__suser__3b8853.zip](https://github.com/scalableminds/webknossos-libs/files/12867664/l4_sample__explorational__suser__3b8853.zip)

normanrz:
### Description:
- Adds support for Zarr3-based volume annotations

### Issues:
- fixes #951 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


fm3:
A lot of the diff looks zarr-volume-annotation unrelated to me. Did something go wrong when switching the base branch? Or should I review the other changes as well? I think I’d need a primer there

normanrz:
> A lot of the diff looks zarr-volume-annotation unrelated to me. Did something go wrong when switching the base branch? Or should I review the other changes as well? I think I’d need a primer there

Yes, sorry the branch got messed up after merging master. I cleaned it up now.

webknossos/webknossos/annotation/annotation.py
```diff
@@ -737,32 +765,56 @@ def export_volume_layer_to_dataset(
...
+                        layer_name,
+                        category=SEGMENTATION_CATEGORY,
+                        largest_segment_id=largest_segment_id,
+                    ),
+                )
+            elif volume_layer.data_format == DataFormat.Zarr3:
+                datasource_properties = dataset_converter.structure(
+                    json.loads(data_zip.read(PROPERTIES_FILE_NAME)), DatasetProperties
+                )
+                assert len(datasource_properties.data_layers) == 1
```
fm3, line 804:
```suggestion
                assert len(datasource_properties.data_layers) == 1, f"Volume data zip must contain exactly one layer, got {len(datasource_properties.data_layers)}"
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -967,18 +980,30 @@ def add_layer_for_existing_files(
...
+            num_channels=num_channels,
+            dtype_per_channel=dtype_per_channel,
+            data_format=data_format,
             **kwargs,
         )
         for mag_dir in layer.path.iterdir():
+            try:
+                Mag(mag_dir.name)  # test if folder name is a valid mag
+            except ValueError:
+                continue
```
fm3, line 1000: could you add a comment explaining which expected cases should trigger this continue?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -954,6 +955,18 @@ def add_layer_like(self, other_layer: Layer, layer_name: str) -> Layer:
         self._export_as_json()
         return self._layers[layer_name]
 
+    def _add_existing_layer(self, layer_properties: LayerProperties) -> Layer:
+        self._ensure_writable()
+
+        assert layer_properties.name not in self.layers
```
fm3, line 961:
```suggestion
        assert layer_properties.name not in self.layers, f"Cannot import layer “{layer-properties.name}” into datasource, as a layer with this name is already present."
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -954,6 +955,18 @@ def add_layer_like(self, other_layer: Layer, layer_name: str) -> Layer:
         self._export_as_json()
         return self._layers[layer_name]
 
+    def _add_existing_layer(self, layer_properties: LayerProperties) -> Layer:
```
fm3, line 958: I feel like something like this already exists somewhere :thinking: but didn’t find it just now, so maybe I dreamt that (or it was removed in the meantime)

fm3 (commented):
Looking good already :)

webknossos/webknossos/dataset/dataset.py
normanrz: There is a method called `add_layer_for_existing_files`. However, it works differently because it tries to guess all the parameters from the data. The new method expects all parameters to be determined already.

fm3 (commented):
LGTM :) When the CI runs through, this should be good to go. I did not test again.

PR #954, opened 2023-10-18
changes from_kwargs to from_config

normanrz:
While trying out the dask executor in voxelytics, I noticed that the signature for creating an executor is not consistent. That makes this breaking change necessary. I don't think it is a big deal, because I doubt anybody else has used it in the meantime.


philippotto:
> While trying out the dask executor in voxelytics, I noticed that the signature for creating an executor is not consistent.

Can you link where else `from_config` is defined? I don't find anything like that in our repo or in the python documentation for executors? :thinking: I'm probably missing the forest for the trees.

philippotto (approved):

PR #956, opened 2023-10-19
upgrades to mypy 1.6

normanrz:
### Description:
- Upgrades mypy including some refactorings in cluster_tools

### Todos:
 - [x] Updated Changelog


cluster_tools/cluster_tools/executors/dask.py
```diff
@@ -64,10 +64,13 @@ def submit(  # type: ignore[override]
             ]
             del kwargs["__cfut_options"]
 
-            __fn = partial(
-                MultiprocessingExecutor._execute_and_persist_function,
-                output_pickle_path,
-                __fn,
+            __fn = cast(
+                Callable[_P, _T],
```
normanrz, line 68: mypy doesn't understand `partial` yet. But, now it understands enough about `Callable`s to complain.

cluster_tools/cluster_tools/executors/multiprocessing_.py
```diff
@@ -113,37 +113,32 @@ def submit(  # type: ignore[override]
         else:
             submit_fn = super().submit  # type: ignore[assignment]
 
-        # Depending on the start_method and output_pickle_path, wrapper functions may need to be
```
normanrz, line 116: I believe this comment was not accurate anymore for a while, because we haven't really used the call stack to hold functions and arguments; only the partial functions. Now, I dropped the call stack altogether in favor of partials.

webknossos/webknossos/dataset/properties.py
```diff
@@ -214,7 +214,7 @@ def mag_unstructure(mag: Mag) -> List[int]:
                 a.name: override(
                     omit_if_default=True, rename=snake_to_camel_case(a.name)
                 )
-                for a in attr.fields(cls)  # pylint: disable=not-an-iterable
+                for a in attr.fields(cls)  # type: ignore[misc]
```
normanrz, line 217: 🤷

cluster_tools/cluster_tools/executors/multiprocessing_.py
philippotto: I think, the comment is still accurate and the mentioned functions and arguments mean the partial functions. However, the new code using partials should be equivalent as far as I can see. 
cc'ing @daniel-wer who implemented this in https://github.com/scalableminds/webknossos-libs/commit/e773c22ffc86d222a9b57dad77b6494993907955.

philippotto (approved):
lgtm :+1:

cluster_tools/cluster_tools/executors/multiprocessing_.py
daniel-wer: Also, I think you reversed the order the setup functions are called in. It might be beneficial to setup the logging, first.

PR #958, opened 2023-10-24
Set default data format with enviroment variable

sm1lla:
### Description:
- The `DEFAULT_DATA_FORMAT` used when writing datasets can now be set with the environment variable `DATA_FORMAT`. If the environment variable is not set the default data format is still wkw. When the data format is set to Zarr3 a lower value is set for `DEFAULT_CHUNKS_PER_SHARD`.

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/dataset/defaults.py
```diff
@@ -1,13 +1,19 @@
+import os
+
 from ..geometry import Vec3Int
 from .data_format import DataFormat
 
 DEFAULT_WKW_FILE_LEN = 32
 DEFAULT_CHUNK_SHAPE = Vec3Int.full(32)
-DEFAULT_CHUNKS_PER_SHARD = Vec3Int.full(32)
+DEFAULT_DATA_FORMAT = (
+    DataFormat.WKW if not os.environ.get("DATA_FORMAT") == "zarr3" else DataFormat.Zarr3
```
normanrz, line 9:
```suggestion
    DataFormat.WKW if not os.environ.get("WK_DEFAULT_DATA_FORMAT") == "zarr3" else DataFormat.Zarr3
```

webknossos/webknossos/dataset/defaults.py
```diff
@@ -1,13 +1,19 @@
...
+)
+DEFAULT_CHUNKS_PER_SHARD = (
+    Vec3Int.full(32) if DEFAULT_DATA_FORMAT == DataFormat.WKW else Vec3Int.full(16)
```
normanrz, line 12: I would make a separate env var `WK_DEFAULT_CHUNKS_PER_SHARD` for this.

normanrz (changes requested):

webknossos/webknossos/dataset/defaults.py
```diff
@@ -1,13 +1,21 @@
...
+
 from ..geometry import Vec3Int
 from .data_format import DataFormat
 
 DEFAULT_WKW_FILE_LEN = 32
 DEFAULT_CHUNK_SHAPE = Vec3Int.full(32)
-DEFAULT_CHUNKS_PER_SHARD = Vec3Int.full(32)
+DEFAULT_DATA_FORMAT = (
+    DataFormat.WKW
+    if not os.environ.get("WK_DEFAULT_DATA_FORMAT") == "zarr3"
```
normanrz, line 10: Could you parse the string directly into the `DataFormat` enum? Probably just `DataFormat(os.environ["WK_DEFAULT_DATA_FORMAT"]) if "WK_DEFAULT_DATA_FORMAT" in os.environ else DataFormat.WKW`

webknossos/webknossos/dataset/defaults.py
```diff
@@ -1,13 +1,21 @@
...
+    else DataFormat.Zarr3
+)
+DEFAULT_CHUNKS_PER_SHARD = Vec3Int.full(
```
normanrz, line 13: Ideally, the env var could be a scalar for `Vec3Int.full()` or a comma-separated 3-tuple for `Vec3Int()`.

normanrz (changes requested):
Please also add a changelog entry for this.

webknossos/webknossos/geometry/vec3_int.py
```diff
@@ -72,6 +72,12 @@ def from_vec_or_int(vec_or_int: Union["Vec3IntLike", int]) -> "Vec3Int":
         else:
             return Vec3Int(vec_or_int)
 
+    @staticmethod
+    def from_str(string: str):
+        s = eval(string)
```
normanrz, line 77: Please don't use `eval`. That is a security vulnerability because it can execute arbitrary code.

Instead you could use `tuple(int(a.strip()) for a in string.split(","))`

normanrz (approved):

PR #959, opened 2023-10-25
More dask features

normanrz:
### Description:
- Adds SIGINT handling to DaskExecutor
- Adds support for resources (e.g. `mem`, `cpus`) to DaskExecutor
- Runs tasks in their own process to not block the GIL, which dask workers need internally to communicate with the scheduler; also propagates env variables to the task processes
- The address for the dask `Client` can be configured via `DASK_ADDRESS` env var (should we rename that?)

### Todos:
 - [x] Updated Changelog


cluster_tools/cluster_tools/executors/dask.py
```diff
@@ -28,23 +33,103 @@
...
+        if self.job_resources is not None:
+            # `mem` needs to be a number for dask, so we need to parse it
+            if "mem" in self.job_resources:
+                self.job_resources["mem"] = _parse_mem(self.job_resources["mem"])
+            if "cpus-per-task" in self.job_resources:
+                self.job_resources["cpus"] = int(
+                    self.job_resources.pop("cpus-per-task")
+                )
+
+        # Clean up if a SIGINT signal is received. However, do not interfere with the
```
normanrz, line 111: This is copied from `ClusterExecutor`. It seems that the SIGINT handler is never unattached. I would've expected that to happen in `shutdown`.

cc @daniel-wer

daniel-wer: Hmm, I don't remember why I didn't add that. However, thinking about it, this could get tricky. Assuming the signal handler might have been changed in the mean time, I'm not sure how a proper cleanup would look like or is possible at all :thinking:

cluster_tools/cluster_tools/executors/dask.py
```diff
@@ -28,23 +33,103 @@
...
+    ret = q.get(timeout=0.1)
+    if "exception" in ret:
+        raise ret["exception"]
+    else:
+        return ret["value"]
+
+
+def _parse_mem(size: str) -> int:
+    units = {"": 1, "K": 2**10, "M": 2**20, "G": 2**30, "T": 2**40}
+    m = re.match(r"^([\d\.]+)\s*([a-zA-Z]{0,3})$", str(size).strip())
```
daniel-wer, line 69: Either only the keys of the `units` dict should be matched here or there should be a check that the parsed unit is in units, later.

daniel-wer (approved):
Code LGTM :+1: 

Is the DaskExecutor tested in the CI or did you test it manually?

cluster_tools/cluster_tools/executors/dask.py
normanrz: Maybe we should use a weakref to avoid a larger memory leak through the executor.

normanrz:
> Is the DaskExecutor tested in the CI or did you test it manually?

Well, there were tests but the CI didn't run them. Now, they do.

cluster_tools/tests/test_all.py
```diff
@@ -328,7 +331,8 @@ def run_map(executor: cluster_tools.Executor) -> None:
         assert list(result) == [4, 9, 16]
 
     for exc in get_executors():
-        run_map(exc)
+        if not isinstance(exc, cluster_tools.DaskExecutor):
```
daniel-wer, line 334: Why is this specific test excluded for the DaskExecutor?

daniel-wer (approved):
Nice, LGTM :+1:

cluster_tools/cluster_tools/executors/dask.py
daniel-wer: Clever solution :+1:

cluster_tools/tests/test_all.py
normanrz: Futures of the DaskExecutor become invalid when the executor is closed. This makes this test invalid. I was thinking about removing this test or making this test fail for all executors (probably a bit of effort).
cc @philippotto

philippotto: My thoughts:
- yes, we should strive for similar behavior between dask and slurm. --> (A) the futures should become invalid for the slurm context, too OR (B) (maybe I'd prefer this?) we wrap/copy the results in/into different future objects that survive the context termination. Or is there a benefit in letting the futures die? the copying could be done upon context exit.
- if we do (A), this would be a breaking change (and likely needs fixing in vx etc). therefore, I'd tackle this in a separate PR.
- either way, the test itself should not be removed without replacement. I think, what the test intends to assert is that the iterator that is returned by `map` contains futures that were kicked off before the iterator is consumed (read the [comment here](https://github.com/scalableminds/webknossos-libs/blob/b688c846d6f3b5074db56f60a5212f8bf4ac61a4/cluster_tools/cluster_tools/schedulers/cluster_executor.py#L617-L621)). essentially, this is covering an implementation detail, but the overall expected behavior is that that the map call *eagerly* submits all futures, but *lazily* awaits its results (so that they don't need to be in RAM all at once). the test exploits the use-futures-after-context-was-shutdown-behavior to test the eager submit (if it was not eager, the test would fail because the submit would be after context exit). if you remove that behavior, the eager submits should still be checked for in my opinion.

I hope this is somewhat comprehensible. If not, let's have a call :call_me_hand:

normanrz: I believe dask transfers the data lazily from the workers or scheduler. That doesn't work anymore, once the client closes. We could wrap the futures to eagerly collect the data.

philippotto: > We could wrap the futures to eagerly collect the data.

Yes, either this, or try to hook into the closing client (so that collection is done in the last moment where it's possible).

normanrz: I would leave that for a followup and merge this as is. Ok?

philippotto: sure :+1:

PR #969, opened 2023-11-28
Fix nested Slurm scheduling for newer Slurm versions

daniel-wer:
### Description:
- Slurm version >=22.05 changed the behavior of the `SLURM_CPU_BIND` variable leading to errors when scheduling nested Slurm jobs
- I added a test and confirmed that the issue was reproducible before this PR
- See https://scm.slack.com/archives/C03B6160ATY/p1700666715939549 for more context

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/cluster_tools/tests/test_slurm.py
+++ b/cluster_tools/tests/test_slurm.py
@@ -533,3 +533,15 @@ def test_preliminary_file_map() -> None:
                 assert (
                     not preliminary_output_path.exists()
                 ), "Preliminary output file should not exist anymore"
+
+
+def test_cpu_bind_regression() -> None:
+    os.environ[
+        "SLURM_CPU_BIND"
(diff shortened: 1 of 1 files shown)


cluster_tools/dockered-slurm/slurm.conf
```diff
@@ -36,7 +36,7 @@ ReturnToService=0
 #SrunEpilog=
 #TaskProlog=
 #TaskEpilog=
-#TaskPlugin=
+TaskPlugin=task/affinity
```
philippotto, line 39: can you add a comment that explains why this is set?

daniel-wer: I added a comment and also added an assertion to the test, checking that the plugin is enabled.

PR #970, opened 2023-12-04
Fix parsing ApiProject in wk api client

fm3:
### Description:
- Parsing the response of the project info json failed. Needs to be ApiUserCompact (uses the backend’s compactWrites method)

### Todos:
 - [x] Updated Changelog


hotzenklotz (approved):
Tested and works.

PR #972, opened 2023-12-12
Project owner is optional (None if you have no read access rights)

fm3:
Normal users/annotators can read the api project but should not get info about the project owner. There it is null/None. This PR reflects that in the libs.

### Todos:
 - [x] Updated Changelog


MatthisCl (approved):
LGTM!

PR #973, opened 2023-12-19
Fix buffered slice writer overwriting data under certain conditions

philippotto:
### Description:
- Fixes that the buffered slice writer could overwrite data when writing less slices than `buffer_size` at an offset that is not aligned.

### Issues:
- fixes a regression caused by #937

### Todos:
 - [ ] Updated Changelog
 - [X] Added / Updated Tests


daniel-wer (approved):
LGTM, nice fix :+1: It's great that you added the test, thank you!

PR #975, opened 2024-01-08
Add calculation of a dataset's whole bounding box to Dataset

MichaelBuessemeyer:
### Description:
This PR adds a method to the dataset class that calculates the bounding box of the dataset (the bounds of all data layers). Previously, users had to iterate over all layers and extend the bounding boxes themselves. This is now solved by the new function `calculate_bounding_box`.

### Issues:
- No issue was created for this :D

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation <-- Not needed imo. (besides the doc string)
 - [x] Added / Updated Tests
 - [ ] Considered adding this to the Examples -> I think the use case is not common enough for this.

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -15,6 +15,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Breaking Changes
 
 ### Added
+- Added a method to the Datasets class that calculates a dataset's bounding box covering all layers. [#975](https://github.com/scalableminds/webknossos-libs/pull/975)
 
 ### Changed
 
--- a/webknossos/tests/dataset/test_dataset.py
+++ b/webknossos/tests/dataset/test_dataset.py
@@ -1052,6 +1052,21 @@ def test_changing_layer_bounding_box(
     assure_exported_properties(ds)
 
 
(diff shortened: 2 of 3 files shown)


MichaelBuessemeyer:
@philippotto Could you please review this PR?

While working on [#3376](https://github.com/scalableminds/voxelytics/issues/3376) I stumbled upon that no function in the dataset class existed that simply returned the whole dataset's bounding box. As it sounded like a useful feature/method, I created this PR.

Do you think are the tests sufficient and that no further documentation is needed?

philippotto (approved):
good stuff!

hotzenklotz:
For volume annotations there is also [this code](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/_utils/infer_bounding_box_existing_files.py). Is this related or does this solve a different problem?

philippotto:
> For volume annotations there is also [this code](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/_utils/infer_bounding_box_existing_files.py). Is this related or does this solve a different problem?

yes, this is a different problem. it only works for a single layer / mag view and does the inferral by inspecting the existent shards on disk.

PR #996, opened 2024-02-08
CLI tool for local merging of zip annotations with fallback layer

erjel:
Hey, 

I would like to provide code for a CLI tool which merges annotation zip files with local fallback layers. The main benefit would be that users which have file-level access to their dataset can merge annotations with fallback layer without streaming the entire merged dataset via the webknossos web server (and using cluster executors to speed up the process). Another benefit is that by pushing this upstream is the better maintainability: The existing gist is not integrated in the CI pipeline and thus suffers from several breaking changes already.  

I am happy to further change the code according to match additional requirements (i.e. provide additional test cases/ change documentation).

Best wishes,
Eric

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/cli/merge_fallback.py
+++ b/webknossos/webknossos/cli/merge_fallback.py
@@ -35,7 +35,7 @@ def main(
             parser=parse_path,
         ),
     ],
-    dataset_folder: Annotated[
+    dataset_directory: Annotated[
         Any,
         typer.Argument(
             help="Path to your WEBKNOSSOS dataset folder.",
@@ -105,7 +105,7 @@ def main(
         annotation.export_volume_layer_to_dataset(output_dataset)
 
     else:
-        fallback_dataset_path = dataset_folder / annotation.dataset_name
+        fallback_dataset_path = dataset_directory / annotation.dataset_name
         fallback_layer = Dataset.open(fallback_dataset_path).get_layer(
             fallback_layer_name
         )
--- a/webknossos/webknossos/cli/merge_fallback.py
+++ b/webknossos/webknossos/cli/merge_fallback.py
@@ -2,23 +2,19 @@
 
 import logging
 from argparse import Namespace
-from functools import partial
 from multiprocessing import cpu_count
 from typing import Any, List, Optional, Tuple
 
 import typer
 from cluster_tools import Executor
-from typing_extensions import Annotated
 from rich.progress import track
-
+from typing_extensions import Annotated
 
 from ..annotation import Annotation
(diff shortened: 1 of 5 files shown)


hotzenklotz:
Thanks, Eric, or the contribution. It is much appreciated. We will review your code over the coming days. 

Do you mind adding a an entry to the `Unreleased` section of the `Changelog.md` file. Thanks.

webknossos/webknossos/cli/merge_fallback.py
```diff
@@ -0,0 +1,200 @@
...
+    ],
+    source_annotation: Annotated[
+        Any,
+        typer.Argument(
+            help="Path to your WEBKNOSSOS zip annotation",
+            show_default=False,
+            parser=parse_path,
+        ),
+    ],
+    dataset_folder: Annotated[
```
fm3, line 38:
```suggestion
    dataset_path: Annotated[
```
(We reserve the word folder for the virtual folder structure in the webknossos dashboard UI. Any on-disk paths are called directory/path)

fm3 (commented):
Nice addition, thanks! I agree that this belongs in this repo and not a gist.

I added one suggestion concerning parameter naming.

I’m leaving final approval to @markbader or @philippotto 

Just a heads up: Note that we may change the implementation in the future, but this does not require action now.
 - Maybe we can find a good spot for this code in the Annotation/Dataset classes
 - Volume annotation downloads will soon be changed to a zarr-based format, which the libs need to be able to read
 - In case the target dataset is not present locally, zarr-streaming may be used to do this merging operation instead. However, that is still fairly slow currently, so not really viable for large datasets, compare https://github.com/scalableminds/webknossos/issues/7438

webknossos/webknossos/cli/merge_fallback.py
erjel: changed the variable name accordingly.

erjel:
> Maybe we can find a good spot for this code in the Annotation/Dataset classes

I thought of this as well:
- `merge_mags()` can easily become a member function of `MagView` similar to `compress`
- `merge_chunk()`  would then become the equivalent of `_compress_cube_job()` in the same file.

but I didn't want to interfere with your internal decisions about the overall architecture. If you would like to have them there, I can modify the PR accordingly.

markbader (approved):
LGTM. Thanks a lot for your contribution @erjel. The CLI part works well, and we will move the functional part within the next days.

PR #1013, opened 2024-03-11
Add Ruff

hotzenklotz:
### Description:
- PR adds ruffs to WK-Libs, replacing `black` + `iSort` as formatters and `pylint` as a linter
- I mostly used the standard config of ruff as described here: https://docs.astral.sh/ruff/configuration/
- Only added Ruff to wk-libs subfolder so far. If we are happy with this config, then we can follow up for `cluster tools` 
- @reviewer: I tried to contain the automatic linting fixes that ruff performed into a single commit. Please review my manual linting fixes individually. See commit dceb636b92128701d983ab95113e01f935a8b3a8 

# TODO
- [x] remove any remaining pylint stuff

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


webknossos/format.sh
```diff
@@ -2,9 +2,8 @@
 set -eEuo pipefail
 
 if [ $# -eq 1 ] && [ "$1" = "check" ]; then
-    poetry run black --check .
-    poetry run isort --check-only .
+    poetry run ruff format --check .
 else
-    poetry run isort .
-    poetry run black .
+    poetry run ruff check --fix .
```
hotzenklotz, line 7: This is my replacement for `isort`, i.e. automatic fixing of the import linting errors. This, however, will list and fix (if possible) any linting errors. It is fast so why not...

webknossos/lint.sh
```diff
@@ -1,8 +1,4 @@
 #!/usr/bin/env bash
 set -eEuo pipefail
 
-# Pylint doesn't lint files in directories that don't have an __init__.py
-# This will maybe be fixed by https://github.com/PyCQA/pylint/issues/352
-# In the meantime, find all python files, except for the ./webknossos/client/_generated directory and lint them
-# Inspired by https://stackoverflow.com/questions/4210042/how-to-exclude-a-directory-in-find-command
-find . -type d \( -path ./webknossos/client/_generated -o -path ./tests/.webknossos-server -o -path ./stubs \) -prune -o -iname "*.py" -print | xargs poetry run python -m pylint -j2
+poetry run ruff check --fix .
```
hotzenklotz, line 4: This command is already included in the `./format.sh` command. Perhaps we should get rid of this file entirely. What do you think?

webknossos/pyproject.toml
```diff
@@ -75,16 +75,13 @@ all = ["pims","tifffile","imagecodecs","JPype1","pylibCZIrw"]
...
 [tool.poetry.dev-dependencies]
 # autoflake
-black = "^23.1.0"
 dask = { version = "^2023.9.1", python = ">=3.9", extras = ["distributed"] }
 hypothesis = "^6.35.0"
 icecream = "^2.1.1"
 inducoapi = "^2.0.2"
-isort = "^5.9.3"
 jsonschema = "^4.6.0"
-mypy = "^1.0.0"
```
hotzenklotz, line 85: We were running `mypy 1.6.x` anyway, so I fixed that as a minimum requirement. The current version is 1.9.x.

webknossos/webknossos/cli/_utils.py
```diff
@@ -17,7 +17,7 @@
 class DistributionStrategy(str, Enum):
     """Enum of available distribution strategies.
 
-    TODO  pylint: disable=fixme
+    TODO
```
philippotto, line 20: it might make sense to add https://docs.astral.sh/ruff/rules/line-contains-todo/. this would catch unintentional todo comments in PRs before merging.

webknossos/webknossos/dataset/_downsampling_utils.py
```diff
@@ -314,7 +314,7 @@ def downsample_cube_job(
         tiles = product(
             *(
                 list(range(0, math.ceil(len / buffer_edge_len)))
-                for len, buffer_edge_len in zip(shape[-3:], buffer_shape)
+                for len, buffer_edge_len in zip(shape[-3:], buffer_shape) # noqa:  A001 Variable `len` is shadowing a Python builtin
```
philippotto: fix directly by renaming to length? should only be used in 316 and 317.

webknossos/webknossos/dataset/_upsampling_utils.py
```diff
@@ -50,7 +50,7 @@ def upsample_cube_job(
             *list(
                 [
                     list(range(0, math.ceil(len)))
-                    for len in target_size.to_np() / buffer_shape.to_np()
+                    for len in target_size.to_np() / buffer_shape.to_np() #  noqa: A001 Variable `len` is shadowing a Python builtin
```
philippotto: same here

webknossos/webknossos/skeleton/tree.py
```diff
@@ -114,7 +114,7 @@ def __init__(
         group: "Group",
         skeleton: "Skeleton",
         color: Optional[Vector4] = None,
-        enforced_id: Optional[int] = None,  # pylint: disable=unused-argument
```
philippotto, line 117: is https://docs.astral.sh/ruff/rules/unused-function-argument/ not enabled? I think, this rule is important because it highlights possible smells. in this case, it's probably due to inheritance..

philippotto (commented):
very cool :) only left some minor comments.

webknossos/webknossos/cli/_utils.py
hotzenklotz: Done

webknossos/webknossos/dataset/_downsampling_utils.py
hotzenklotz: Done

webknossos/webknossos/skeleton/tree.py
hotzenklotz: Done

webknossos/lint.sh
hotzenklotz: @philippotto Any thoughts on this?

webknossos/format.sh
philippotto: Does `poetry run ruff check --select I --fix .` work? copied from [here](https://docs.astral.sh/ruff/formatter/#sorting-imports). I think, I would prefer to keep formatting and linting separated if easily possible.

webknossos/lint.sh
philippotto: I just commented [here](https://github.com/scalableminds/webknossos-libs/pull/1013/files/d25afc9bc1ca7d45fbd6888ca552ce1ac8d955c2#r1524619872). if it's not possible to separate linting and formatting, a single file would make sense (`format-and-lint.sh`).

webknossos/format.sh
hotzenklotz: For an earlier version I had the same idea and this works. I applied your suggestion on my latest commit.

philippotto (approved):
awesome :man_dancing:

PR #1016, opened 2024-03-12
Sanitize layer names in from_images

markbader:
### Description:
- in WEBKNOSSOS layer names have to match the pattern  [A-Za-z0-9_\\-\\.]*. This wasn't ensured by wklibs yet. This PR adds a check for correct layer names in the `add_layer` method and sanitizes the generated layer names of `from_images`.

### Todos:
 - [x] Updated Changelog
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -103,6 +103,8 @@ _DATASET_URL_REGEX = re.compile(
     + r"(?P<organization_id>[^/]*)/(?P<dataset_name>[^/]*)(/(view)?)?"
     + r"(\?token=(?P<sharing_token>[^#]*))?"
 )
+_ALLOWED_LAYER_NAME_REGEX = re.compile(r"^[A-Za-z0-9_\-]+[A-Za-z0-9_\-\.]*$")
+_UNALLOWED_LAYER_NAME_CHARS = re.compile(r"[^A-Za-z0-9_\-\.]")
 
 
 def _find_array_info(layer_path: Path) -> Optional[ArrayInfo]:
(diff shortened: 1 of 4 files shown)


webknossos/webknossos/dataset/dataset.py
```diff
@@ -777,6 +792,10 @@ def add_layer(
 
         self._ensure_writable()
 
+        assert _ALLOWED_LAYER_NAME_REGEX.match(
+            layer_name
+        ), f"The layer name '{layer_name}' is invalid. It must only contain letters, numbers, underscores, hyphens and dots."
+
```
fm3, line 798: could this also check the does-not-start-with-dort condition?

markbader: Yes, the _ALLOWED_LAYER_NAME_REGEX checks that at least one allowed character that is not a dot has to be at the beginning of the string.

webknossos/webknossos/dataset/dataset.py
```diff
@@ -103,6 +103,8 @@
     + r"(?P<organization_id>[^/]*)/(?P<dataset_name>[^/]*)(/(view)?)?"
     + r"(\?token=(?P<sharing_token>[^#]*))?"
 )
+_ALLOWED_LAYER_NAME_REGEX = re.compile(r"^[A-Za-z0-9_\-]+[A-Za-z0-9_\-\.]*$")
```
hotzenklotz, line 106: Please add a comment explaing this REGEX. Something like in your PR description. Which characters are actually allowed

markbader: Done.

hotzenklotz (approved):

fm3 (approved):

Issue #1029, opened 2024-04-02
Investigate support for nd tifffiles with tifffile dependency


markbader:
## Detailed Description
- Affected library: e.g. webknossos
- The current default PIMS reader for tiff image files, does not provide all information about additional axes in the PIMS specific iter_axes and bundle_axes format. It stores them in an `_tiff` variable. It would be useful to adapt that behavior by adding a PIMS reader class that handles these nd cases or making the default tiff reader skip images that are not handled properly to give them to the bioformats reader. This is useful to avoid using the `use_bioformats` flag for images with more than 3 dimensions.

## Use Cases & Context
- we added nd support to wklibs but reading tifffiles is currently only supported with `use_bioformats` set to `True`. It would be great to avoid that, to simplify the image upload worker.

Issue #960, opened 2023-10-27
Dataset.from_images should correctly handle 3D tiff


philippotto:
## Context
Conversion of tiff datasets to wkw (or zarr).

## Expected Behavior
Converting one 3D tiff should output a 3D dataset.

## Current Behavior
Converting one 3D tiff outputs a 2D dataset (only 1 section).

## Steps to Reproduce the bug
1. See dataset [here](https://scm.slack.com/archives/C02H5T8Q08P/p1698395309748159)
2. Try to convert it with Dataset.from_images

PR #1043, opened 2024-04-18
Tiffreader for nd data

markbader:
### Description:
- To avoid the behavior of the default PIMs tiff reader, this PR adds a tiff reader based on the tifffile library. While the default tiff reader always produces 3d arrays, the custom reader creates ndarrays with a shape corresponding to the metadata of the tiff.

- fixes: #1029 
- fixes: #960

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/Changelog.md
```diff
@@ -13,6 +13,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 [Commits](https://github.com/scalableminds/webknossos-libs/compare/v0.14.22...HEAD)
 
 ### Breaking Changes
+- Preferring a custom tiff reader over the default PIMS reader to convert tiff files. This enables the recognition of axis information and the support of tifffiles with more than 3 dimensions. Nonetheless this leads to changed behavior when converting tiff files. Tiffs with axes other than c, x, y and z are no longer supported for conversion to WKW, please convert these to Zarr3 Datasets. [1043](https://github.com/scalableminds/webknossos-libs/pull/1043)
```
normanrz, line 16: Does converting to WKW still work if there is an additional axis (e.g. t) but with just 1 slice?

webknossos/webknossos/dataset/_utils/pims_images.py
```diff
@@ -165,6 +169,7 @@ def __init__(
                 # An image slice should always consist of a 2D image. If there are multiple channels
                 # the data of each channel is part of the image slices. Possible shapes of an image
                 # slice are (#y_shape, #x_shape), (1, #y_shape, #x_shape) or (3, #y_shape, #x_shape).
+                # self._bundle_axes = list(images.bundle_axes)
```
normanrz, line 172:
```suggestion
```

webknossos/tests/dataset/test_add_layer_from_images.py
normanrz, line 1: Could we add more test images with axes configurations?

webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
```diff
@@ -0,0 +1,116 @@
...
+    raise RuntimeError(
+        "Cannot import pims, please install it e.g. using 'webknossos[all]'"
+    ) from e
+
+try:
+    import tifffile
+except ImportError as e:
+    raise RuntimeError(
+        "Cannot import tifffile, please install it e.g. using 'webknossos[tifffile]'"
+    ) from e
```
normanrz, line 20: If both pims and tifffile are missing, it would be nice to just have one error message that tells users to install `webknossos[pims,tifffile]`.

webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
```diff
@@ -0,0 +1,116 @@
...
+        assert _tmp is not None, "No pages found in tiff file."
+        self._dtype = _tmp.dtype or np.dtype("uint8")
+        self._shape = _tmp.shape
+        with tempfile.TemporaryFile() as tmp:
+            self._memmap = np.memmap(
+                tmp,
+                dtype=self._dtype,
+                shape=_tiff.shape,
+                mode="w+",
+            )
```
normanrz, line 61: Why is the memmap in a tmp file? Isn't that creating a copy of the image?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -663,38 +663,38 @@ def from_images(
...
-
-        for layer_name, filepaths in filepaths_per_layer.items():
-            filepaths.sort(key=z_slices_sort_key)
-            category: LayerCategoryType
-            if layer_category is None:
-                category = (
-                    "segmentation"
-                    if guess_if_segmentation_path(filepaths[0])
-                    else "color"
+        with get_executor_for_args(None, executor) as executor:
```
normanrz, line 666: Can the executor be reused for multiple layers?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1289,7 +1289,7 @@ def add_layer_from_images(
                 set(layer.bounding_box.axes).difference("x", "y", "z")
             ) and layer.data_format != DataFormat.Zarr3:
```
normanrz, line 1290: `DataFormat.Zarr` should also be fine, no?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1289,7 +1289,7 @@ def add_layer_from_images(
                 set(layer.bounding_box.axes).difference("x", "y", "z")
```
normanrz, line 1289: See comment above. I think we should check whether additional axes only have a length of 1. Then these axe could be collapsed.

webknossos/Changelog.md
markbader: Currently, it is not supported. I can add a logic for that.

webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
markbader: Yes, I tried to use the tiff file memmap function to don't create a copy, but it only works for a few tiff files. OME tiffs as our example data for a 4d image are not memory-mappable, according to the tifffile library.

webknossos/webknossos/dataset/dataset.py
markbader: I added this line, because I ran into an issue like “Cannot schedule new futures after shutdown”. Reusing the executor seems to fix it and enhances the error log.

webknossos/Changelog.md
markbader: I implemented a logic that sets the bounding box to 3d if additional axes have a shape of 1. I couldn't find fitting test data for that case so far.

webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
markbader: Sure, I can catch the error thrown be the tifffile library and use my code as a fallback.

webknossos/tests/dataset/test_add_layer_from_images.py
markbader: Sure. Any suggestions for good test data, or should I generate some artificial tiff files?

webknossos/webknossos/dataset/dataset.py
markbader: It didn't work before, but I fixed the issue and changed this part in code, to allow Zarr and Zarr3.

webknossos/tests/dataset/test_add_layer_from_images.py
markbader: I added `Multi channel Z series` and `Single channel` to the testdatasets and mentioned the ome consortium in the readme.

webknossos/webknossos/dataset/_array.py
```diff
@@ -593,7 +593,11 @@ def create(cls, path: Path, array_info: ArrayInfo) -> "ZarritaArray":
             ArrayV2.create(
                 store=path,
                 shape=(array_info.shape),
-                chunks=(array_info.num_channels,) + array_info.chunk_shape.to_tuple(),
+                chunks=(array_info.num_channels,)
+                + tuple(
```
normanrz, line 597: Please add a comment what is happening here.

normanrz (approved):
Great work!

webknossos/webknossos/dataset/_array.py
markbader: Done.

PR #1052, opened 2024-04-23
Adapt BufferedSliceWriter, to enable configuration after initialization.

### LINKED ISSUE #968: Add set_z method to BufferedSliceWriter for lazy initialization
## Detailed Description
At the moment, the BufferedSliceWriter has to be instantiated with a z offset which cannot be changed afterwards. Changing the z offset after instantiation should be possible, though. I'd differentiate the following cases:

1) Changing the offset before any write was done. This should be easy since nothing has been written yet.
2) Changing the offset after a flush was done. This should also be relatively easy, since the data was already flushed.
3) Changing the offset in a state where unflushed data exists. In this case, we could force a flush and then reset the offset.

## Use Cases & Context
My use case would be covered by only tackling (1). In my case, I want to create a buffered slice writer and then set the z shortly before the first write is done. This would help to simplify some code in VX (see [this comment](https://github.com/scalableminds/voxelytics/pull/3383#discussion_r1407575399)).

markbader:
### Description:
- This PR adapts the BufferedSliceWriter, to allow resetting the offsets after initialization. 

### Issues:
- fixes #968 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/tests/dataset/test_buffered_slice_utils.py
+++ b/webknossos/tests/dataset/test_buffered_slice_utils.py
@@ -312,11 +312,14 @@ def test_buffered_slice_writer_reset_offset(tmp_path: Path) -> None:
 
         # Write some slices
         with mag1.get_buffered_slice_writer() as writer:
-            writer.reset_offset(absolute_offset=(8, 16, 8))
-            for z in range(0, shape[2]):
+            for z in range(0, shape[2] - 8):
+                section = data[:, :, z]
+                writer.send(section)
+            writer.reset_offset(absolute_offset=(0, 0, shape[2] - 8))
+            for z in range(shape[2] - 8, shape[2]):
                 section = data[:, :, z]
                 writer.send(section)
 
-    written_data = mag1.read(absolute_offset=(8, 16, 8), size=shape)
+    written_data = mag1.read(absolute_offset=(0, 0, 0), size=shape)
 
     assert np.all(data == written_data)
--- a/webknossos/tests/dataset/test_buffered_slice_utils.py
+++ b/webknossos/tests/dataset/test_buffered_slice_utils.py
@@ -315,11 +315,20 @@ def test_buffered_slice_writer_reset_offset(tmp_path: Path) -> None:
             for z in range(0, shape[2] - 8):
                 section = data[:, :, z]
                 writer.send(section)
-            writer.reset_offset(absolute_offset=(0, 0, shape[2] - 8))
+            writer.reset_offset(absolute_offset=(0, 0, shape[2]))
             for z in range(shape[2] - 8, shape[2]):
                 section = data[:, :, z]
                 writer.send(section)
 
-    written_data = mag1.read(absolute_offset=(0, 0, 0), size=shape)
+    written_data_before_reset = mag1.read(
+        absolute_offset=(0, 0, 0), size=(shape[0], shape[1], shape[2] - 8)
+    )
+    written_data_after_reset = mag1.read(
+        absolute_offset=(0, 0, shape[2]), size=(shape[0], shape[1], 8)
+    )
+
+    written_data = np.concatenate(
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -238,29 +231,77 @@ def _flush_buffer(self) -> None:
...
+                "[DEPRECATION] Using offset for a buffered slice writer is deprecated. "
+                + "Please use the parameter relative_offset or absolute_offset in Mag(1) instead.",
+                DeprecationWarning,
+            )
+
+        if offset is not None:
+            self._bbox = BoundingBox(
+                self._view.bounding_box.topleft_xyz + Vec3Int(offset) * self._view.mag,
+                Vec3Int.zeros(),
+            )
```
philippotto, line 285: isn't this identical to the code in __init__? can this by DRY-ed?

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -238,29 +231,77 @@ def _flush_buffer(self) -> None:
...
+    def reset_offset(
+        self,
+        offset: Optional[Vec3IntLike] = None,  # deprecated, relative in current mag
+        relative_offset: Optional[Vec3IntLike] = None,  # in mag1
+        absolute_offset: Optional[Vec3IntLike] = None,  # in mag1
+        relative_bounding_box: Optional[NDBoundingBox] = None,  # in mag1
+        absolute_bounding_box: Optional[NDBoundingBox] = None,  # in mag1
+    ) -> None:
+        if getattr(self, "_slices_to_write", False):
+            self._flush_buffer()
```
philippotto, line 264: It's a bit weird to use getattr here because the propert should always exist. I assume it is because self._slices_to_write is initialized below the reset_offset call. However, it can probably simply be moved above that?

webknossos/tests/dataset/test_buffered_slice_utils.py
```diff
@@ -293,3 +293,30 @@ def test_basic_buffered_slice_writer_multi_shard_multi_channel(tmp_path: Path) -
...
+
+    with warnings.catch_warnings():
+        warnings.filterwarnings("error")  # This escalates the warning to an error
+
+        # Write some slices
+        with mag1.get_buffered_slice_writer() as writer:
+            writer.reset_offset(absolute_offset=(8, 16, 8))
+            for z in range(0, shape[2]):
+                section = data[:, :, z]
+                writer.send(section)
```
philippotto, line 318: I'd adapt the test in two ways:
- write a bit less than chunk_size in z (e.g., shape[2] - 5)
- add a second for loop that writes at different z positions

Of course, the test then needs two assertions (for the two blocks of data).

That way, this also tests the flushing of the buffer that you implemented :)

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
markbader: Yes I moved the code into the `reset_offset` method.

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
markbader: Sure, that is much better.

webknossos/tests/dataset/test_buffered_slice_utils.py
markbader: I adapted the test, to first write all the data except, the last 8 slices and after that add the last eight slices with a gap of 8 slices to the Layer. After that, I concatenate both blocks of data to compare them with the original data.

philippotto (approved):
Excellent :tada: Really great to have this feature now :)

PR #1054, opened 2024-04-24
Add layer_name to kwargs of from_images

### LINKED ISSUE #1053: Add layer_name to Dataset.from_images
## Detailed Description
Please add an optional `layer_name` argument to `Dataset.from_images`. When multiple layers are created, the layer name shall be the common prefix.

markbader:
### Description:
- This PR adds the argument `layer_name` to the `Dataset.from_images` method. If `from_images` generates a Dataset with one layer, it uses the given `layer_name` as name for this layer. In case multiple layers are created, the given layer name is a shared prefix of all layers.

### Issues:
- fixes #1053 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -17,6 +17,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Added
 
 ### Changed
+- Added `layer_name` as optional argument to `Dataset.from_images` method. If the created dataset contains only a single layer, `layer_name` is used, otherwise the given `layer_name` is a common prefix for all layers. [1054](https://github.com/scalableminds/webknossos-libs/pull/1054)
 
 ### Fixed
 
--- a/webknossos/tests/dataset/test_from_images.py
+++ b/webknossos/tests/dataset/test_from_images.py
@@ -25,13 +25,14 @@ def test_compare_tifffile(tmp_path: Path) -> None:
         tmp_path,
         (1, 1, 1),
         compress=True,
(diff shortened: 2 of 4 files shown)


normanrz (commented):
Can you comment how the layer names for multiple layers are constructed? I don't think I understand that yet. Would it make sense to strip the current prefix (ie. file name based) in that case?

markbader:
It depends, we currently use the Callable `map_filepath_to_layername` to generate the layer names. It can either be used by selecting one of the predefined methods defined in `ConversionLayerMapping` or by delivering a custom Callable to convert the file paths to layer names.
The default behavior is `INSPECT_SINGLE_FILE`, which distinguishes two different cases. If the first file that is found is 2D `ENFORCE_LAYER_PER_FOLDER` is used, otherwise `ENFORCE_LAYER_PER_FILE` will be applied.
`ENFORCE_LAYER_PER_FOLDER` uses the folder name, where the image file is stored in.
`ENFORCE_LAYER_PER_FILE` uses the filename.
I think as there are many options for functions that can convert the file path to a layer name, it would not be easy to strip a part of it, and still get unique layer names. Either we decide, that another behavior should be used when the `layer_name` is set, e.g., just numbering the layers after the common prefix, or we stay with the full filename or folder name based approach. But I would prefer the filename approach because it makes the layers easier to distinguish.

normanrz (approved):
Thanks for explaining.

PR #1067, opened 2024-05-07
Replace backslashes in windows upload

### LINKED ISSUE #1066: upload command on windows: backslashes on server, invalid dataset
A user created a valid dataset on a windows machine with the `webknossos convert` command, then called `webknossos upload` with a valid token. The upload went through, but the directory structure got lost: the files on the server had backslashes in the paths, like `'color\2-2-1\z0\y7\x1.wkw'`. Instead, when sending files to upload, the client should always replace the client’s path separator by `/`.

markbader:
### Description:
- When using `webknossos convert` and `webknossos upload` on a windows machine, the directory structure got lost due to backslashes in the `relative_path`. This should be fixed with this PR.

### Issues:
- fixes #1066 

### Todos:
 - [x] Updated Changelog


hotzenklotz (approved):
Changes look good to me. I have not tested this on a Windows machine though.

PR #1070, opened 2024-05-14
Add codespell config, fix typos

fm3:
### Description:
- Adds config for [codespell](https://github.com/codespell-project/codespell), fixes typos once.

Question: Do we want to add the tool to the dev-dependencies and/or assert no typos in the CI?


hotzenklotz:
> Question: Do we want to add the tool to the dev-dependencies and/or assert no typos in the CI?

I don't think we need to run this in the CI for every commit. There are some many checks already that often slow down PRs from getting merged that I don't feel like spelling adds enough value. 
Running this periodically, manually is fine with me.

PR #1082, opened 2024-05-28
Update max value of test_align_with_mag_against_numpy_implementation.

markbader:
### Description:
- The test `test_align_with_mag_against_numpy_implementation` fails occasionally when numbers are too big. This PR sets a new max value for the test to avoid these failures.

### Todos:
 - [x] Updated Changelog


webknossos/Changelog.md
```diff
@@ -19,6 +19,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 
 ### Changed
 - Moved functional parts of merge volume annotation CLI to Dataset and Annotation classes. [1055](https://github.com/scalableminds/webknossos-libs/pull/1055)
+- Set a new max value for test_align_with_mag_against_numpy_implementation to avoid failures due to high numbers. [1082](https://github.com/scalableminds/webknossos-libs/pull/1082)
```
philippotto, line 22: usually, we put a `#` in front of the issue number. however, i see that it slipped in like this quite a few times. could you fix the other occurrences, too?

```suggestion
- Set a new max value for test_align_with_mag_against_numpy_implementation to avoid failures due to high numbers. [#1082](https://github.com/scalableminds/webknossos-libs/pull/1082)
```

markbader: Nice that you noticed this. I added the `#` where necessary.

PR #1083, opened 2024-05-28
Fix cli download

### LINKED ISSUE #985: Webknossos CLI download annotation
## Context
When using the webknossos CLI to download an annotation, the download command is overloaded to take care of dataset downloads and annotation downloads. An annotation should be downloaded when no dataset is found. This is triggered by an AssertionError. Currently no assertion is thrown. Instead, an UnexpectedStatusError is raised.

## Expected Behavior
When a URL to an Annotation is given, the Annotation should be downloaded.

markbader:
### Description:
- When using the webknossos CLI to download an annotation, the download command is overloaded to take care of dataset downloads and annotation downloads. This PR uses the client to resolve a possible short link and compares the url to a dataset url regex and an annotation url regex to distinguish both cases.

### Issues:
- fixes #985 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -21,6 +21,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Moved functional parts of merge volume annotation CLI to Dataset and Annotation classes. [1055](https://github.com/scalableminds/webknossos-libs/pull/1055)
 
 ### Fixed
+- We've resolved an issue that was affecting the download of annotations through the Command Line Interface (CLI). Additionally, we've improved the CLI to better differentiate between dataset URLs and annotation URLs before initiating different download methods. [1083](https://github.com/scalableminds/webknossos-libs/pull/1083)
 
 
 ## [0.14.22](https://github.com/scalableminds/webknossos-libs/releases/tag/v0.14.22) - 2024-05-13
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -21,7 +21,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Moved functional parts of merge volume annotation CLI to Dataset and Annotation classes. [1055](https://github.com/scalableminds/webknossos-libs/pull/1055)
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


webknossos/Changelog.md
```diff
@@ -21,6 +21,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Moved functional parts of merge volume annotation CLI to Dataset and Annotation classes. [1055](https://github.com/scalableminds/webknossos-libs/pull/1055)
 
 ### Fixed
+- We've resolved an issue that was affecting the download of annotations through the Command Line Interface (CLI). Additionally, we've improved the CLI to better differentiate between dataset URLs and annotation URLs before initiating different download methods. [1083](https://github.com/scalableminds/webknossos-libs/pull/1083)
```
fm3, line 24:
```suggestion
- Fixed an issue with downloading annotations through the Command Line Interface. [#1083](https://github.com/scalableminds/webknossos-libs/pull/1083)
```

(I’d go for a more compact style in the changelog, and only mention the user-facing change, which is the fix, not the refactoring)

webknossos/webknossos/cli/download.py
```diff
@@ -72,15 +74,28 @@ def main(
...
+                    layers=layers,
+                    mags=mags,
+                )
+            except RuntimeError as err:
+                print(f"Dataset could not be downloaded: {err}")
+        elif re.match(_ANNOTATION_URL_REGEX, url):
+            try:
+                Annotation.download(annotation_id_or_url=url).save(target)
+            except AssertionError as err:
+                print(f"Annotation could not be downloaded: {err}")
```
fm3, line 95: Do the other cli commands also implement this catch-print pattern? Is there a specific reason we don’t want to show the full stacktrace?

Also, why specifically catch AssertionError here, and RuntimeError above, but not all Exceptions or Errors?

I think I’d vote to bubble up the exceptions, that way the exit-code will also be 1 if the command was not successful. Consequently, the elese-branch could also raise instead of print. What do you think?

fm3 (changes requested):

webknossos/webknossos/cli/download.py
markbader: I agree. I removed the try/catch and raised an error in the else case. It is useful when the exit code is 1 in case the command fails. Whether we show the stack trace, or not, depends somehow on the users. For debugging purposes, it is easier to have a stack trace, but it is also a highly technical error message that might confuse some users.

fm3 (approved):

PR #1095, opened 2024-06-05
Fix cube jobs

### LINKED ISSUE #1092: Downsampling and upsamping fails for nd zarr3 datasets
## Context
- Calling `.downsample()` on a zarr3 dataset leads to an issue, as the method `downsample_cube` cannot handle more than 3 dimensions.

## Expected Behavior
- Downsampling and upsampling should work for nd datasets just like for 3d datasets.

markbader:
### Description:
- When executing cube jobs like downsample, upsample or compress, the jobs fail when the views have more than 3 dimensions. This PR enables the jobs to handle more than 3 dimensions.

### Issues:
- fixes #1092

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/examples/convert_4d_tiff.py
```diff
@@ -8,9 +8,12 @@ def main() -> None:
     dataset = wk.Dataset.from_images(
         Path(__file__).parent.parent / "testdata" / "4D" / "4D_series",
         "testoutput/4D_series",
-        voxel_size=(10, 10, 10),
+        layer_category="color",
```
normanrz, line 11:
```suggestion
        layer_category=wk.COLOR_CATEGORY,
```

webknossos/tests/dataset/test_downsampling.py
```diff
@@ -454,3 +454,32 @@ def test_downsample_2d(tmp_path: Path) -> None:
...
+        "color",
+        COLOR_CATEGORY,
+        bounding_box=source_layer.bounding_box,
+        data_format="zarr3",
+        dtype_per_channel="int8",
+    )
+
+    source_mag = source_layer.get_mag("1")
+
+    with pytest.warns(UserWarning):
```
normanrz, line 478: What warning is this?

webknossos/webknossos/dataset/view.py
```diff
@@ -571,6 +612,8 @@ def get_view(
         The default is to return the same view as the current bounding box,
         in case of a `MagView` that's the layer bounding box.
         One can supply one of the following keyword argument combinations:
+        * `relative_bbox`in Mag(1)
+        * `absolute_bbox` in Mag(1)
```
normanrz, line 616: These are called `relative_bounding_box` and `absolute_bounding_box` elsewhere.

normanrz (approved):

webknossos/tests/dataset/test_downsampling.py
markbader: In the method `add_copy_mag` a warning is generated that the values changed in a way they are not comparable anymore:
`UserWarning: [WARNING] Properties changed in a way that they are not comparable anymore. Most likely the bounding box naming or axis order changed.`
When exporting the dataset properties to a JSON file, the `_last_read_properties` only have xyz axes for an empty initialized layer, but after copying the mag it has tzyx axes.

webknossos/webknossos/dataset/view.py
markbader: I changed the naming accordingly.

PR #1100, opened 2024-06-06
Suffixes with capital letters

### LINKED ISSUE #1099: Suffixes with capital letters should be valid too
## Context
- Some image files have suffixes written in capital letters. As they are case-sensitive compared against a list of allowed suffixes, conversion sometimes fails.

## Expected Behavior
- e.g. `.DCM` should be as valid as `.dcm`

markbader:
### Description:
- This PR makes the lookup for supported PIMS suffixes case-insensitive.

### Issues:
- fixes #1099

### Todos:
 - [x] Updated Changelog


hotzenklotz (approved):
LGTM

PR #1130, opened 2024-07-02
Add new add_nx_graphs method to skeleton.py

### LINKED ISSUE #475: Add from_nx_graphs method to Skeleton
We should be able to construct a Skeleton from networkx graphs or a dict of them for groups. This should also be able to globalize the node-ids and take additional parameters, just as [`generate_nml`](https://github.com/scalableminds/wknml/blob/master/wknml/nml_generation.py#L99-L104) in the old `wknml` package.

MatthisCl:
### Description:
- To replace wknml in voxelytics we need support to add nx.Graphs to the Skeleton object.

### Issues:
- fixes #475 

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests


MatthisCl:
Original Issue:
"We should be able to construct a Skeleton from networkx graphs or a dict of them for groups. This should also be able to globalize the node-ids and take additional parameters, just as [generate_nml](https://github.com/scalableminds/wknml/blob/master/wknml/nml_generation.py#L99-L104) in the old wknml package"

### Some notes

Why not as the suggested static method?:
We would always have to generate a dummy skeleton inside the function.

Why not take additional parameters?:
Most of the parameters where moved to the Annotation object (eg. taskBoundingBox etc.). If needed we will create a dummy Annotation first and then call add_nx_graphs on its skeleton.

Globalizing Node id's:
In the old wknml package one could pass as parameter if nodes should be globalized (eg. unique id's). Node_id's are already unique in wk when adding via add_group/add_tree/add_node. Not sure if i am wrong but i don't think we need non-globalized id's, also that would requiere to manually overwrite certain id's which doesn't look like a good idea^^

webknossos/webknossos/skeleton/skeleton.py
```diff
@@ -104,6 +105,43 @@ def save(self, out_path: Union[str, PathLike]) -> None:
...
+    ) -> None:
+        """
+        A utility to add nx graphs [NetworkX graph object](https://networkx.org/) to a wk skeleton object. Accepts both a simple list of multiple skeletons/trees or a dictionary grouping skeleton inputs.
+
+        Arguments:
+        tree_dict (Union[List[nx.Graph], Dict[str, List[nx.Graph]]]): A list of wK tree-like structures as NetworkX graphs or a dictionary of group names and same lists of NetworkX tree objects.
+        """
+
+        if not isinstance(tree_dict, dict):
+            tree_dict = {"main_group": tree_dict}
```
philippotto, line 119: What was the behavior of the old `generate_nml`? Did it name the root group "main_group", too?

webknossos/webknossos/skeleton/skeleton.py
```diff
@@ -104,6 +105,43 @@ def save(self, out_path: Union[str, PathLike]) -> None:
...
+
+        for group_name, trees in tree_dict.items():
+            group = self.add_group(group_name)
+            for tree in trees:
+                tree_group = group.add_tree("tree")
```
philippotto, line 124: all trees will be named "tree"? maybe `generate_nml` can give inspiration here, too.
also, I wouldnt call the variable `tree_group`. maybe `wk_tree`.

philippotto (changes requested):
Cool! Could you add a unit test for this new feature? I would also be interested in how this compares to the old generate_nml behavior. Ideally, there would be a unit test that uses the old and new code and compares some properties (maybe even total equality). Once we got rid of the old generate_nml code, we can remove it together with that unit test.

Regarding `globalize_tree_ids`: I think, providing False for that parameter mostly means "I'm certain that the ids are already unique and I want to keep my ids". However, I'm not sure whether that behavior was used at all. Could you doublecheck the usages of the old generate_nml?

MatthisCl:
> Cool! Could you add a unit test for this new feature? I would also be interested in how this compares to the old generate_nml behavior. Ideally, there would be a unit test that uses the old and new code and compares some properties (maybe even total equality). Once we got rid of the old generate_nml code, we can remove it together with that unit test.
> 
> Regarding `globalize_tree_ids`: I think, providing False for that parameter mostly means "I'm certain that the ids are already unique and I want to keep my ids". However, I'm not sure whether that behavior was used at all. Could you doublecheck the usages of the old generate_nml?

Just checked in vx, `globalize_id's` is never used when calling `generate_nml` and the default is true, so we didn't use the behaviour.

Regarding the unit test i would add a test in the `test_skeleton.py` to check if adding `nx.Graphs` works correctly.

We don't use `generate_nml` in the wk-libs, so i think the unit test for comparing it with the new `add_nx_graphs` should be written in vx later? Also doesn't compare directly, as the new workflow will be create trees -> add to skeleton -> skeleton.save rather than generating the nml from the trees directly. (See https://github.com/scalableminds/voxelytics/issues/2762). Not sure if the methods that exist right now cover all usecases though

philippotto:
> Just checked in vx, `globalize_id's` is never used when calling `generate_nml` and the default is true, so we didn't use the behaviour.

Great :+1:

> Regarding the unit test i would add a test in the `test_skeleton.py` to check if adding `nx.Graphs` works correctly.

Sounds good!

> We don't use `generate_nml` in the wk-libs, so i think the unit test for comparing it with the new `add_nx_graphs` should be written in vx later? Also doesn't compare directly, as the new workflow will be create trees -> add to skeleton -> skeleton.save rather than generating the nml from the trees directly. (See [scalableminds/voxelytics#2762](https://github.com/scalableminds/voxelytics/issues/2762)). Not sure if the methods that exist right now cover all usecases though

I'd prefer to have the test in this PR. I imagine it like this: generate some trees in some group hierarchy and then use the old wknml code and the new wklibs code to produce an NML each. then assert that the two NMLs are equal (maybe not on a byte-level, but you could parse both NMLs with the wklibs code and compare from there).
the motivation for that test would be that the old approach and the new one both produce the same data.

webknossos/webknossos/skeleton/skeleton.py
MatthisCl: Yes

webknossos/webknossos/skeleton/skeleton.py
MatthisCl: I changed tree_group to wk_tree, also names will now be tree_0 tree_1 ... for each group

webknossos/webknossos/skeleton/skeleton.py
MatthisCl: Yes

MatthisCl:
> I'd prefer to have the test in this PR. I imagine it like this: generate some trees in some group hierarchy and then use the old wknml code and the new wklibs code to produce an NML each. then assert that the two NMLs are equal (maybe not on a byte-level, but you could parse both NMLs with the wklibs code and compare from there). the motivation for that test would be that the old approach and the new one both produce the same data.

Could you review the two unit tests and the change to the tree naming :) Would merge after that

philippotto (approved):
thanks for adding the tests :+1: lgtm!

PR #1136, opened 2024-07-15
Unit support

### LINKED ISSUE #1076: Support other units than nm for dataset voxel size
WEBKNOSSOS now supports unit-aware voxel sizes as of https://github.com/scalableminds/webknossos/pull/7783

- The libs should also understand datasource jsons in this format, but the old ones too (default to nm)
- It should be possible to specify the unit during dataset creation
- The libs API client should use HTTP API version 7

markbader:
### Description:
- Adds a `voxel_unit` property to Datasets and adds support for the new format of scale parameter in `datasource-properties.json`

### Issues:
- fixes #1076 

### Todos:
 - [ ] Updated Changelog
 - [ ] Added / Updated Tests


webknossos/webknossos/dataset/dataset.py
```diff
@@ -117,6 +118,61 @@
 # This regex matches any character that is not allowed in a layer name.
 _UNALLOWED_LAYER_NAME_CHARS = re.compile(r"[^A-Za-z0-9_$@\-\.]")
 
+_VOXEL_UNITS = {
```
fm3, line 121: maybe move to a different file? I feel like dataset.py is already a bit too big :sweat_smile:

fm3 (commented):
Nice, this can read both old and new jsons :)

I think it would be helpful to use similar models and names here as in the webknossos frontend and backend code. In wk we introduced a type VoxelSize, which is basically `ApiScaleWithUnit` here now. I would have expected the python dataset class to also feature just one field that now has such a type (with both factor and unit), rather than now two fields voxel_size and voxel_unit. That way, unit can even be non-optional if a voxel size is present (and default to nanometer when reading old-format data). Of course, this breaks python api compatibility, so it’s not sure if the downsides are stronger. cc @philippotto @normanrz 

Also, could the unit be an enum in python? In wk, we have called the type LengthUnit.

I would also advocate to always use the term voxel_size rather than scale, except where absolutely needed (i.e. the json for backwards compatibility). For example, I would suggest renaming `ApiScaleWithUnit` to `ApiVoxelSize` even though the field in the ApiDataSource is still named `scale`. What do you think?

normanrz:
We shouldn't break the API with this PR. Adding a new unit-aware voxel size property is fine. 

`scale` should be deprecated (with a warning) but kept for compatibility.

webknossos/testdata/tiff/datasource-properties.wkw-fixture.json
```diff
@@ -11,18 +11,33 @@
...
-}
+  "scale": {
+    "unit": "nanometer",
+    "factor": [
+      11.24,
+      11.24,
+      25.0
+    ]
+  }
+}
```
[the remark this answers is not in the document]
markbader: I added the newlines, they were somehow removed from my format on save command.

webknossos/webknossos/dataset/dataset.py
markbader: I made a unit.py file within the dataset folder, I am not sure if it is a really fitting place, but it is similar to data_format.py

webknossos/webknossos/dataset/dataset.py
```diff
@@ -341,17 +349,17 @@ def __init__(
             self._layers[layer_properties.name] = layer
 
         if dataset_existed_already:
-            if voxel_size is None:
+            if voxel_size_with_unit is None:
                 warnings.warn(
-                    "[DEPRECATION] Please always supply the voxel_size when using the constructor Dataset(your_path, voxel_size=your_voxel_size)."
+                    "[DEPRECATION] Please always supply the voxel_size or voxel_size_with_unit when using the constructor Dataset(your_path, voxel_size=your_voxel_size)."
```
fm3, line 354: Further above it reads as though we want to allow both voxel_size and voxel_size_with_unit for the moment. Then I guess this warning should also happen only if both are missing (meaning the old keyword scale was used)?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -717,12 +739,18 @@ def layers(self) -> Dict[str, Layer]:
 
     @property
     def voxel_size(self) -> Tuple[float, float, float]:
-        return self._properties.scale
+        conversion_factor = get_conversion_factor(self._properties.scale.unit)
```
fm3, line 742: self._properties.scale is a VoxelSize, right? Could VoxelSize have a to_nanometer method, returning Tuple[float, float, float]? Then the get_conversion_factor method could be removed (or made private).

webknossos/webknossos/dataset/length_unit.py
```diff
@@ -0,0 +1,125 @@
...
+    "foot": LengthUnit.FOOT,
+    "yd": LengthUnit.YARD,
+    "yard": LengthUnit.YARD,
+    "mi": LengthUnit.MILE,
+    "mile": LengthUnit.MILE,
+    "pc": LengthUnit.PARSEC,
+    "parsec": LengthUnit.PARSEC,
+}
+
+_LENGTH_UNIT_TO_CONVERSION_FACTOR = {
```
fm3, line 88:
```suggestion
_LENGTH_UNIT_TO_NANOMETER = {
```

webknossos/webknossos/dataset/length_unit.py
```diff
@@ -0,0 +1,125 @@
...
+    LengthUnit.ANGSTROM: 0.1,
+    LengthUnit.INCH: 25400000.0,
+    LengthUnit.FOOT: 304800000.0,
+    LengthUnit.YARD: 914400000.0,
+    LengthUnit.MILE: 1609344000000.0,
+    LengthUnit.PARSEC: 3.085677581e25,
+}
+
+
+def get_unit_from_str(unit: str) -> LengthUnit:
```
fm3, line 118:
```suggestion
def length_unit_from_str(unit: str) -> LengthUnit:
```

markbader:
Thanks for your feedback!

As discussed in Slack, I introduced a new field within the dataset class that features both, factor and unit. It is called `voxel_size_with_unit` now. The old `voxel_size` is still available and stores the value in nanometers. 

I also put the units in an Enum class and called it `LengthUnit`, `ApiScaleWithUnit` was renamed to `ApiVoxelSize`, and when downloading or opening a Dataset the short and long versions of the LenthUnit names are supported. When saving a dataset, the long unit name is used.

webknossos/webknossos/dataset/length_unit.py
```diff
@@ -0,0 +1,125 @@
...
+    if unit in _STR_TO_UNIT_MAP:
+        return _STR_TO_UNIT_MAP[unit]
+    raise ValueError(f"Unknown unit: {unit}")
+
+
+def get_conversion_factor(unit: LengthUnit) -> float:
```
fm3, line 124:
```suggestion
def to_nanometer(unit: LengthUnit) -> float:
```

fm3 (commented):
Awesome, this is looking much better :) I added a couple more comments, mostly on naming. Next, I’ll do some testing.

The CLI does not yet support supplying a unit, right? If not, we should create a follow-up issue. (Or implement it directly, if that it is simple)

fm3 (commented):
Testing went smoothly :)

webknossos/webknossos/dataset/dataset.py
markbader: That is true, but I think it should already be the case. In line 261-268 the old arguments `scale` and `voxel_size` are converted to `voxel_size_with_unit`. When the `voxel_size_with_unit` variable is still not set, `scale` and `voxel_size` are unset too.
I am not sure about the purpose of the warning, but I suppose that it was possible to open an existing dataset with the Dataset constructor without specifying a voxel size in an old version of the libs.

fm3: I see, thanks! Let’s keep it like this for now then. Maybe we can do a round of breaking changes (of stuff that has been deprecated for a while) some time soon.

markbader:
I resolved the naming issues and added a unit field to the converters of the CLI. 
Another open task is the integration of the unit field in the nml. I think this might be better in a follow-up PR, what do you think?

webknossos/webknossos/dataset/length_unit.py
markbader: Done.

webknossos/webknossos/dataset/length_unit.py
markbader: Done.

webknossos/webknossos/dataset/dataset.py
markbader: Yes, I added the method to VoxelSize, looks much better :slightly_smiling_face:

webknossos/webknossos/cli/convert.py
```diff
@@ -37,14 +40,21 @@ def main(
...
         typer.Option(
             help="The size of one voxel in source data in nanometers. "
             "Should be a comma separated string (e.g. 11.0,11.0,20.0).",
             parser=parse_voxel_size,
-            metavar="VOXEL_SIZE",
+            metavar="VoxelSize",
+            show_default=False,
         ),
     ],
+    length_unit: Annotated[
```
fm3, line 52: this is the name of the cli argument, right? Maybe just unit or voxel_size_unit would be better here. While I think LenghtUnit is the right name for the type, the variable here can have more semantics in context

webknossos/webknossos/cli/convert_knossos.py
```diff
@@ -246,15 +251,21 @@ def main(
...
+            metavar="VoxelSize",
             show_default=False,
         ),
     ],
+    length_unit: Annotated[
+        LengthUnit,
+        typer.Option(
+            help="The unit of the voxel size.",
+        ),
+    ] = "nanometer",  # type:ignore
```
fm3, line 268: that’s a lot of "nanometer" strings in all these files. Maybe a DEFAULT_UNIT (and maybe additionally DEFAULT_UNIT_STR) or something like that could be defined somewhere, next to the VoxelSize class, and be imported here?

webknossos/webknossos/dataset/dataset.py
fm3: When users want to create a new dataset and supply a voxel size with length unit (not via cli), they will need to import those classes, right? How are they exposed? Do they need to be forwarded in any init files?

webknossos/webknossos/cli/convert.py
markbader: Sounds good, then I would prefer unit, as it is shorter.

webknossos/webknossos/cli/convert_knossos.py
markbader: Done.

fm3 (approved):
:tada:

PR #1137, opened 2024-07-15
Remove trailing slashes from webknossos_url

### LINKED ISSUE #910: Ignore trailing slashes in URL comparisons
When opening a remote dataset, the supplied `webknossos_url` is compared to the active webknossos context's url. If only one of the two has a trailing slash, the comparison fails and the context's token is reset to `null`. This leads to permission errors when opening the remote dataset, because a token is no longer available.
Instead the url comparison should succeed, regardless of any trailing slashes.

Code context: https://github.com/scalableminds/webknossos-libs/blob/ab4e538b45f41108429442c68477f7c84c23e8d9/webknossos/webknossos/dataset/dataset.py#L410

Python's `rstrip` method should be helpful for fixing this.

It might also make sense to scan the code for further occurrences of this issue.

markbader:
### Description:
- When comparing webknossos context url (which is always without trailing slashes) and `webknossos_url` parameter, possible trailing slashes in the latter case are removed.

### Issues:
- fixes #910

### Todos:
 - [x] Updated Changelog


hotzenklotz (approved):
LGTM

PR #1150, opened 2024-07-29
Add chunk and shard shape to CLI convert command options.

### LINKED ISSUE #1145: Add chunk-shape and shard-shape to convert CLI
Add chunk-shape and shard-shape to convert CLI

markbader:
### Description:
- Add chunk shape and shard shape to the CLI convert command.

### Issues:
- fixes #1145 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/geometry/vec3_int.py
+++ b/webknossos/webknossos/geometry/vec3_int.py
@@ -112,6 +112,8 @@ class Vec3Int(VecInt):
     def from_str(string: str) -> "Vec3Int":
         if re.match(r"\(\d+,\d+,\d+\)", string):
             return Vec3Int(tuple(map(int, re.findall(r"\d+", string))))
+        elif re.match(r"^\d+,\d+,\d+$", string):
+            return Vec3Int(tuple(map(int, string.split(","))))
 
         return Vec3Int.full(int(string))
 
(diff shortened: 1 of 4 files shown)


webknossos/webknossos/geometry/vec3_int.py
```diff
@@ -112,6 +112,8 @@ def from_vec_or_int(vec_or_int: Union["Vec3IntLike", int]) -> "Vec3Int":
     def from_str(string: str) -> "Vec3Int":
         if re.match(r"\(\d+,\d+,\d+\)", string):
             return Vec3Int(tuple(map(int, re.findall(r"\d+", string))))
+        elif re.match(r"^\d+,\d+,\d+$", string):
```
hotzenklotz, line 115: Vielleicht ein kurzer Kommentar, was die Regex macht. Speziell im Vergleich zu bestehenden regex.

hotzenklotz (approved):
🚀

webknossos/webknossos/geometry/vec3_int.py
markbader: Sure, I added a comment for both regex and fixed an issue with the first regex. The first regex evaluated to true even if the correct pattern, e.g., `(10,20,30)` is only a substring of `string`. So the map could deliver more than three integers. This should be fixed now.

PR #1152, opened 2024-07-30
Adapt from images to convert single file

### LINKED ISSUE #1149: Convert single file with CLI
Using `webknossos convert` requires a folder as source. It would be nice, if I could use a single image (e.g. an OME-TIFF file) as input.

markbader:
### Description:
- This PR adapts the `from_images` method of the `Dataset` class to convert a single image file into a layer if the `input_path` is a file rather than a directory.

### Issues:
- fixes #1149

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -598,12 +598,12 @@ class Dataset:
         executor: Optional[Executor] = None,
     ) -> "Dataset":
         """
-        This method imports image data in a folder as a WEBKNOSSOS dataset. The
-        image data can be 3D images (such as multipage tiffs) or stacks of 2D
-        images. In case of multiple 3D images or image stacks, those are mapped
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/dataset/dataset.py
```diff
@@ -630,11 +632,17 @@ def from_images(
         if use_bioformats is not False:
             valid_suffixes.update(pims_images.get_valid_bioformats_suffixes())
 
-        input_files = [
-            i.relative_to(input_upath)
-            for i in input_upath.glob("**/*")
-            if i.is_file() and i.suffix.lstrip(".").lower() in valid_suffixes
-        ]
+        if input_upath.is_dir():
```
normanrz, line 635: On non-local file systems, is_file is significantly cheaper than is-dir. Can you flip the if/else block?

normanrz (approved):

PR #1155, opened 2024-08-01
Require organisation id in Annotation constructor.

### LINKED ISSUE #1015: Upload of wk.Annotation should require organization_id
When creating a new annotation via `a = wk.Annotation(name, dataset_name=dataset_name, voxel_size=voxel_size)` it should be required to pass an organization name for the specified dataset to avoid ambiguity.

Otherwise during upload to wk, the lookup can be ambiguous and may fail if other organizations have datasets of that name.

Related: https://github.com/scalableminds/webknossos/issues/7681

markbader:
### Description:
- When calling the `Annotation()` constructor, an `organisation_id` is mandatory now. The `organization_id` might be provided directly within the `Annotation()` call, or it is inferred from current webknossos context.

### Issues:
- fixes #1015

### Todos:
 - [ ] Updated Changelog

### DIFF
--- a/webknossos/webknossos/annotation/annotation.py
+++ b/webknossos/webknossos/annotation/annotation.py
@@ -51,6 +51,7 @@ from upath import UPath
 from zipp import Path as ZipPath
 
 import webknossos._nml as wknml
+from webknossos.client.context import _get_context
 
 from ..dataset import (
     SEGMENTATION_CATEGORY,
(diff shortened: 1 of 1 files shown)


normanrz:
Not sure, I am happy with this. Providing an organization id might be cumbersome for users. And, I don't see why it wouldn't be possible to infer that upon upload.
Fwiw, we could even make the dataset name optional. It could be supplied upon upload.
cc @fm3

normanrz (approved):

PR #1162, opened 2024-08-06
Downgrade Numpy

Tobias314:
### Description:
- Downgrade numpy to <2.0 because of compatibility issues in packages using wk-libs

### Todos:
 - [ ] Updated Changelog


webknossos/pyproject.toml
```diff
@@ -40,7 +40,7 @@ loxun = "^2.0"
 natsort = "^8.4.0"
 networkx = "^3"
 numcodecs = "^0.12"
-numpy = "^2"
+numpy = "^1.26"
```
normanrz, line 43: What about?
```suggestion
numpy = ">=1.26"
```

markbader:
We decided to update the dependencies of voxelytics, so we can keep numpy v2 within the libs.

PR #1163, opened 2024-08-06
Enable metadata access for remote datasets.

### LINKED ISSUE #1148: Add dataset/folder metadata
## Detailed Description

Webknossos introduces metadata for datasets and folders in https://github.com/scalableminds/webknossos/pull/7886. The libs should be able to view and edit this metadata.

markbader:
### Description:
- Implements getters and setters for the new metadata field in remote datasets and folders.

### Issues:
- fixes #1148

### Todos:
 - [x] Updated Changelog

 - [x] blocked by https://github.com/scalableminds/webknossos/pull/7886
 
 ### How to test:
 Just run the example in `accessing_metadata.py` and enter your webknossos token when asked. After execution, the dataset `l4_sample` and the folder `Datasets/Demo-Datasets` should have an entry with metadata.


webknossos/webknossos/client/api_client/models.py
```diff
@@ -49,13 +49,21 @@ class ApiBoundingBox:
     depth: int
 
 
+@attr.s(auto_attribs=True)
+class ApiAdditionalAxes:
```
fm3, line 53:
```suggestion
class ApiAdditionalAxis:
```
(I think this is singular)

webknossos/examples/accessing_metadata.py
```diff
@@ -0,0 +1,32 @@
+import webknossos as wk
+from webknossos.client.api_client.models import ApiMetadata
```
fm3, line 2: The `api_client.models` classes should stay private to our own code, to enable future changes to the HTTP API without changing the user-exposed python API.

Please create an abstraction layer between ApiMetadata and the user-exposed metadata in Dataset and Folder classes. Maybe even without a `type` property, but using python types instead? The type property could be inferred in the conversion. 
Maybe it could even be just a dict? Not sure about that, maybe @philippotto or @MichaelBuessemeyer have a preference.

fm3 (changes requested):
Nice that this is added so quickly :)

I added a couple of comments

webknossos/examples/accessing_metadata.py
philippotto: I think, a dict mapping from key to primitive would be cool, but I'm not sure it's worth the complexity. in the end, this api should be easy and clear to use. directly using primitives might make people believe that they can also put python objects there. so, it would definitely need thorough validation.
I don't have a strong opinion about this, though.

philippotto: > because it seems that the key is not unique so far

to which implementation are you referring? the wklibs one? because in wk, it should be unique (if not, we should fix it :sweat_smile:)

markbader: The wk one, it seems that the creation of duplicates is restricted by the frontend but not in the backend. I can send the same key-value pair multiple times with the API without an error. I just tested it with the `l4_test_mark` dataset.
Screenshot 2024-09-05 at 09-58-33 WEBKNOSSOS

@MichaelBuessemeyer

MichaelBuessemeyer: Sorry for entering the thread so late
>because in wk, it should be unique (if not, we should fix it 😅)

The frontend does the validation that no key is duplicated. The backend just takes what the frontend sends and saves & serves it. 

Do you think a validation would be helpful? I can certainly build that :)

markbader: > Do you think a validation would be helpful?

I changed the wk-libs implementation to use a dict now. So when using the libs as intended, the key should be unique too. Nevertheless, a validation would be helpful to ensure that the API is used correct.

fm3 (commented):
Looking good :)
I tested and the access in the style of the example worked for me. However, intuitively, I first tried `dataset.metadata["test"] = "asdf"`, which did not error, but also did have any effect. Do you think it would be simple to support this case as well? (Or alternatively somehow freeze this object, so that an error is provided, suggesting the other access style?) If this is very hard to achieve, then this current version is also ok.

webknossos/examples/accessing_metadata.py
fm3: > Do you think a validation would be helpful? I can certainly build that :)

We should either fully support duplicates, or ensure that none arise. So with the current design, I’d say yes, the backend should assert this in the update routes. Could you create an issue for that? It’s not super urgent, though.

MichaelBuessemeyer: I created an issue for this: https://github.com/scalableminds/webknossos/issues/8068

And as I wanted a rather quick and simple task right now, I simply started the implementation :see_no_evil:

markbader:
> I first tried `dataset.metadata["test"] = "asdf"`, which did not error, but also did have any effect. Do you think it would be simple to support this case as well? 

As it was currently just a dict neither freezing nor overwriting `__setitem__` is super easy without unwanted side effects. I implemented a subclass of dict now, that fetches data on each access and updates the data when the dict is updated. This might be more intuitive, but it is also slower. I guess we don't want to update the metadata very often, so this might be okay. What is your opinion on this approach?

fm3:
Interesting idea! I like this access pattern from the user perspective. However, I don’t know enough about python and its conventions to fully know what unwanted side-effects this specialized dict may have :thinking: Also I’m not certain I can review its implementation. Maybe we can get a second opinion? Maybe @daniel-wer could also have a look?

webknossos/webknossos/dataset/_metadata.py
```diff
@@ -0,0 +1,104 @@
...
+    Union,
+)
+
+from webknossos.client.api_client.models import ApiDataset, ApiFolder, ApiMetadata
+from webknossos.utils import infer_metadata_type, parse_metadata_value
+
+_T = TypeVar("_T", bound="Metadata")
+
+
+class Metadata(dict):
```
normanrz, line 19: Have you considered implementing a `MutableMapping` directly? I think that would remove some trouble with the persistence that a `dict` does automatically.

```suggestion
class Metadata(MutableMapping):
```

fm3 (commented):
I tested again and it works for me now :)
@normanrz could you do a final review round please?

normanrz (approved):
Fine from my side!

fm3 (approved):

PR #1170, opened 2024-08-13
Announce dataset creation with API

### LINKED ISSUE #1138: Announce datasets with permissions
See https://github.com/scalableminds/webknossos/issues/7457

Add a function to the wklibs to announce a dataset via the wk API before writing that dataset to the filesystem.

markbader:
### Description:
- Announce a dataset via the wk API before writing it to the file system. Team managers can create an entry in the database with the dataset name and assign a team, non-privileged webknossos users can then access their datasets.

### Issues:
- fixes #1138

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/client/api_client/wk_api_client.py
+++ b/webknossos/webknossos/client/api_client/wk_api_client.py
@@ -207,3 +207,14 @@ class WkApiClient(AbstractApiClient):
         return self.post_multipart_with_json_response(
             route, ApiTaskCreationResult, multipart_data=data, files=files
         )
+
+    def trigger_reload(
+        self,
+        organization_name: str,
+        dataset_name: str,
+        sharing_token: Optional[str] = None,
+    ):
+        route = (
+            f"/triggers/reload/{organization_name}/{dataset_name}?token={sharing_token}"
+        )
+        self._post(route)


webknossos/webknossos/client/api_client/datastore_api_client.py
```diff
@@ -47,7 +50,7 @@ def dataset_reserve_upload(
         reserve_upload_information: ApiReserveDatasetUploadInformation,
         token: Optional[str],
         retry_count: int,
-    ) -> None:
+    ) -> httpx.Response:
```
fm3, line 53: I think I’d prefer not to return the response here. The callers aren’t supposed to do anything with it. In case of an error response, an exception is already thrown. So unless you have a special reason, please change back the return type and instead remove the `return` statement below. This should also make the httpx import unnecessary

webknossos/webknossos/client/api_client/datastore_api_client.py
```diff
@@ -56,6 +59,25 @@ def dataset_reserve_upload(
...
 
+    def dataset_trigger_reload(
+        self,
+        organization_name: str,
+        dataset_name: str,
+        token: Optional[str] = None,
+    ) -> httpx.Response:
+        route = f"/triggers/reload/{organization_name}/{dataset_name}"
+        query: Query = {"token": token}
+        return self._post(route, query=query)
```
fm3, line 70: same as above, I don’t think the return is needed

webknossos/webknossos/client/api_client/datastore_api_client.py
```diff
@@ -56,6 +59,25 @@ def dataset_reserve_upload(
...
+
+    def dataset_reserve_manual_upload(
+        self,
+        dataset_announce: ApiDatasetAnnounceUpload,
+        token: Optional[str],
+    ) -> httpx.Response:
+        route = "/datasets/reserveManualUpload"
+        query: Query = {"token": token}
+        return self._post_json(route, dataset_announce, query)
```
fm3, line 79: same as above, I don’t think the return is needed

webknossos/webknossos/dataset/dataset.py
```diff
@@ -396,6 +396,56 @@ def open(cls, dataset_path: Union[str, PathLike]) -> "Dataset":
...
+            initial_team_ids=initial_team_ids,
+            folder_id=folder_id,
+        )
+        token = token or context.token
+        upload_url = _cached_get_upload_datastore(context)
+        datastore_api = context.get_datastore_api_client(upload_url)
+        datastore_api.dataset_reserve_manual_upload(dataset_announce, token=token)
+
+    @classmethod
+    def trigger_reload(
```
fm3, line 430: Do all datasets have this method? I’m not sure how to express that this is only for datasets present in a webknossos datastore :thinking: If you only create the dataset locally, this does not make sense. Maybe you have a good idea

fm3 (commented):
Looking good already :) I added a few comments

webknossos/webknossos/client/api_client/datastore_api_client.py
markbader: Alright, I removed the `return` statements. Previously, some methods like `dataset_finish_upload` returned the result of `_post_json` which always was `None`. I thought it would make sense to either make `_post_json` return something or remove the return statement from `dataset_finish_upload` and `dataset_reserve_upload`.

webknossos/webknossos/dataset/dataset.py
markbader: I don't have a good idea either. We could rename the method to `trigger_reload_in_datastore`, make the docstring more specific or put it somewhere else in the libs, like a new datastore class or a collection of wk-api utils, do you have any preferences?

fm3: I guess with renaming + better docstring we are already on a good way :)

fm3 (approved):

PR #1182, opened 2024-09-02
Fix nightly tests

hotzenklotz:
### Description:
- Replaced `docker-compose` command with `docker compose` in test setup. Hopefully that fixes the nightly tests.

### Issues:
- fixes https://github.com/scalableminds/webknossos-libs/actions/runs/10661673097/job/29547543258#step:6:16

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests


markbader (commented):
LGTM :)

Within the CI workflow `ci.yml` the `register_cluster.sh` script is called. This script also uses the docker-compose command that might be replaced as well.

hotzenklotz:
> Within the CI workflow `ci.yml` the `register_cluster.sh` script is called. This script also uses the docker-compose command that might be replaced as well.

Good catch. I updated that as well.

markbader (approved):

PR #1186, opened 2024-09-04
Nd bounding box intersected_with doc

xfolka:
### Description:
- Clarified the documentation for intersected_with method in nd_bounding_box

### Issues:
- fixes #...

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation

### DIFF
--- a/webknossos/webknossos/geometry/nd_bounding_box.py
+++ b/webknossos/webknossos/geometry/nd_bounding_box.py
@@ -555,8 +555,8 @@ class NDBoundingBox:
         """
         Returns the intersection of two bounding boxes.
         If there is no intersection (resulting bounding box is empty) and dont_assert is False (default)
-        the function raises an assertion error. 
-        If dont_assert is set to True this method returns an empty bounding box if there is no intersection.  
+        the function raises an assertion error.
+        If dont_assert is set to True this method returns an empty bounding box if there is no intersection.
 
         Args:
             other (NDBoundingBox): The other bounding box to intersect with.
--- a/webknossos/webknossos/geometry/nd_bounding_box.py
+++ b/webknossos/webknossos/geometry/nd_bounding_box.py
@@ -555,12 +555,12 @@ class NDBoundingBox:
         """
         Returns the intersection of two bounding boxes.
         If there is no intersection (resulting bounding box is empty) and dont_assert is False (default)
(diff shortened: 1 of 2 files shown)


webknossos/webknossos/geometry/nd_bounding_box.py
```diff
@@ -554,12 +554,13 @@ def padded_with_margins(
...
-
-        If dont_assert is set to False, this method may return empty bounding boxes (size == (0, 0, 0))
+        If there is no intersection (resulting bounding box is empty) and dont_assert is False (default)
+        the function asserts. 
+        If dont_assert is set to True this method returns an empty bounding box if there is no intersection.  
 
         Args:
             other (NDBoundingBox): The other bounding box to intersect with.
-            dont_assert (bool): If True, the method may return empty bounding boxes.
+            dont_assert (bool): If True, the method may return empty bounding boxes. Default is False.
```
hotzenklotz, line 563: It only return s a single BB, right? Therefore, the use of plural is not correct here.
```suggestion
            dont_assert (bool): If True, the method may return an empty bounding box. Default is False.
```

webknossos/webknossos/geometry/nd_bounding_box.py
```diff
@@ -554,12 +554,13 @@ def padded_with_margins(
     def intersected_with(self: _T, other: _T, dont_assert: bool = False) -> _T:
         """
         Returns the intersection of two bounding boxes.
-
-        If dont_assert is set to False, this method may return empty bounding boxes (size == (0, 0, 0))
+        If there is no intersection (resulting bounding box is empty) and dont_assert is False (default)
+        the function asserts.
```
hotzenklotz, line 558:
```suggestion
        the function raises an assertion error. 
```

hotzenklotz (changes requested):
Thanks for the contribution. Please see my comments.

hotzenklotz:
According to the CI you need to properly format your changes to pass. Please, run `./format.sh` for automatic formatting with `ruff`.

webknossos/webknossos/geometry/nd_bounding_box.py
xfolka: suggestions are implemented

hotzenklotz (approved):

PR #1187, opened 2024-09-04
Add option to add remote mags and remote layer to an existing dataset

MichaelBuessemeyer:
### Description:
This PR adds option to add remote mags and remote layer to an existing dataset via `dataset.add_remote_layer(layer / URL, layer_name` and `layer.add_remote_mag(mag / URL). 

When using `add_remote_mag` make sure that the layer to which the mag is added has matching dtype, category and so on.

### Testing:
- I wrote tests for this, but the tests working with the plain URLs somehow do not work. Maybe I need to configure / code the tests differently :see_no_evil: 
I used this script for local testing. Run a local wk to make this work. Else change the local host URLs to webknossos.org urls.
```
small_dataset = Dataset.open(<local dataset path>)
new_mag = MagView._ensure_mag_view(
    "http://localhost:9000/data/zarr/sample_organization/l4_sample/segmentation/1"
)
new_layer = small_dataset.add_layer(
    "test_remote_layer",
    "segmentation",
    data_format=new_mag.info.data_format,
    dtype_per_channel=new_mag.get_dtype(),
)
new_layer.add_remote_mag(new_mag)
new_mag = MagView._ensure_mag_view(
    "http://localhost:9000/data/zarr/sample_organization/l4_sample/segmentation/2-2-1/"
)
new_layer.add_remote_mag(new_mag)
new_layer.add_remote_mag(
    "http://localhost:9000/data/zarr/sample_organization/l4_sample/segmentation/4-4-1/"
)
small_dataset.add_remote_layer(
    "http://localhost:9000/data/zarr/sample_organization/l4_sample/segmentation",
    "another_segmentation",
)
```

### Issues:
- fixes #https://github.com/scalableminds/voxelytics/issues/3469 partially only

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -3,6 +3,7 @@ from tempfile import TemporaryDirectory
 from typing import Iterator
 
 import pytest
+from upath import UPath
 
 import webknossos as wk
 from webknossos import Dataset, MagView
@@ -66,6 +67,33 @@ def test_add_remote_mags_from_mag_view(
         ), "Added remote mag's path does not match remote path of mag added."
 
 
+@pytest.mark.skip(
+    reason="The test is flaky on the when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
+)
+def test_add_remote_mags_from_path(
+    sample_remote_mags: list[wk.MagView],
+    sample_remote_dataset: wk.Dataset,
+) -> None:
+    for remote_mag in sample_remote_mags:
+        mag_path = remote_mag.path
+        layer_type = remote_mag.layer.category
+        assert is_remote_path(mag_path), "Remote mag does not have remote path."
+        # Additional .parent calls are needed as the first .parent only removes the trailing slash.
+        layer_name = f"test_remote_layer_{mag_path.parent.name}_{mag_path.name}_path"
+        new_layer = sample_remote_dataset.add_layer(
+            layer_name,
+            layer_type,
+            data_format=remote_mag.info.data_format,
+            dtype_per_channel=remote_mag.get_dtype(),
+        )
+        new_layer.add_remote_mag(str(remote_mag.path))
+        added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
+        # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
+        assert (
+            added_mag.path == mag_path or added_mag.path == mag_path.parent
+        ), "Added remote mag's path does not match remote path of mag added."
+
+
 def test_add_remote_layer_from_object(
     sample_remote_layer: list[wk.Layer], sample_remote_dataset: wk.Dataset
 ) -> None:
@@ -77,4 +105,21 @@ def test_add_remote_layer_from_object(
         assert (
             is_remote_path(new_layer.path)
             and layer.path.as_uri() == new_layer.path.as_uri()
+        ), "Added layer should have a remote path matching the remote layer added."
+
+
+@pytest.mark.skip(
+    reason="The test is flaky on the when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
+)
+def test_add_remote_layer_from_path(
+    sample_remote_layer: list[wk.Layer],
+    sample_remote_dataset: wk.Dataset,
+) -> None:
+    for layer in sample_remote_layer:
+        assert is_remote_path(layer.path), "Remote mag does not have remote path."
+        layer_name = f"test_remote_layer_{layer.category}_path"
+        sample_remote_dataset.add_remote_layer(UPath(layer.path), layer_name)
+        new_layer = sample_remote_dataset.layers[layer_name]
+        assert (
+            is_remote_path(new_layer.path) and new_layer.path == layer.path
         ), "Added layer should have a remote path matching the remote layer added."
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -33,9 +33,7 @@ def sample_remote_mags() -> list[wk.MagView]:
         "https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/segmentation/2-2-1/",
         "https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/segmentation/4-4-2/",
     ]
-    mags = []
-    for url in mag_urls:
-        mags.append(MagView._ensure_mag_view(url))
+    mags = [MagView._ensure_mag_view(url) for url in mag_urls]
     return mags
 
 
@@ -61,8 +59,6 @@ def test_add_remote_mags_from_mag_view(
             data_format=remote_mag.info.data_format,
             dtype_per_channel=remote_mag.get_dtype(),
         )
-        print("using path")
-        print(remote_mag.path)
         new_layer.add_foreign_mag(remote_mag)
         added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
         # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
@@ -87,8 +83,6 @@ def test_add_remote_mags_from_path(
             data_format=remote_mag.info.data_format,
             dtype_per_channel=remote_mag.get_dtype(),
         )
-        print("using path")
-        print(remote_mag.path)
         new_layer.add_remote_mag(remote_mag.path)
         added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
         # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
@@ -105,7 +99,8 @@ def test_add_remote_layer_from_object(
         sample_remote_dataset.add_foreign_layer(layer, layer_name)
         new_layer = sample_remote_dataset.layers[layer_name]
         assert (
-            is_remote_path(new_layer.path) and layer.path.as_uri() == new_layer.path.as_uri()
+            is_remote_path(new_layer.path)
+            and layer.path.as_uri() == new_layer.path.as_uri()
         ), "Added layer should have a remote path matching the remote layer added."
 
 
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -51,7 +51,6 @@ def test_add_remote_mags_from_mag_view(
         mag_path = remote_mag.path
         layer_type = remote_mag.layer.category
         assert is_remote_path(mag_path), "Remote mag does not have remote path."
-        # Additional .parent calls are needed as the first .parent only removes the trailing slash.
         layer_name = f"test_remote_layer_{mag_path.parent.name}_{mag_path.name}_object"
         new_layer = sample_remote_dataset.add_layer(
             layer_name,
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -68,7 +68,7 @@ def test_add_remote_mags_from_mag_view(
 
 
 @pytest.mark.skip(
-    reason="The test is flaky on the when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
+    reason="The test is flaky when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
 )
 def test_add_remote_mags_from_path(
     sample_remote_mags: list[wk.MagView],
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -58,7 +58,7 @@ def test_add_remote_mags_from_mag_view(
             data_format=remote_mag.info.data_format,
             dtype_per_channel=remote_mag.get_dtype(),
         )
-        new_layer.add_foreign_mag(remote_mag)
+        new_layer.add_remote_mag(remote_mag)
         added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
         # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
         assert (
@@ -82,7 +82,7 @@ def test_add_remote_mags_from_path(
             data_format=remote_mag.info.data_format,
             dtype_per_channel=remote_mag.get_dtype(),
         )
-        new_layer.add_foreign_mag(str(remote_mag.path))
+        new_layer.add_remote_mag(str(remote_mag.path))
         added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
         # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
         assert added_mag.path == mag_path or added_mag.path == mag_path.parent, "Added remote mag's path does not match remote path of mag added."
@@ -95,7 +95,7 @@ def test_add_remote_layer_from_object(
     for layer in sample_remote_layer:
         assert is_remote_path(layer.path), "Remote mag does not have remote path."
         layer_name = f"test_remote_layer_{layer.category}_object"
-        sample_remote_dataset.add_foreign_layer(layer, layer_name)
+        sample_remote_dataset.add_remote_layer(layer, layer_name)
         new_layer = sample_remote_dataset.layers[layer_name]
         assert (
             is_remote_path(new_layer.path)
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -78,4 +78,3 @@ def test_add_remote_layer_from_object(
             is_remote_path(new_layer.path)
             and layer.path.as_uri() == new_layer.path.as_uri()
         ), "Added layer should have a remote path matching the remote layer added."
-
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -109,7 +109,7 @@ def test_add_remote_layer_from_object(
 
 
 @pytest.mark.skip(
-    reason="The test is flaky on the when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
+    reason="The test is flaky when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
 )
 def test_add_remote_layer_from_path(
     sample_remote_layer: list[wk.Layer],
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -35,7 +35,6 @@ def sample_remote_mags() -> list[wk.MagView]:
     ]
     mags = []
     for url in mag_urls:
-        print("calling ensure mag on", url)
         mags.append(MagView._ensure_mag_view(url))
     return mags
 
@@ -64,7 +63,7 @@ def test_add_remote_mags_from_mag_view(
         )
         print("using path")
         print(remote_mag.path)
-        new_layer.add_remote_mag(remote_mag)
+        new_layer.add_foreign_mag(remote_mag)
         added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
         # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
         assert (
@@ -103,10 +102,10 @@ def test_add_remote_layer_from_object(
     for layer in sample_remote_layer:
         assert is_remote_path(layer.path), "Remote mag does not have remote path."
         layer_name = f"test_remote_layer_{layer.category}_object"
-        sample_remote_dataset.add_remote_layer(layer, layer_name)
+        sample_remote_dataset.add_foreign_layer(layer, layer_name)
         new_layer = sample_remote_dataset.layers[layer_name]
         assert (
-            is_remote_path(new_layer.path) and new_layer.path == layer.path
+            is_remote_path(new_layer.path) and layer.path.as_uri() == new_layer.path.as_uri()
         ), "Added layer should have a remote path matching the remote layer added."
 
 
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -66,29 +66,6 @@ def test_add_remote_mags_from_mag_view(
         ), "Added remote mag's path does not match remote path of mag added."
 
 
-"""
-def test_add_remote_mags_from_path(
-    sample_remote_mags: list[wk.MagView], tmp_path: Path, sample_remote_dataset: wk.Dataset
-) -> None:
-    for remote_mag in sample_remote_mags:
-        mag_path = remote_mag.path
-        layer_type = remote_mag.layer.category
-        assert is_remote_path(mag_path), "Remote mag does not have remote path."
-        # Additional .parent calls are needed as the first .parent only removes the trailing slash.
-        layer_name = f"test_remote_layer_{mag_path.parent.name}_{mag_path.name}_path"
-        new_layer = sample_remote_dataset.add_layer(
-            layer_name,
-            layer_type,
-            data_format=remote_mag.info.data_format,
-            dtype_per_channel=remote_mag.get_dtype(),
-        )
-        new_layer.add_remote_mag(str(remote_mag.path))
-        added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
-        # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
-        assert added_mag.path == mag_path or added_mag.path == mag_path.parent, "Added remote mag's path does not match remote path of mag added."
-"""
-
-
 def test_add_remote_layer_from_object(
     sample_remote_layer: list[wk.Layer], sample_remote_dataset: wk.Dataset
 ) -> None:
@@ -102,17 +79,3 @@ def test_add_remote_layer_from_object(
             and layer.path.as_uri() == new_layer.path.as_uri()
         ), "Added layer should have a remote path matching the remote layer added."
 
-
-"""
-@pytest.skip("skipping")
-def test_add_remote_layer_from_path(
-    sample_remote_layer: list[wk.Layer], tmp_path: Path, sample_remote_dataset: wk.Dataset
-) -> None:
-    for layer in sample_remote_layer:
-        assert is_remote_path(layer.path), "Remote mag does not have remote path."
-        layer_name = f"test_remote_layer_{layer.category}_path"
-        sample_remote_dataset.add_remote_layer(UPath(layer.path), layer_name)
-        new_layer = sample_remote_dataset.layers[layer_name]
-        assert is_remote_path(new_layer.path) and new_layer.path == layer.path, "Added layer should have a remote path matching the remote layer added."
-
-"""
--- a/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
+++ b/webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
@@ -1,0 +1,125 @@
+from pathlib import Path
+from tempfile import TemporaryDirectory
+from typing import Iterator
+
+import pytest
+
+import webknossos as wk
+from webknossos import Dataset, MagView
+from webknossos.utils import is_remote_path
+
+pytestmark = [pytest.mark.with_vcr]
+
+
+@pytest.fixture(scope="module")
+def sample_bbox() -> wk.BoundingBox:
+    return wk.BoundingBox((2807, 4352, 1794), (10, 10, 10))
+
+
+@pytest.fixture(scope="module")
+def sample_remote_dataset(sample_bbox: wk.BoundingBox) -> Iterator[wk.Dataset]:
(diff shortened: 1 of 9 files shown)


MichaelBuessemeyer:
It seems to me like the libs already support adding new mags / layers to remote datasets. Therefore, maybe calling the new methods `add_foreign_mag/layer` might be better suited.

webknossos/webknossos/dataset/layer.py
```diff
@@ -532,16 +590,18 @@ def delete_mag(self, mag: Union[int, str, list, tuple, np.ndarray, Mag]) -> None
                 "Deleting mag {} failed. There is no mag with this name".format(mag)
             )
 
+        full_mag_path = _find_mag_path(
+            self.dataset.path, self.name, mag.to_layer_name(), self.mags[mag].path
+        )
+        assert is_fs_path(full_mag_path), "Cannot delete remote mags."
+
```
MichaelBuessemeyer, line 597: This throws an error in the tests. It is really true, that wklibs support deleting remote mags? e.g. on s3?

MichaelBuessemeyer:
> Therefore, maybe calling the new methods add_foreign_mag/layer might be better suited.

@daniel-wer I am now renaming the methods to `*_foreign_*` as this makes more sense as wklibs already supports handling remote datasets, but the new feature itself is having a layer / mag from another foreign dataset

webknossos/webknossos/dataset/layer.py
```diff
@@ -451,6 +475,7 @@ def add_mag_for_existing_files(
                     if mag_array_info.data_format == DataFormat.WKW
                     else None
                 ),
+                # TO DO: I think nd support is missing here.
```
MichaelBuessemeyer, line 478: I just noticed this while reading the code. Should I open an issue for this? Or do you think this is irrelevant?

webknossos/webknossos/dataset/mag_view.py
```diff
@@ -32,11 +33,15 @@
...
+def _find_mag_path(
+    dataset_path: Path,
+    layer_name: str,
+    mag_name: str,
+    path: Optional[str | Path] = None,
 ) -> Path:
+    path = UPath(path) if path else None
     if path is not None:
-        return dataset_path / path
+        return path
```
MichaelBuessemeyer, line 45: This is one of the dangerous parts of this pr imo:

I changed the functionality so that in case a mag_path is passed to the method, the path is used and not post appended to the dataset path. This was necessary to make it work with foreign mags as they have absolute paths, which would break in the `return dataset_path / path` code line.

=> the dangerous code parts are where the fourth parameter (path) is actually passed to this method. I marked the code lines

webknossos/webknossos/dataset/layer.py
```diff
@@ -1092,18 +1187,21 @@ def _setup_mag(self, mag: Mag, path: Optional[str] = None) -> None:
         mag_name = mag.to_layer_name()
 
         self._assert_mag_does_not_exist_yet(mag)
-
+        mag_path_maybe = UPath(path) if path else path
         try:
             cls_array = BaseArray.get_class(self._properties.data_format)
-            info = cls_array.open(
-                _find_mag_path_on_disk(self.dataset.path, self.name, mag_name, path)
-            ).info
```
MichaelBuessemeyer, line 1100: This is the only code line where the fourth argument is passed to `_find_mag_path_on_disk`. See my comment below.

My problem is that I do not know the code base well enough to foresee a scenario where my changes might break the code.

MichaelBuessemeyer (commented):
Ok that should be it. I commented everything that I think might be dangerous changes / changes about which I am unsure whether they might be breaking some wklibs functionallity.

daniel-wer:
Should @markbader review this, maybe?

MichaelBuessemeyer:
>Should @markbader review this, maybe?

Yeah maybe. to me it looks like a state for a first review :)

daniel-wer:
> It seems to me like the libs already support adding new mags / layers to remote datasets. Therefore, maybe calling the new methods add_foreign_mag/layer might be better suited.

Can you clarify why you renamed this? As far as I understand, the libs support opening remote datasets and it's possible to add layers and mags to them. Do you think `add_remote_mag` could be misinterpreted to mean that this is the method to add a mag to a remote dataset? I would not expect this to be an issue since the method is not called `add_mag_to_remote_...`. Instead, I think it might be confusing users what "foreign" mags or layers are. Using the term "remote" would be more clear as that terminology is used in other parts of the API as well (i.e. remote datasets). What do you and @markbader think? :)

daniel-wer:
I'm not sure whether I fully understood, but are you saying the method you introduced `add_remote/foreign_layer` can be used to add a layer from the local file system to a remote dataset?
I would have expected it to not matter where a dataset is actually saved. Doesn't wklibs usually abstract from that (using upath etc.). Instead, I would have expected `add_remote/foreign_layer` only to be used when actually adding a remote layer :thinking: 
Or did I misunderstand you?

MichaelBuessemeyer:
> I'm not sure whether I fully understood, but are you saying the method you introduced add_remote/foreign_layer can be used to add a layer from the local file system to a remote dataset?

Yes, this is one (untested) way the new methods can be used. 

> I would have expected it to not matter where a dataset is actually saved.

Yes, that's the idea. But my reasoning was in a specific scenario where the dataset itself is remote and the added layer is a local (from disk) one.


Ok I think we have two different images of what is remote in mind: I think this image describes it very well: I think your understanding of remote is from the datasets perspective while mine is from the perspective of where wklibs is running.

image


Does that make sense? Honestly, I do not have a strong preference on which mental model to decided. Which was the one used before? yours @daniel-wer ?

daniel-wer:
Thank you for explaining! I'll think out loud.

I think the current "mental model" is that remote datasets/layers/mags are those that are not on the "same server" (accessed via file system) as the webknossos datastore or where webknossos-libs runs. At least that is my understanding and reading of the docs. See for example, [this section](https://docs.webknossos.org/webknossos/datasets.html#:~:text=streaming%20them%20from%20a%20remote%20server/the%20cloud) or [this example](https://docs.webknossos.org/webknossos-py/examples/remote_datasets.html).
The use case we want to enable in voxelytics is to be able to predict on remote datasets/layers/mags which in turn requires to be able to add the remote layers/mags to local datasets. To do so, I still find the term add_remote_layer/mag well aligned with the current mental model.
You want to enable another use case, namely adding a local layer to a remote dataset. I guess that's probably a little more difficult than the first use case, because the remote dataset does not have a representation on the local file system, but rather the layer is added in the webknossos database, right? Does that require to label the layer as remote or foreign? Sorry I'm not familiar with that code.

In any case, maybe we don't overcomplicate things and focus on the first use case or is there an issue for the second one as well?

MichaelBuessemeyer (commented):
Thanks @markbader for the feedback. It should now be applied

webknossos/webknossos/dataset/layer.py
```diff
@@ -201,7 +203,26 @@ def __init__(self, dataset: "Dataset", properties: LayerProperties) -> None:
...
+            strip_trailing_slash(UPath(maybe_mag_path_str))
+            if maybe_mag_path_str
+            else None
+        )
+        is_remote = maybe_mag_path_upath and is_remote_path(maybe_mag_path_upath)
+        return (
+            maybe_mag_path_upath.parent
+            if maybe_mag_path_upath and is_remote
+            else self.dataset.path / self.name
+        )
```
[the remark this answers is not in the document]
normanrz: Maybe we should raise an exception if the layer path cannot be uniquely inferred?

daniel-wer:
@MichaelBuessemeyer What do you think about my last comment?

If possible, I'd like to get this merged and pulled into voxelytics in the next week, before I'm out of office :) Let me know if there are any open discussions or other blockers for you!

webknossos/webknossos/dataset/mag_view.py
```diff
@@ -32,11 +33,15 @@
 from .view import View
 
 
-def _find_mag_path_on_disk(
-    dataset_path: Path, layer_name: str, mag_name: str, path: Optional[str] = None
+def _find_mag_path(
+    dataset_path: Path,
+    layer_name: str,
+    mag_name: str,
+    path: Optional[str | Path] = None,
```
markbader, line 40:
```suggestion
    path: Optional[Union[str, Path]] = None,
```

The Union notation you used is only supported since python 3.10, as we support python 3.9 as well we need the `Union[...]` notation.

webknossos/webknossos/dataset/layer.py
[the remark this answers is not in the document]
MichaelBuessemeyer: I added an assertion :tada:

webknossos/webknossos/dataset/mag_view.py
MichaelBuessemeyer: Oh nice, thanks for finding this :pray:

MichaelBuessemeyer:
> To do so, I still find the term add_remote_layer/mag well aligned with the current mental model.

In this specific scenario, I agree with you. But I think the add remote /foreign layer / mag is more mighty than this. At least it has the potential add layers located on any server or accessible file system. 

> You want to enable another use case, namely adding a local layer to a remote dataset. I guess that's probably a little more difficult than the first use case, because the remote dataset does not have a representation on the local file system, but rather the layer is added in the webknossos database, right? Does that require to label the layer as remote or foreign? Sorry I'm not familiar with that code.

Yeah that's definitely more complicated. What happens is that wklibs updates the datasource-properties.json which include the path to each mag. Thus, the wk server & wklibs should be able to access layers from other "remote" datasets.

> In any case, maybe we don't overcomplicate things and focus on the first use case or is there an issue for the second one as well?

sure :+1: 
 Done

webknossos/webknossos/dataset/mag_view.py
MichaelBuessemeyer: Nice the ci is green now :tada:

markbader (commented):
Looks good to me now. I just commented on the Datasets `is_remote_dataset` property, as I can't see the value of it yet.

webknossos/webknossos/dataset/dataset.py
```diff
@@ -327,6 +328,7 @@ def __init__(
 
         self.path: Path = dataset_path
         self._properties: DatasetProperties = self._load_properties()
+        self.is_remote_dataset = is_remote_path(self.path)
```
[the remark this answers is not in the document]
MichaelBuessemeyer: > Is this property used anywhere else?

Nope, good find :pray: 

> Are there cases where we have a Dataset with a remote path?

Na, not really. I guess this is just code legacy from previous versions of this pr. Thanks for finding this. I removed it :)

webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
```diff
@@ -0,0 +1,118 @@
...
+        )
+        new_layer.add_remote_mag(remote_mag)
+        added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
+        # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
+        assert (
+            added_mag.path == mag_path or added_mag.path == mag_path.parent
+        ), "Added remote mag's path does not match remote path of mag added."
+
+
+"""
```
daniel-wer, line 69: Why are the two tests in this file commented out? If there's a valid reason for it, please add an explaining comment :)

daniel-wer:
@MichaelBuessemeyer @markbader Thank you for your work on this!
I just saw that `webknossos.utils` includes an `is_fs_path` method. Could you clarify whether that's the exact opposite of the `is_remote_path` method (meaning the one returns true iff the other returns false and vice versa)? If so, one could implement one using the other to make that more clear, and it would also simplify/unify some imports in voxelytics if one of the two methods could be used exclusively.

webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
MichaelBuessemeyer: Thanks for finding this. The problem with the commented out tests is that they fail reliably and I don't know why. When I run the `test_add_remote_mags_from_path` test, loading the datasource properties json from `https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json` succeeds n times but the n+1st time fails. I added the following logging to `dataset.py _load_properties`:
```python
    def _load_properties(self) -> DatasetProperties:
        with (self.path / PROPERTIES_FILE_NAME).open(
            encoding="utf-8"
        ) as datasource_properties:
            contents = datasource_properties.read()
            print("file path")
            print(self.path / PROPERTIES_FILE_NAME)
            print("--------------------------")
            print("datasource_properties contents")
            print(contents)
            data = json.loads(contents)
        return dataset_converter.structure(data, DatasetProperties)
```
And the output of the failed test is:
```
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json
--------------------------
datasource_properties contents
{"id":{"name":"l4_sample_dev","team":"scalable_minds"},"dataLayers":[{"name":"color","category":"color","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint8","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"defaultViewConfiguration":{"color":[255,219,219],"intensityRange":[70,180]},"numChannels":1,"dataFormat":"zarr"},{"name":"segmentation","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint32","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"largestSegmentId":5238529,"numChannels":1,"dataFormat":"zarr","category":"segmentation"}],"scale":{"factor":[11.239999771118164,11.239999771118164,28],"unit":"nanometer"}}
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json
--------------------------
datasource_properties contents
{"id":{"name":"l4_sample_dev","team":"scalable_minds"},"dataLayers":[{"name":"color","category":"color","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint8","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"defaultViewConfiguration":{"color":[255,219,219],"intensityRange":[70,180]},"numChannels":1,"dataFormat":"zarr"},{"name":"segmentation","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint32","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"largestSegmentId":5238529,"numChannels":1,"dataFormat":"zarr","category":"segmentation"}],"scale":{"factor":[11.239999771118164,11.239999771118164,28],"unit":"nanometer"}}
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json
--------------------------
datasource_properties contents
{"id":{"name":"l4_sample_dev","team":"scalable_minds"},"dataLayers":[{"name":"color","category":"color","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint8","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"defaultViewConfiguration":{"color":[255,219,219],"intensityRange":[70,180]},"numChannels":1,"dataFormat":"zarr"},{"name":"segmentation","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint32","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"largestSegmentId":5238529,"numChannels":1,"dataFormat":"zarr","category":"segmentation"}],"scale":{"factor":[11.239999771118164,11.239999771118164,28],"unit":"nanometer"}}
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json
--------------------------
datasource_properties contents
{"id":{"name":"l4_sample_dev","team":"scalable_minds"},"dataLayers":[{"name":"color","category":"color","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint8","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"defaultViewConfiguration":{"color":[255,219,219],"intensityRange":[70,180]},"numChannels":1,"dataFormat":"zarr"},{"name":"segmentation","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint32","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"largestSegmentId":5238529,"numChannels":1,"dataFormat":"zarr","category":"segmentation"}],"scale":{"factor":[11.239999771118164,11.239999771118164,28],"unit":"nanometer"}}
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json
--------------------------
datasource_properties contents
{"id":{"name":"l4_sample_dev","team":"scalable_minds"},"dataLayers":[{"name":"color","category":"color","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint8","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"defaultViewConfiguration":{"color":[255,219,219],"intensityRange":[70,180]},"numChannels":1,"dataFormat":"zarr"},{"name":"segmentation","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint32","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"largestSegmentId":5238529,"numChannels":1,"dataFormat":"zarr","category":"segmentation"}],"scale":{"factor":[11.239999771118164,11.239999771118164,28],"unit":"nanometer"}}
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json
--------------------------
datasource_properties contents
{"id":{"name":"l4_sample_dev","team":"scalable_minds"},"dataLayers":[{"name":"color","category":"color","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint8","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"defaultViewConfiguration":{"color":[255,219,219],"intensityRange":[70,180]},"numChannels":1,"dataFormat":"zarr"},{"name":"segmentation","boundingBox":{"topLeft":[2679,4224,1719],"width":256,"height":256,"depth":150},"elementClass":"uint32","mags":[{"mag":[1,1,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[2,2,1],"axisOrder":{"x":1,"y":2,"z":3,"c":0}},{"mag":[4,4,2],"axisOrder":{"x":1,"y":2,"z":3,"c":0}}],"largestSegmentId":5238529,"numChannels":1,"dataFormat":"zarr","category":"segmentation"}],"scale":{"factor":[11.239999771118164,11.239999771118164,28],"unit":"nanometer"}}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": []
}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": []
}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
Downloading layer=color mag=1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
Downloading layer=color mag=2-2-1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
Downloading layer=color mag=4-4-2 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
Downloading layer=segmentation mag=1 ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
Downloading layer=segmentation mag=2-2-1 ━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
Downloading layer=segmentation mag=4-4-2 ━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00
----------------------------------------------------------------------------------------------- Captured stdout call ------------------------------------------------------------------------------------------------
file path
/tmp/tmp5izrse_t/ds/datasource-properties.json
--------------------------
datasource_properties contents
{
    "id": {
        ...
    },
    "scale": {
        ...
    },
    "dataLayers": [
        ...
    ]
}
file path
https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json

--------------------------
datasource_properties contents


```

As you can see the output shows the first requests to `https://data-humerus.webknossos.org/data/zarr/scalable_minds/l4_sample_dev/datasource-properties.json` result in a valid json result but the last logged access to the file just returns an empty string :/


Very random thought @fm3 is maybe some kind of ddos protection / rate limiting active for `data-humerus.webknossos.org` that might explain this behaviour :thinking:?

MichaelBuessemeyer: But nevertheless, I remove the commented out tests as they can be recovered from the commit history of this branch

fm3: Shouldn’t think that there is such a rate limiting in place there. Could you also print the request reply etc? This is opened using upath, right? Maybe it doesn’t handle the requests correctly?
Also, why does the test have to open the json so many times?

MichaelBuessemeyer:
> @MichaelBuessemeyer @markbader Thank you for your work on this!
I just saw that webknossos.utils includes an is_fs_path method. Could you clarify whether that's the exact opposite of the is_remote_path method (meaning the one returns true iff the other returns false and vice versa)? If so, one could implement one using the other to make that more clear, and it would also simplify/unify some imports in voxelytics if one of the two methods could be used exclusively.

I'd say they are not strictly the opposite of each other as there are more implementations of upath which are not checked in either of the methods. See the list of official upath implementations from the gh repo

image

But I think we only use the few implementations that are checked in `is_fs_path` and `is_remote_path`. So I'll give it a try to define `is_remote_path` as the bool'sche negation of `is_fs_path` and lets see what the ci says. (Although the ci currently fails due to unknown reasons to me :/

@markbader Maybe you have an Idea why this is failing :see_no_evil:?

MichaelBuessemeyer:
> But I think we only use the few implementations that are checked in is_fs_path and is_remote_path. So I'll give it a try to define is_remote_path as the bool'sche negation of is_fs_path and lets see what the ci says. (Although the ci currently fails due to unknown reasons to me :/

Ok this should work (at least the ci still fails at the same point as before)
Here is the CI error:
```
FAILED tests/dataset/test_dataset.py::test_ome_ngff_metadata[output_path1] - FileNotFoundError: https://ngff.openmicroscopy.org/0.4/schemas/image.schema
```

Strangely `tests/dataset/test_dataset.py::test_ome_ngff_metadata[output_path0]` does not fail :woozy_face:

markbader:
I get the same CI error in https://github.com/scalableminds/webknossos-libs/pull/1196 since yesterday, Both output_path0 and output_path1 fail but they are splitted in the CI (so output_path0 fails in webknossos_linux(3.12,1) and output_path1 in webknossos_linux(3.12, 3)). I don't know where this error comes from, but I think it is unrelated to the failing tests that you wrote. I had this json decode error before and thought they might be caused by vcr-py. But I didn't found a good way for debugging this.

normanrz:
I also think, this might be related to vcr-py. All the more reason to get rid of it. Can we skip these tests for now and move on with merging?

MichaelBuessemeyer:
> I also think, this might be related to vcr-py. All the more reason to get rid of it. Can we skip these tests for now and move on with merging?


ok on it :)

webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
```diff
@@ -0,0 +1,125 @@
...
+        new_layer.add_remote_mag(remote_mag)
+        added_mag = sample_remote_dataset.layers[layer_name].mags[remote_mag.mag]
+        # checking whether the added_mag.path matches the mag_url with or without a trailing slash.
+        assert (
+            added_mag.path == mag_path or added_mag.path == mag_path.parent
+        ), "Added remote mag's path does not match remote path of mag added."
+
+
+@pytest.mark.skip(
+    reason="The test is flaky on the when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
```
markbader, line 71:
```suggestion
    reason="The test is flaky when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
```

webknossos/tests/dataset/test_dataset_add_remote_mag_and_layer.py
```diff
@@ -0,0 +1,125 @@
...
+        sample_remote_dataset.add_remote_layer(layer, layer_name)
+        new_layer = sample_remote_dataset.layers[layer_name]
+        assert (
+            is_remote_path(new_layer.path)
+            and layer.path.as_uri() == new_layer.path.as_uri()
+        ), "Added layer should have a remote path matching the remote layer added."
+
+
+@pytest.mark.skip(
+    reason="The test is flaky on the when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
```
markbader, line 112:
```suggestion
    reason="The test is flaky when trying to fetch the required datasource-properties.json from data-humerus.webknossos.org. Disable it for now."
```

markbader (approved):
LGTM :rocket: I noticed two minor spelling issues and a merge conflict needs to be solved, as I disabled the ngff_metadata test in another PR already. Then it should be ready to merge.

daniel-wer:
@MichaelBuessemeyer Sorry, I would have wanted to add a changelog entry before merging, but bulldozer was too eager. We've already spoken to Norman to reconfigure bulldozer to only merge when the automerge label is active. This is somehow misconfigured in this repo.

PR #1191, opened 2024-09-09
Fix merging annotations with compressed fallback layers

erjel:
### Description:
If an annotation is merged with a compressed fallback layer, the bounding box of the  `output_mag` needs to account for compressed shard writes outside the original bounding box beforehand. If the bounding box is not adapted accordingly prior to concurrent calls of `merge_chunk`, the merge operation will corrupt the target dataset properties and result in a `JSONDecodeError`.

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation -> not applicable
 - [x] Added / Updated Tests
 - [x] Considered adding this to the Examples -> not applicable


hotzenklotz (approved):
Hi, thanks for the contribution. Looks great!

PR #1201, opened 2024-10-11
Fix add existing tree to annotation

### LINKED ISSUE #1197: Cannot add existing tree to new annotation
## Context
Calling `.add_tree()` with an existing tree fails.

grafik
```
File ...\site-packages\webknossos\skeleton\[group.py:246](https://group.py:246/), in Group.__hash__(self)
    245 def __hash__(self) -> int:
--> 246     return self._id
 
AttributeError: 'Group' object has no attribute '_id'
```

MatthisCl:
### Description:
- When trying to deepcopy a group, hashfunction is needed before initialization. This leads to an error as self._id is not set before initialization. One could change the hashfunction to not be dependent on the id or change the hashfunction

### Issues:
- fixes #1197 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/skeleton/skeleton.py
```diff
@@ -282,5 +282,8 @@ def write(self, out_path: PathLike) -> None:
         warn_deprecated("Skeleton.write", "skeleton.save")
         self.save(out_path)
 
+    def __eq__(self, other: Any) -> bool:
+        return type(other) is type(self) and self._id == other._id
```
philippotto, line 286: this class inherits from group which also defines `__eq__`, right? couldn't the method here be omitted?

markbader: Yes, I think I tested it before, but it didn't work. I just figured that `@attr.define` adds a standard `__eq__` implementation. This can be avoided when adding `@attr.define(eq=False)`. Thanks for mentioning it, I removed the `__eq__` and the `__hash__` method from the `Skeleton` class.

philippotto (approved):
great, thank you!

PR #1204, opened 2024-10-24
Fix axes configuration when adding existing zarr arrays

markbader:
### Description:
- This PR fixes an issue where `cxyz` axes are expected for existing zarr arrays. The default behavior is changed and some assumptions about configuration of the axes are removed.

### Todos:
 - [x] Updated Changelog


normanrz (commented):
Could you please add a test for this case?

normanrz (approved):

PR #1209, opened 2024-11-07
fix buffer_bbox offset

MatthisCl:
### Description:
- Upsampling doesn't work as soon as the buffer in the upsample_cube_job needs to write more than once. Problem was a wrong offset calculation to determine position in the file buffer to write the data to.
- Added new assertion to test_upsampling, which now checks if the mean of the data in both layers is the same.

### Todos:
 - [x] Added / Updated Tests


markbader (approved):
Looks good to me, thanks for fixing this :rocket: 
 
 Could you add a line in the changelog, that this error was fixed?

PR #1212, opened 2024-11-14
Fixed PimsTiffReader

### LINKED ISSUE #1208: Single file tiff import leads to storage full
Converting a single tiff file can lead to filling up the storage of a computer. It occurs, when [mem-mapping of the tiff file](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/_utils/pims_tiff_reader.py#L50) fails.
In that case a new file is created where the data is being copied. That is suboptimal. Even worse, this is happening in the constructor which is called multiple times. I believe the copying should be avoided and the logic in `get_frame_2D` changed so that the necessary pages are loaded on-demand.

normanrz:
### Description:
- Refactors the `PimsTiffReader` to read the data directly from the tiff file without creating a memmap-able copy first.

### Issues:
- fixes #1208 

### Testing:
The tests ran through and I verified that the converted data is exactly the same as on current master.

### Todos:
 - [x] Updated Changelog


webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -461,34 +499,49 @@ def test_test_images(
...
+def test_test_images(
+    tmp_path: Path,
+    url: Union[str, List[str]],
+    filename: Union[str, List[str]],
+    kwargs: Dict,
+    dtype: str,
+    num_channels: int,
+    size: Tuple[int, int, int],
+) -> None:
+    _test_test_images(tmp_path, url, filename, kwargs, dtype, num_channels, size)
```
markbader, line 520: What was the purpose of adding this new method and just giving it all arguments that are already present in this test method?

normanrz: pytest complained about test functions having a non-None return value.

normanrz: Alternatively, we can remove the functionality to call these modules as CLI tools.

markbader (approved):
Looks good to me. I tested it with nd-data and a big tiff dataset, that causes huge memory usage beforehand. It works with this updated version. :rocket:

PR #1198, opened 2024-09-26
Replace vcr-py for tests

markbader:
### Description:
- This PR should remove vcr-py and introduce a proxy server that replays API requests for the tests.

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [x] Updated Changelog

#### Fixing tests:
- [x] Add annotation to local wk instance (e.g. tests/test_annotation.py::test_annotation_from_url)
- [x] Zarr Streaming doesn't seem to work for local wk instance(e.g. tests/test_annotation.py::test_bounding_box_roundtrip)
- [x] Remove date-strings from datasets generated within tests (e.g. tests/test_examples.py::test_upload_image_data)
- [x] Replaying responses for tests/dataset/test_dataset_add_remote_mag_and_layer.py does not work yet.

PR #1222, opened 2024-11-28
Fix Mag setup in case of non public UPath.

valentin-pinkau:
### Description:
- This PR fixes the setup of mags when their path is non public

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


markbader (approved):

markbader:
Failing tests are probably due to changed requests that are not yet captured by vcr-py. You might try to run `./test --refresh-snapshots` in the `webknossos` directory. But as vcr-py has some problems recently, I am not sure if it will fix all tests.

valentin-pinkau:
This is currently blocked by #1198

PR #1225, opened 2024-12-02
Added more doc strings

hotzenklotz:
### Description:
- I updated even more classes to use Google-Style doc strings and enhance the documentation that way

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [ ] Added / Updated Tests


webknossos/webknossos/dataset/mag_view.py
```diff
@@ -346,8 +564,35 @@ def get_views_on_disk(
...
+        Examples:
+            ```python
+            # Process each data file in parallel
+            def process_chunk(view: View) -> None:
+                data = view.read()
+                # Process data...
+                if not view.read_only:
+                    view.write(processed_data)
+
+            with ProcessPoolExecutor() as executor:
```
hotzenklotz, line 587: This is likely incorrect. Don't know the correct usage for executors, though.

webknossos/webknossos/dataset/mag_view.py
```diff
@@ -360,15 +605,33 @@ def compress(
...
+            args: ⚠️ Deprecated. Use executor parameter instead.
+            executor: Optional executor for parallel compression.
+
+        Examples:
+            ```python
+            # Compress in-place
+            mag1.compress()
+
+            # Compress to new location with parallel processing
+            with ProcessPoolExecutor() as executor:
```
hotzenklotz, line 624: This is likely incorrect. Don't know the correct usage, though.

webknossos/webknossos/dataset/mag_view.py
```diff
@@ -474,6 +737,30 @@ def merge_with_view(
...
+        merging annotations or overlays.
+
+        Args:
+            other: The MagView to merge into this one.
+            executor: Executor for parallel merging operations.
+
+        Examples:
+            ```python
+            # Merge annotation data over base layer
+            with ProcessPoolExecutor() as executor:
```
hotzenklotz, line 753: This is likely incorrect. Don't know the correct usage, though.

webknossos/webknossos/dataset/view.py
```diff
@@ -531,14 +641,39 @@ def read_xyz(
...
+                data. Areas outside the dataset are zero-padded.
+
+        Examples:
+            ```python
+            # Read entire view's data in XYZ order
+            view = layer.get_mag("1").get_view(size=(100, 100, 10))
+            xyz_data = view.read_xyz()  # Returns (X, Y, Z) array
+
+            # Read with relative bounding box
+            bbox = NDBoundingBox((10, 10, 0), (50, 50, 10))
```
markbader, line 668: NDBoundingBox constructor needs additional arguments e.g.:`axis=("x", "y", "z")` and `index=(1, 2, 3)`.

webknossos/webknossos/dataset/view.py
```diff
@@ -606,36 +740,68 @@ def get_view(
...
+                relative_offset=(50, 60, 70),
+                size=(999, 120, 230),
+                read_only=True
+            )
 
-        # works because `read_only=True`
-        valid_sub_view = view.get_view(relative_offset=(50, 60, 70), size=(999, 120, 230), read_only=True)
-        ```
+            # Use bounding box instead of offset+size
+            bbox = NDBoundingBox((10, 10, 0), (50, 50, 10))
```
markbader, line 795: Same as above, either use BoundingBox or add `axis` and `index` arguments.

webknossos/webknossos/dataset/view.py
```diff
@@ -768,35 +934,49 @@ def get_buffered_slice_writer(
...
+            buffer_size (int, optional): Number of slices to buffer before writing.
+                Defaults to 1.
+            relative_offset (Optional[Vec3IntLike], optional): Offset relative to view's
+                position in Mag(1). Defaults to None.
+            absolute_offset (Optional[Vec3IntLike], optional): Absolute offset in Mag(1).
+                Defaults to None.
+            relative_bounding_box (Optional[NDBoundingBox], optional): Bounding box relative
+                to view's position in Mag(1). Defaults to None.
+            absolute_bounding_box (Optional[NDBoundingBox], optional): Absolute bounding box
+                in Mag(1). Defaults to None.
```
markbader, line 952: This argument list is incomplete. Missing: offset (deprecated), dimension, json_update_allowed, use_logging

webknossos/webknossos/dataset/view.py
```diff
@@ -768,35 +934,49 @@ def get_buffered_slice_writer(
...
+            ```python
+            view = layer.get_mag("1").get_view(size=(100, 100, 10))
+
+            # Create a buffered writer with default settings
+            writer = view.get_buffered_slice_writer()
+
+            # Write slices efficiently
+            for z in range(10):
+                slice_data = np.zeros((100, 100))  # Your slice data
+                writer.write(slice_data)
```
markbader, line 967: The BufferedSliceWriter receives data when the `send` method is called.

webknossos/webknossos/dataset/view.py
```diff
@@ -828,31 +1008,53 @@ def get_buffered_slice_reader(
...
+            relative_offset (Optional[Vec3IntLike], optional): Offset relative to view's
+                position in Mag(1). Defaults to None.
+            absolute_offset (Optional[Vec3IntLike], optional): Absolute offset in Mag(1).
+                Defaults to None.
+            relative_bounding_box (Optional[NDBoundingBox], optional): Bounding box relative
+                to view's position in Mag(1). Defaults to None.
+            absolute_bounding_box (Optional[NDBoundingBox], optional): Absolute bounding box
+                in Mag(1). Defaults to None.
+            use_logging (bool, optional): Whether to log reading operations.
+                Defaults to False.
```
markbader, line 1031: This args list does not include offset and size

webknossos/webknossos/dataset/view.py
```diff
@@ -828,31 +1008,53 @@ def get_buffered_slice_reader(
...
+                process_slice(slice_data)
 
-        The reader must be used as a context manager using the `with` syntax (see example below).
-        Entering the context returns an iterator yielding slices (np.ndarray).
+            # Read y-slices with custom buffer size
+            reader = view.get_buffered_slice_reader(
+                buffer_size=10,
+                dimension=1,  # y-axis
+                relative_offset=(10, 0, 0)
+            )
```
markbader, line 1050: The BufferedSliceReader is a context manager. The usage shown in the examples is wrong. It should be used like:

```python
    view = ...
    with view.get_buffered_slice_reader() as reader:
        for slice_data in reader:
            ...
```

webknossos/webknossos/dataset/view.py
```diff
@@ -876,36 +1078,64 @@ def for_each_chunk(
...
+            executor (Optional[Executor], optional): Executor for parallel processing.
+                If None, processes chunks sequentially. Defaults to None.
+            progress_desc (Optional[str], optional): Description for progress bar.
+                If None, no progress bar is shown. Defaults to None.
+            chunk_size (Optional[Vec3IntLike], optional): ⚠️ Deprecated. Use chunk_shape instead.
+                Defaults to None.
+
+        Examples:
+            ```python
+            from concurrent.futures import ThreadPoolExecutor
```
markbader, line 1103: Why is a ThreadPoolExecutor imported here?

webknossos/webknossos/dataset/view.py
```diff
@@ -955,34 +1185,67 @@ def map_chunk(
...
+            stats = view.map_chunk(
+                named_partial(chunk_statistics, min_value=0.1),
+                chunk_shape=(128, 128, 128)
+            )
 
-        def some_work(view: View, some_parameter: int) -> None:
-            # perform operations on the view
-            ...
+            # Parallel processing with process pool
+            with ProcessPoolExecutor(max_workers=4) as executor:
```
markbader, line 1229: As above, we usually use `get_executor_for_args(None)`.

webknossos/webknossos/dataset/view.py
```diff
@@ -1139,6 +1451,46 @@ def content_is_equal(
...
+            bool: True if the content of both views is identical, False otherwise.
+
+        Examples:
+            ```python
+            # Compare views sequentially
+            if view1.content_is_equal(view2):
+                print("Views are identical")
+
+            # Compare in parallel with progress tracking
+            with ThreadPoolExecutor(max_workers=4) as executor:
```
markbader, line 1478: Same as above, `with get_executor_for_args() as executor`

webknossos/webknossos/dataset/view.py
```diff
@@ -1139,6 +1451,46 @@ def content_is_equal(
...
         executor: Optional[Executor] = None,
     ) -> bool:
+        """Compare the content of this view with another view.
+
+        Performs a chunk-by-chunk comparison of the data in both views. This is more
+        memory efficient than reading entire views at once for large datasets.
+
+        Args:
+            other (View): The view to compare against.
+            chunk_shape (Optional[Vec3IntLike], optional): Size of chunks to use for
```
markbader, line 1461: chunk_shape is no argument of this method.

webknossos/webknossos/dataset/view.py
```diff
@@ -1139,6 +1451,46 @@ def content_is_equal(
...
+                comparison in Mag(1). If None, uses file dimensions. Defaults to None.
+            executor (Optional[Executor], optional): Executor for parallel comparison.
+                If None, compares sequentially. Defaults to None.
+            progress_desc (Optional[str], optional): Description for progress bar.
+                If None, no progress bar is shown. Defaults to None.
```
markbader, line 1466: The deprecated `args` argument is missing in this list.

webknossos/webknossos/annotation/annotation.py
```diff
@@ -489,6 +509,40 @@ def open_as_remote_dataset(
...
+
+        This is a convenience method that combines downloading an annotation and converting it
+        to a remote dataset in one step. It's useful when you want to work with the annotation
+        data as a dataset without storing it locally.
+
+        Args:
+            annotation_id_or_url: Either an annotation ID or complete WEBKNOSSOS URL.
+                Example URL: https://webknossos.org/annotations/[id]
+            annotation_type: Optional type of annotation (Explorational, Task, etc.).
+                Deprecated and no longer required.
```
markbader, line 522: Most other deprecated arguments start with: `⚠️ Deprecated.`

markbader (changes requested):
Looks mostly good already. 
Nevertheless, I found some issues like missing arguments, wrongly used methods and as you already commented some issues with the usage of executors. As there are quite a lot of newly added comments, I hope that I found most of the mistakes in the generated doc strings.

hotzenklotz:
Thanks for the review. I applied most of your feedback and removed the parallel executor example for now. Perhaps that's too much detail for each method and should be covered somewhere else in more detail (in the future).

webknossos/webknossos/dataset/mag_view.py
```diff
@@ -474,6 +733,23 @@ def merge_with_view(
...
+        and overlaying the other view's data where present. This is particularly useful for
+        merging annotations or overlays.
+
+        Args:
+            other: The MagView to merge into this one.
+            executor: Executor for parallel merging operations.
+
+        Notes:
+            - Both views must have same magnification
+            - Other view must have chunks_per_shard = 1
```
markbader, line 748: Chunks per shard = 1 does not sound right, as the value must be specified for x, y and z. What is asserted in the function is that `chunks_per_shard` is 1 in all dimensions, i.e. that exactly one chunk is stored in a file. This property is called `file_len=1` in the assertion.
```suggestion
            - Other view must have file_len = 1
```

markbader (approved):
Looks much better. I found only a few issues and created suggestions for them.

webknossos/webknossos/dataset/view.py
```diff
@@ -48,11 +48,36 @@ def _assert_check_equality(args: Tuple["View", "View", int]) -> None:
...
+    Write operations are restricted to the defined bounding box. Views are designed to be easily passed
+    around as parameters and can be used to efficiently work with subsets of larger datasets.
+
+    Attributes:
+        _path (Path): Path to the mag view.
+        _array_info (ArrayInfo): Information about the array structure.
+        _bounding_box (Optional[NDBoundingBox]): The bounding box defining the view's region.
+        _read_only (bool): Whether the view is read-only.
+        _cached_array (Optional[BaseArray]): Cached array data.
+        _mag (Mag): Magnification level of the view.
```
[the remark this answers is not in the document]
hotzenklotz: Good catch. I updated the docstrings for the properties accordingly.

PR #1226, opened 2024-12-03
Add note that minio is necessary for tests on MacOS.

markbader:
### Description:
- Adds a line to the CONTRIBUTING.md file for installation of the test setup on MacOS.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,7 +140,7 @@ npm install --global proxay
 # or if you're using yarn
 yarn global add proxay
 ```
-When running the `./test.sh` script on iOS, it is also necessary to install `minio` using the following command:
+When running the `./test.sh` script on MacOS, it is also necessary to install `minio` using the following command:
 
 ```bash
 brew install minio
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -140,6 +140,11 @@ npm install --global proxay
 # or if you're using yarn
 yarn global add proxay
 ```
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -140,6 +140,11 @@ npm install --global proxay
 # or if you're using yarn
 yarn global add proxay
 ```
+When running the `./test.sh` script on iOS, it is also necessary to install `minio` using the following command:
```
hotzenklotz, line 143:
```suggestion
When running the `./test.sh` script on MacOS, it is also necessary to install `minio` using the following command:
```

hotzenklotz: iOS? That is the mobile phone OS from Apple. Do you mean MacOS, their desktop operating system?

hotzenklotz: And why is this an OS specific requirement? `minio` is an S3-API compatible object storage service. I believe we use that for testing S3 compatibility locally. Hence, you probably need that on Linux/(Windows) too if you want to run the tests, right?

hotzenklotz:
And why is this an OS specific requirement? minio is an S3-API compatible object storage service. I believe we use that for testing S3 compatibility locally. Hence, you probably need that on Linux/(Windows) too if you want to run the tests, right?

See https://vscode.dev/github/scalableminds/webknossos-libs/blob/more-docstrings/webknossos/tests/constants.py#L37

normanrz:
On Linux, we use docker for minio. On macOS, we use the locally installed minio. https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/tests/constants.py#L33-L51

PR #1227, opened 2024-12-03
set compression mode to LZ4 instead of LZ4HC.

valentin-pinkau:
### Description:
- LZ4 is ~8 times faster than LZ4HC at the cost of about 16% larger files for affinity predictions and 100% larger files for segmentations.
- Segmentations can still be compressed quite well, the compressed affinities (LZ4) are 15x larger than the compressed segmentations (LZ4)

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


valentin-pinkau:
We decided not to change the default.

PR #1174, opened 2024-08-25
tensorstore 

brokkoli71:
### Description:
- Use tensorstore for zarr arrays
- Reduces the number of IO ops when opening a dataset
- based on changes from #1026

### Todos:
 - [x] Updated Changelog
 - [X] Added / Updated Tests

PR #1229, opened 2024-12-11
Remove examples from tests

### LINKED ISSUE #1172: Enhance examples usability
## Detailed Description

We should have another look over the examples of the webknossos repository. Currently, the assumption is that they are read over the rendered documentation. Therefore, the docstring within the example files is rather short. Users that explore the GitHub Repo might need further explanations within the example python files.

Not all examples can be executed. Examples like announce dataset uploads are partly only described in comments and might be difficult to actually apply for users.

markbader:
### Description:
- The examples should contain less code for testing. This should increase readability and make the examples easier to use.

### Issues:
- fixes #1172

### Todos:
 - [ ] Updated Changelog
 - [ ] Added / Updated Tests

Add tests for things that are only tested in the examples:
- [ ] test apply_merger_mode
- [ ] test


normanrz:
Since #1174 is merged now, shall we close this?

PR #1233, opened 2024-12-27
Make `typing-extensions` a regular dependency

amotta:
Without this, installing `cluster_tools` from PyPI and then running `python slurm_example.py` results in the following error:

```
(connrecon) [mottales@vcl1062 connrecon]$ python slurm_example.py
  File "/tungstenfs/scratch/gfriedri/mottales/code/connrecon/slurm_example.py", line 5, in <module>
    import cluster_tools
  File "/tungstenfs/scratch/gfriedri/mottales/code/connrecon/.pixi/envs/default/lib/python3.11/site-packages/cluster_tools/__init__.py", line 4, in <module>
    from cluster_tools.executor_protocol import Executor
  File "/tungstenfs/scratch/gfriedri/mottales/code/connrecon/.pixi/envs/default/lib/python3.11/site-packages/cluster_tools/executor_protocol.py", line 14, in <module>
    from typing_extensions import ParamSpec
ModuleNotFoundError: No module named 'typing_extensions'
```

(In my opinion, it would be nicer if `typing_extensions` were, in fact, only a dev-dependency. But I don't know Python and its typing workflows well enough to implement this fix. So, this here is just the minimal-change fix.)

### DIFF
--- a/cluster_tools/pyproject.toml
+++ b/cluster_tools/pyproject.toml
@@ -6,6 +6,9 @@ authors = [{name= "scalable minds", email="hello@scalableminds.com"}]
 readme = "README.md"
 license = {text = "MIT"}
 requires-python = ">=3.9"
+dependencies = [
+    "typing-extensions ~=4.12.0",
+]
 
 [project.urls]
 Repository = "https://github.com/scalableminds/webknossos-libs"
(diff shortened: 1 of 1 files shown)


hotzenklotz (approved):

hotzenklotz:
@amotta Thanks for the PR. The changes make sense to me. `typing_extensions` is already a regular dependency for the `webknossos` package for exactly the same reasoning as described by you. Hence, we should do the same for the `cluster_tools`. 

Please also sign the contributor license agreement (CLA) so we can get this merged.

amotta:
Thanks, Tom! I could not sign the CLA last week (license agreement did not load), but have done so now. Thanks for making this library available as open source.

PR #1235, opened 2025-01-14
cluster_tools: In worker, use Python's defaults if `logging_config` is missing

amotta:
This commit changes `cluster_tools`, such that Python's default values are used in the absence of a user-specified `logging_config`. In particular, this commit "increases" the default logging level from `logging.DEBUG` to Python's default of `logging.WARNING`. Without this change, some packages (e.g., H5py) are flooding the log. Not only does this result in large log files (tens of MBs per file for some of my jobs), but it also slows down execution.


cluster_tools/cluster_tools/remote.py
```diff
@@ -115,10 +115,7 @@ def setup_logging(
         )
         meta_data["logging_setup_fn"](log_file_path)
     else:
-        logging_config = meta_data.get(
-            "logging_config",
-            {"level": logging.DEBUG, "format": "%(asctime)s %(levelname)s %(message)s"},
-        )
+        logging_config = meta_data.get("logging_config", dict())
```
hotzenklotz, line 118: Nitpick: Using an empty dict will also reset our default logging `format` to Python's default (`"%(levelname)s:%(name)s:%(message)s"` [Source Code](https://github.com/python/cpython/blob/1a84bdc2371ada60c01c72493caba62c9860007b/Lib/logging/__init__.py#L546)). I think that is acceptable.

hotzenklotz (approved):
Thanks for the contribution.

hotzenklotz:
FYI, this change should be released as `cluster_tools` version [v0.16.3](https://github.com/scalableminds/webknossos-libs/releases/tag/v0.16.3).

PR #1238, opened 2025-01-22
Add list_bounding_boxes() for zarr arrays

normanrz:
### Description:
- Add list_bounding_boxes() for zarr arrays, which improves speed for compress
- Use zstd compression per default for zarr arrays

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/_array.py
```diff
@@ -592,9 +594,57 @@ def write(self, bbox: NDBoundingBox, data: np.ndarray) -> None:
...
+            if chunk_coords_list is None:
+                continue
+
+            if shape.axes[0] == "c":
+                chunk_coords = Vec3Int(chunk_coords_list[1:])
+            else:
+                chunk_coords = Vec3Int(chunk_coords_list)
+
+            output.append(BoundingBox(chunk_coords * shard_shape, shard_shape))
+        return iter(output)
```
markbader, line 646: What are the advantages of first collecting all BoundingBoxes and only then making them into an Iterator object?
I think a `yield BoundingBox(...)` would be more readable.

markbader (approved):
LGTM :slightly_smiling_face:

webknossos/webknossos/dataset/_array.py
normanrz: It is a bit tricky, because then the NotImplementedError is only raised upon the first `next` call. I wrapped the function and addd a comment. That should work as well.

PR #1241, opened 2025-01-27
Fix trees don't behave like nx Graphs

### LINKED ISSUE #1232: Trees should behave like networkx graphs
## Context
Affected library: webknossos

## Expected Behavior
Trees should behave like networkx graphs

## Current Behavior
iterating over nodes with data=True is different.

## Steps to Reproduce the bug
```
s = wk.Skeleton((1, 1, 1), dataset_name="test_dataset")
g = s.add_group("test")
tree = g.add_tree("t")
tree.add_node((1, 2, 3))
for node_id, node_data in tree.nodes(data=True):
    print(node_id, node_data)
```
Should print the node  id and the data dict, but I get the Node and an empty dict.

## Your Environment for bug
webknossos-libs 0.16.2

markbader:
### Description:
- When using tree.nodes() the expectation when working with networkx Graphs is that a NodeView is returned, that maps node ids to their attributes. When calling it on a webknossos Tree, the nodes id stores the entire Node object and the attributes are empty. As Tree inherits from nx.Graph it should behave as a nx.Graph.

### Issues:
- fixes #1232 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


valentin-pinkau (approved):
Thank you! LGTM!

PR #1242, opened 2025-01-27
Make Zarr3 default and purge deprecations

### LINKED ISSUE #1080: Harden concurrency assertions
We want to switch our default data format to zarr3. In contrast to wkw, zarr3 doesn't allow concurrent writes (as uncompressed wkw does) at all. Therefore, we should be more careful when writing partial shards. Perhaps, we should elevate some of the warnings to assertions to better enforce shard-aligned writing.

normanrz:
### Description:
- Makes Zarr3 the default data format
- Removes deprecated functions, properties, arguments
- Changes the writing behavior to include a `allow_resize` and `allow_unaligned` argument to make it more robust against accidental concurrency issues
- MagViews are compressed by default

### Issues:
- fixes #1080

### Todos:
- [x] List all deprecations in the changelog
- [x] Add tests for new `allow_resize` behavior
- [x] Add `allow_unaligned` arg
- [x] Fix `from_images` for unexpected bboxes

 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests
 - [x] Updated examples


webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -264,7 +264,7 @@ def download_and_unpack(
 
 BIOFORMATS_ARGS: list[tuple[str, str, dict, str, int, tuple[int, int, int], int]] = [
     (
-        "https://samples.scif.io/wtembryo.zip",
+        "https://static.webknossos.org/data/wklibs-samples/wtembryo.zip",
```
normanrz, line 267: Using mirrors of the testdata to avoid dependence on external servers.

webknossos/webknossos/dataset/view.py
```diff
@@ -299,9 +241,8 @@ def _get_mag1_bbox(
     def write(
         self,
         data: np.ndarray,
-        offset: Optional[Vec3IntLike] = None,  # deprecated, relative, in current mag
-        json_update_allowed: bool = True,
```
philippotto, line 303: why doesn't the `View.write` method have a `allow_resize` parameter (only `MagView.write` has it currently)?

webknossos/webknossos/dataset/view.py
```diff
@@ -425,33 +347,55 @@ def write(
...
             for current_mag_bbox, chunked_data in self._prepare_compressed_write(
-                current_mag_bbox, data, json_update_allowed
+                current_mag_bbox, data
             ):
                 self._array.write(current_mag_bbox, chunked_data)
         else:
             self._array.write(current_mag_bbox, data)
 
+    def _check_shard_alignment(self, bbox: NDBoundingBox) -> None:
+        """Check that the bounding box is aling with the shard grid"""
```
philippotto, line 385:
```suggestion
        """Check that the bounding box is aligned with the shard grid"""
```

webknossos/webknossos/dataset/view.py
```diff
@@ -425,33 +347,55 @@ def write(
...
+        shard_shape = self.info.shard_shape
+        shard_bbox = bbox.align_with_mag(shard_shape, ceil=True)
+        if shard_bbox.intersected_with(self.bounding_box.in_mag(self._mag)) != bbox:
+            raise ValueError(
+                f"The bounding box to write {bbox} is not aligned with the shard shape {shard_shape}. "
+                + "Performance will be degraded as existing shard data has to be read, combined and "
```
philippotto, line 391: how about also mentioning the problems for parallel writes?

webknossos/webknossos/dataset/view.py
```diff
@@ -425,33 +347,55 @@ def write(
...
 
+        if not allow_unaligned:
+            if self._data_format == DataFormat.WKW and not self._is_compressed():
+                try:
+                    self._check_shard_alignment(current_mag_bbox)
+                except ValueError:
+                    shard_shape = self.info.shard_shape
+                    warnings.warn(
+                        f"[WARNING] The bounding box to write {current_mag_bbox} is not aligned with the shard shape {shard_shape}. "
+                        "This was supported for uncompressed WKW datasets, but is deprecated now because of concurrency issues. "
```
philippotto, line 370:
```suggestion
                        "This was supported for uncompressed WKW datasets, but is deprecated now because of issues with performance and concurrent writes. "
```

webknossos/webknossos/dataset/view.py
```diff
@@ -425,33 +347,55 @@ def write(
...
+                        "Please use an explicit `allow_unaligned` argument to fix this."
```
philippotto, line 371: "Either, ensure that you write shard-aligned chunks OR pass allow_unaligned=True. When using the latter, take care to not write concurrently."

philippotto (commented):
great stuff :+1: I especially like the removal of all the deprecations.

However, I think I'm not yet fully understanding when `allow_resize` can be specified and when not. see my one comment for that.

cluster_tools/tests/test_all.py
```diff
@@ -253,6 +256,9 @@ def test_unordered_sleep(exc: cluster_tools.Executor) -> None:
 
     with exc:
         durations = [5, 0]
+        # Slurm can be a bit slow to start up, so we need to increase the sleep time
+        if isinstance(exc, cluster_tools.SlurmExecutor):
+            durations = [20, 0]
```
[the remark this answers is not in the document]
normanrz: I changed this, because the CI was very flaky. I don't know why it only surfaced now. The thing is that this test tests the "unordered map", but asserts that the results are ordered in a certain way. I assume that slurm is a bit slow with scheduling and process startup, which leads to the flakiness; i.e. the 5 second sleep doesn't dominate the runtime of the job. Another solution would be to skip this test on slurm or only check for set-equality instead of ordered-equality.

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -48,21 +44,21 @@ def __init__(
...
+        allow_unaligned: bool = False,
     ) -> None:
         """see `View.get_buffered_slice_writer()`"""
 
         self._view = view
         self._buffer_size = buffer_size
         self._dtype = self._view.get_dtype()
         self._use_logging = use_logging
-        self._json_update_allowed = json_update_allowed
+        self._allow_unaligned = allow_unaligned
```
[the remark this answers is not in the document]
normanrz: `allow_resize` doesn't exist for a `View`, which the `BufferedSliceWriter` wraps, because it isn't connected to the `Layer`. So, implicitly, a `BufferedSliceWriter` is always `allow_resize=False`

webknossos/webknossos/dataset/view.py
normanrz: See my comment above. A `View` doesn't have a reference to its `Layer`, so it cannot resize the `Layer`.

cluster_tools/tests/test_all.py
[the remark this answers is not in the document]
philippotto: ah ok, gotcha :+1: then, let's go with the current solution.

> I don't know why it only surfaced now.

the durations used to be a bit higher but were changed ~2 months ago. maybe this explains it a bit.

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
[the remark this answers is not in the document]
philippotto: ok, interesting. I'm a bit surprised, because the old json update allowed parameter existed for both classes. but looking at the code, it seems like the `View` did not really respect that parameter (it only warned but the json was never updated).

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1900,20 +1886,11 @@ def add_layer_from_images(
...
-                warnings.filterwarnings(
-                    "ignore",
-                    message=_BLOCK_ALIGNMENT_WARNING,
-                    category=UserWarning,
-                    module="webknossos",
-                )
-                # There are race-conditions about setting the bbox of the layer.
-                # The bbox is set correctly afterwards, ignore errors here:
+                # For n-d datasets, the `axisOrder` property is stored with mags.
+                # At this point, we don't have any mags yet, so we can't compare the persisted properties.
```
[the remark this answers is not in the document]
normanrz: Does this make it clearer?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1984,23 +1961,88 @@ def add_layer_from_images(
...
+        category: LayerCategoryType,
+        data: np.ndarray,
+        *,
+        data_format: Union[str, DataFormat] = DEFAULT_DATA_FORMAT,
+        downsample: bool = True,
+        chunk_shape: Optional[Vec3IntLike] = None,
+        chunks_per_shard: Optional[Vec3IntLike] = None,
+        axes: Optional[Iterable[str]] = None,
+    ) -> Layer:
+        if axes is not None:
```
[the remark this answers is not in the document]
normanrz: I didn't see anything similar somewhere else, but extracted it into a `derive_nd_bounding_box_from_shape` function.

normanrz:
@markbader This PR is now ready for your review.

webknossos/tests/dataset/test_dataset.py
```diff
@@ -270,11 +269,11 @@ def test_ome_ngff_metadata(output_path: Path) -> None:
 
 
 def test_create_default_layer() -> None:
-    ds_path = prepare_dataset_path(DataFormat.WKW, TESTOUTPUT_DIR)
+    ds_path = prepare_dataset_path(DataFormat.Zarr3, TESTOUTPUT_DIR)
```
markbader, line 272: I would use the `DEFAULT_DATA_FORMAT` from `webknossos.dataset.defaults` for this test, to avoid changing this test when the default changes.

markbader (approved):
Looks great, thanks for cleaning up all the deprecated code and the great overview of changes in the changelog!

PR #1246, opened 2025-01-28
Raise error for Dataset.open() if dataset is not found.

### LINKED ISSUE #1245: Misleading error for Dataset.open() with none-existing paths
## Context
Trying to open a none-existing dataset leads to a misleading error message:

## Expected Behavior

An error like: No Dataset was found on the given path. Would be way more helpful.

## Current Behavior

Currently an AttributeError is raised with the message: 'Sentinel' object has no attribute 'factor'

## Steps to Reproduce the bug
```python
import webknossos as wk

wk.Dataset.open("bliblablub")
``` 

## Your Environment for bug

wklibs v16.3

markbader:
### Description:
- When Dataset.open() is used for none-existing path the current error message is misleading. This PR improves the raised error.

### Issues:
- fixes #1245 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/dataset.py
```diff
@@ -348,6 +348,10 @@ def __init__(
                 raise FileNotFoundError(
                     f"Cannot open read-only dataset, could not find data at {self.path}."
                 )
+            if voxel_size_with_unit is _UNSPECIFIED_SCALE_FROM_OPEN:
```
valentin-pinkau, line 351: this somehow encodes, that we come from Dataset.open and expect the dataset to be available?
Can we make this more explicit?

valentin-pinkau (commented):
Thanks for looking into this. Can we make the behavior clearer? Please have a look at my comment.

webknossos/webknossos/dataset/dataset.py
markbader: I could get the caller of the __init__ function with `caller = inspect.stack()[1].function` and do something like `if caller == "open":` but I am unsure if this solution would be much better.

valentin-pinkau: another option would be to check, if the dataset is available in the `open()` method?

markbader: Yes that shounds better, I changed it that way.

normanrz: I only see this conversation now. This change introduces another redundant IO operation, which I am not happy with.

PR #1248, opened 2025-01-30
fix FileNotFoundError

normanrz:
CI should be enough


markbader (approved):
Thanks for fixing this :tada:

PR #1252, opened 2025-02-03
Update requires python to <3.14

markbader:
### Description:
- Updates requires-python to >=3.9,<3.13 to include patches of version 3.13


hotzenklotz (approved):
LGTM

PR #1254, opened 2025-02-04
Add fix for issue in setuptools.

markbader:
### Description:
- The pypi release failed because of an issue with setuptools (https://github.com/pypa/setuptools/issues/4759). This PR adds the license-files entry with an empty list to its pyproject.toml file as sugested in this discussion: https://github.com/astral-sh/uv/issues/9513.


hotzenklotz (approved):
LGTM. Hopefully, that let's us publish to pypi successfully now.

PR #1256, opened 2025-02-21
Parse dataset id from downloaded nml

MichaelBuessemeyer:
### Description:
Prior to these changes, wklibs missed to parse the datasetId provided in an nml of an annotation. This leads to issue / the libs crashing when parsing annotations from a renamed dataset. 

To reproduce and confirm the fix run something like with an annotation of a renamed dataset.
```python
annotation_id = "http://localhost:9000/annotations/<some existing annotation>/#128,128,128,0,1.3"

with wk.webknossos_context(token="<token>", url="http://localhost:9000"):
    dataset = wk.Annotation.open_as_remote_dataset(annotation_id)
```
On the master this code should crash but not on this branch.

### Issues:
- fixes bug report: https://scm.slack.com/archives/C02H5T8Q08P/p1739953765086629

### Todos:
 - [x] Updated Changelog


hotzenklotz (approved):
Code changes look. 

Perhaps add a unit test to prevent this from happening again. Maybe something like:
```
test_dataset_id = "http://localhost:9000/datasets/<some existing dataset>"
test_annotation_id = "http://localhost:9000/annotations/<some existing annotation>/#128,128,128,0,1.3"

with wk.webknossos_context(token="<token>", url="http://localhost:9000"):
    dataset = wk.Dataset.open_remote(test_dataset_id)
    dataset_from_annotation = wk.Annotation.open_as_remote_dataset(test_annotation_id)

    assert dataset.id == dataset_from_annotation.id
```

MichaelBuessemeyer:
> assert dataset.id == dataset_from_annotation.id

Sadly its not that easy. A dataset does not have an `id` and `open_as_remote_dataset` returns a `Dataset` not a `RemoteDataset` which would at least have `_dataset_id`. :/

I tried to add a test now that creates its own dataset, uploads and downloads an annotation, renames the DS and then tries to access the dataset from the re-downloaded annotation.

Hopefully, I can get this working in the Ci :woozy_face:

webknossos/testdata/annotations/l4dense_motta_et_al_demo_v2__explorational__4a6356.zip
MichaelBuessemeyer, line 1: I added the `datasetId` property to the nml to check against that in the test `test_annotation_from_wkw_zip_file` test

webknossos/tests/test_annotation.py
```diff
@@ -287,3 +289,31 @@ def test_nml_with_volumes(nml_path: Path) -> None:
...
+    with wk.webknossos_context("http://localhost:9000", test_token):
+        url = annotation_from_file.upload()
+        annotation = wk.Annotation.download(url)
+    assert annotation.dataset_name == "name_to_replace"
+    assert len(list(annotation.skeleton.flattened_trees())) == 1
+
+    renameable_dataset.name = "replaced_name"
+    # Test whether the DS can still be accessed via the annotation after the renaming.
+    dataset_from_annotation = annotation.get_remote_annotation_dataset()
+    assert dataset_from_annotation._dataset_id == renameable_dataset._dataset_id
```
MichaelBuessemeyer, line 319: As I am currently struggling to get this test working. I removed this test so we can add it in a follow up pr to not block merging the fix

MichaelBuessemeyer: see https://github.com/scalableminds/webknossos-libs/issues/1258

PR #1264, opened 2025-03-04
Add total file size in dataset upload information.

markbader:
### Description:
- This PR adds the field totalFileSizeInBytes to the dataset upload information.

### Issues:
- https://github.com/scalableminds/webknossos/pull/8385#discussion_r1946261223

### Todos:
 - [x] Updated Changelog


MichaelBuessemeyer:
Thanks a lot for this PR @markbader :tada: 

The passed `total_file_size_in_bytes` seems to be correct, but while testing with branch `enforce-storage-quota-before-upload` of PR https://github.com/scalableminds/webknossos/pull/8385 I noticed that sent `chunkSize` of each upload request (a request that uploads a chunk of the file to be uploaded), always sends a static value regardless of the actual uploaded size. At least that is my strong guess here and is very likely why the upload does not work for me.

Here is the log from my local testing.
[upload.log](https://github.com/user-attachments/files/19320148/upload.log)
 It directly shows at the beginning, that file `colordollar/4-4-1/header.wkw`  is claimed to have `104857600` bytes although this is far too large.

Could you please take another look at this? To test this you can checkout the commit https://github.com/scalableminds/webknossos/pull/8385/commits/3caa167250eb830875c9e3cb0e85ff48dd4bccde in webknossos to get the same logging help as I got. If you need help, feel free to ask :smile_cat: 

When this is fixed, the PR should be ready to go :dancers:

MichaelBuessemeyer:
Note: Maybe (in case one can upload zip files with wklibs. I don't know about that) in the future this might need further adjustment as we might decide to measure uncompressed size when uploading. See: https://github.com/scalableminds/webknossos/pull/8385#discussion_r2001133788

MichaelBuessemeyer (changes requested):
Just ticking the review off to get it removed from my list. Please re-request again once its time :D

MichaelBuessemeyer (commented):
Awesome thanks for debugging this :tada:

As commented [here](https://github.com/scalableminds/webknossos/pull/8385#discussion_r2008040104), this should hopefully be correct now :tada:

MichaelBuessemeyer (approved):
Neat, my test just worked so this should be ready to go, once the WK pr is merged :tada:

PR #1266, opened 2025-03-10
Show pims import errors only once

markbader:
### Description:
- This PR moves the import of different pims_readers, to avoid warnings for every pims image reader

### Todos:
 - [ ] Updated Changelog


normanrz:
@markbader  Looks good. What is missing here?

markbader:
@normanrz The pims import errors were reduced but still printed twice. As the module is imported twice while converting images to wkw or zarr. I moved the generation of the warning to the from_images() method now. This will cover most cases but does not print the errors when users use add_layer_from_images(). 
Additionally, I added a check for a slurm environment, to avoid the wk-libs version check on slurm.

PR #1267, opened 2025-03-10
Fixes the array shape when using Dataset.from_images

normanrz:
### Description:
- Fixes the array shape when using Dataset.from_images

### Issues:
- See https://scm.slack.com/archives/C02H5T8Q08P/p1741590153117259

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/layer.py
```diff
@@ -360,12 +360,26 @@ def bounding_box(self, bbox: NDBoundingBox) -> None:
         """
         Updates the offset and size of the bounding box of this layer in the properties.
         """
+        self._set_bounding_box(bbox)
```
markbader, line 363: The bounding_box setter should probably not call the `_set_bounding_box` method with it's default `expand_only = True`. Either we change the default as @philippotto suggested or we call it with `expand_only = False`:

```suggestion
        self._set_bounding_box(bbox, expand_only=False)
```

normanrz: That is a good point. Maybe we should just keep both in sync. It only affects metadata. Chunks outside the bbox are not deleted.

webknossos/webknossos/dataset/_array.py
```diff
@@ -511,42 +510,35 @@ def read(self, bbox: NDBoundingBox) -> np.ndarray:
...
-                )
-                new_domain = tensorstore.IndexDomain(
-                    new_aligned_bbox.ndim + 1,
-                    shape=(self.info.num_channels,)
-                    + new_aligned_bbox.bottomright.to_tuple(),
-                    implicit_upper_bounds=tuple(
-                        True for _ in range(new_aligned_bbox.ndim + 1)
-                    ),
-                    labels=array.domain.labels,
-                )
```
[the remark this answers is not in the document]
normanrz: Yes. align_with_shards wasn't used anywhere, so it was True always.

markbader (approved):
LGTM :) I noticed that test_mulitple_multitiffs takes quite a lot of memory, my laptop closes VS Code when I try to run it. Perhaps we should take another look at it.

PR #1270, opened 2025-03-17
Pass wk token with x-auth-token

markbader:
### Description:
- webknossos allows auth tokens as query or as header. As queries are shown in logs, we want to avoid putting user tokens there. This PR uses the `x-auth-token` header in favor of `?token=...`  query when sending requests to the datastore.

### Issues
- fixes: #1273 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


markbader:
I won't merge this now, because I think it might be good to add these two routes that are used by voxelytics:
- /annotations/:annotation_typ/:annotation_id/edit
- /datastores

markbader:
@normanrz What's your opinion about this branch. I added a RemoteAnnotation and a Datastore class to give users access to methods of the Api Client. The RemoteAnnotation is currently not really useful, as it only allows users to get infos about annotations and set their name and description (and even setting the description does not work currently: https://github.com/scalableminds/webknossos/issues/8475). 

Should this be merged anyway or are there other features a RemoteAnnotation class or the Datastore class should support?

normanrz:
I think this is good for now. We can expand on the RemoteAnnotation later, if we need/want to.

PR #1272, opened 2025-03-19
Fix OME metadata

normanrz:
### Description:
- Use OME-Zarr 0.5 metadata for Zarr3 datasets
- Only write OME-Zarr metadata for layers that are child directories of the dataset (not symlinks)

### Issues:
- see https://scm.slack.com/archives/C02Q35Q28P5/p1741598485944569

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -66,7 +66,7 @@ from .defaults import (
     ZATTRS_FILE_NAME,
     ZGROUP_FILE_NAME,
 )
-from .ome_metadata import write_ome_0_4_metadata
+from .ome_metadata import write_ome_metadata
 from .remote_dataset_registry import RemoteDatasetRegistry
 from .remote_folder import RemoteFolder
 from .sampling_modes import SamplingModes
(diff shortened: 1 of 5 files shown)


webknossos/webknossos/dataset/dataset.py
```diff
@@ -2794,7 +2794,9 @@ def _export_as_json(self) -> None:
                 json.dump({"zarr_format": 3, "node_type": "group"}, outfile, indent=4)
 
         for layer in self.layers.values():
-            write_ome_0_4_metadata(self, layer)
+            # Only write out OME metadata if the layer is a child of the dataset
+            if layer.path.resolve().parent == self.path:
```
daniel-wer, line 2798: Could `self.path` also be a symlink in which case it should also be resolved for this comparison?

daniel-wer (approved):
Nice, thanks for fixing this! :+1: See my one comment please.

webknossos/webknossos/dataset/dataset.py
normanrz: Good point. I also added caching for the path resolution, because on remote systems that may be one request per layer on every save.

daniel-wer (approved):
Sweet, LGTM :+1:

PR #1274, opened 2025-03-24
Change default dimension_separator for Zarr2

normanrz:
### Description:
- Changes the default dimension separator for new Zarr2 arrays to `/`. 

### Issues:
- see https://forum.image.sc/t/webknossos-uses-as-a-dimension-separator/110373

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/utils.py
+++ b/webknossos/webknossos/utils.py
@@ -324,7 +324,11 @@ def rmtree(path: Path) -> None:
 
 def copytree(in_path: Path, out_path: Path) -> None:
     def _walk(path: Path, base_path: Path) -> Iterator[Tuple[Path, Tuple[str, ...]]]:
-        yield (path, tuple(p for p in path.parts if p not in base_path.parts))
+        # base_path.parts is a prefix of path.parts; strip it
+        assert len(path.parts) >= len(base_path.parts)
+        assert path.parts[: len(base_path.parts)] == base_path.parts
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/utils.py
```diff
@@ -324,7 +324,11 @@ def _walk(path: Path) -> Iterator[Path]:
 
 def copytree(in_path: Path, out_path: Path) -> None:
     def _walk(path: Path, base_path: Path) -> Iterator[Tuple[Path, Tuple[str, ...]]]:
-        yield (path, tuple(p for p in path.parts if p not in base_path.parts))
+        # base_path.parts is a prefix of path.parts; strip it
+        assert len(path.parts) >= len(base_path.parts)
+        assert path.parts[: len(base_path.parts)] == base_path.parts
+        yield (path, path.parts[len(base_path.parts) :])
+
```
markbader, line 331: Looks good. What was the reasoning behind using parts and adding them again? Was this due to the use of upath that we can't use a method like Path.relative_to()?

markbader (approved):
LGTM! I just had one question regarding the _walk() method.

webknossos/webknossos/utils.py
normanrz: In the calling `copytree` code, the parts need to be added to the `in_path` and `out_path`. Not sure the latter would work with relative_to.

PR #1275, opened 2025-03-25
fix dataset upload (use id instead of directory name)

markbader:
### Description:
- This PR fixes issue with uploading datasets where the dataset name is already in use.

### Todos:
 - [ ] Updated Changelog


webknossos/webknossos/dataset/dataset.py
```diff
@@ -1224,7 +1224,9 @@ def upload(
         )
 
         return self.open_remote(
-            upload_dataset(self, new_dataset_name, converted_layers_to_link, jobs)
+            dataset_id=upload_dataset(
```
hotzenklotz, line 1227: Perhaps, we move the `upload` call outside the the `open_remote` call. That should make the stacktrace much more readable in case of an exception - whether an error is related to open or upload.

hotzenklotz (approved):
LGTM.

Perhaps we can add another test that uploads the same DS twice.

webknossos/webknossos/dataset/dataset.py
markbader: Sure, I changed this.

PR #1278, opened 2025-03-25
Better S3 Support for MagViews

Tobias314:
### Description:
- Context is creating a "downsamped" version (dataset with original + downsampled mags) of an original dataset without modifying the original dataset but by creating a "shallow" copy instead
- The underlying problem is that I want to create a new dataset B referencing Mag1 from another (remote) S3 dataset A and then adding downsampled Mag2, Mag4, Mag8, ... to dataset B without modifying A
- Previous code had logic which inferred `layer.path` from the paths of the layer's mags, which leads to problems for layers with both remote and local mags
- Proposed solution defines `layer.path` to always be `dataset_path / layer.name`

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


Tobias314:
@normanrz @markbader I think we should discuss this PR. The proposed solution changes/redefines the behavior of `layer.path` which probably is a breaking change (also visible by failing tests). However, since `layer.path` is not specified in the datasource_properties.json, it is questionable whether there is a good way to define it for layers with both local and remote mags...

Maybe we should discuss what the expected behavior should be and how to create a new dataset layer with some (not all) of its mags pointing to mags of other (remote) datasets?

webknossos/webknossos/dataset/_array.py
```diff
@@ -431,8 +432,14 @@ def _make_kvstore(path: Path) -> str | dict[str, str | list[str]]:
...
-            if endpoint_url := path.storage_options.get("client_kwargs", {}).get(
-                "endpoint_url", None
+            if (
+                (
+                    endpoint_url := path.storage_options.get("client_kwargs", {}).get(
+                        "endpoint_url", None
+                    )
+                )
+                or (endpoint_url := path.storage_options.get("endpoint_url", None))
+                or (endpoint_url := os.environ.get("S3_ENDPOINT_URL", None))
```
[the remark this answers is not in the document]
normanrz: I think we should rely on the UPath

webknossos/webknossos/dataset/_array.py
```diff
@@ -780,6 +787,7 @@ def create(cls, path: Path, array_info: ArrayInfo) -> "Zarr3Array":
                     ],
                 },
                 "create": True,
+                "open": True,
```
normanrz, line 790: Hmm?

webknossos/webknossos/dataset/layer.py
```diff
@@ -226,40 +226,13 @@ def __init__(self, dataset: "Dataset", properties: LayerProperties) -> None:
...
-                mag.path is None and maybe_layer_path is None
-            ) or strip_trailing_slash(UPath(mag.path)).parent == maybe_layer_path
-            assert is_same_layer, "All mags of a layer must point to the same layer."
-        is_remote = maybe_layer_path and is_remote_path(maybe_layer_path)
-        return (
-            maybe_layer_path
-            if maybe_layer_path and is_remote
-            else self.dataset.path / self.name
-        )
+        return self.dataset.path / self.name
```
normanrz, line 235: Have you checked the usages of `Layer.path` in wklibs and vx?

webknossos/webknossos/dataset/layer.py
```diff
@@ -691,7 +664,15 @@ def add_existing_remote_mag_view(
             f"Cannot add mag {mag} as it already exists for layer {self}"
         )
         self._setup_mag(mag, mag_path)
-        self._properties.mags.append(mag_view._properties)
+        # since the remote mag view might belong to another dataset, it's property's path might be None, therefore, we get the path from the mag_view itself instead of it's properties
+        self._properties.mags.append(
+            MagViewProperties(
+                mag=mag_view.mag,
+                path=str(mag_view.path),
```
normanrz, line 671: Just casting `str(..)` probably doesn't work, because wk expects a different syntax for S3 paths, i.e. `s3://{endpoint_url/{bucket_name}/{prefix}` instead of `s3://{bucket_name}/{prefix}`

normanrz (commented):
In general, I am not opposed to this change. We just need to make sure it works in both wklibs and vx.

webknossos/webknossos/dataset/_array.py
```diff
@@ -1,3 +1,4 @@
+import os
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_array.py
normanrz:
```suggestion
            if endpoint_url := path.storage_options.get("client_kwargs", {}).get(
                "endpoint_url", None
```

normanrz:
Superceded by https://github.com/scalableminds/webknossos-libs/pull/1288

PR #1281, opened 2025-03-31
Explore test coverage

markbader:
### Description:
- This PR introduces coverage to check which code paths are not tested yet. It uses the docker image of the current webknossos.org instance for refreshing snapshots in the nighly tests.


.github/workflows/ci.yml
```diff
@@ -198,17 +198,66 @@ jobs:
           sed -i '/def _launch/a\ \ \ \ \ \ \ \ raise Exception("fork is not allowed.")' /home/runner/.local/share/uv/python/cpython-${{ matrix.python-version }}.*-linux-x86_64-gnu/lib/python${{ matrix.python-version }}/multiprocessing/popen_fork.py
           cat /home/runner/.local/share/uv/python/cpython-${{ matrix.python-version }}.*-linux-x86_64-gnu/lib/python${{ matrix.python-version }}/multiprocessing/popen_fork.py
 
+      - name: Python tests (with coverage)
+        if: ${{ matrix.python-version == '3.11' }}
+        timeout-minutes: 30
+        env:
+          PYTHON_VERSION: ${{ matrix.python-version }}
+          COVERAGE_FILE: "~/.local/share/coverage/${{ matrix.group }}.coverage"
```
hotzenklotz, line 206: What happens in the test case with coverage above if there is no coverage file to restore? Does it crash?

.github/workflows/ci.yml
```diff
@@ -198,17 +198,66 @@ jobs:
...
+
+      - name: Install coverage.py
+        run: pip install coverage
+
+      - name: Download all coverage reports
+        uses: actions/download-artifact@v4
+        with:
+          path: ~/coverage-files
+
+      - name: Debug List downloaded files
```
hotzenklotz, line 248: Can this be removed?

.github/workflows/ci.yml
```diff
@@ -198,17 +198,66 @@ jobs:
...
+        run: ls -R ~/coverage-files
+
+      - name: Generate coverage report
+        env:
+          COVERAGE_FILE: "~/coverage-files/result.coverage"
+        run: |
+          coverage combine ~/coverage-files/**/*.coverage
+          coverage report -m
```
hotzenklotz, line 256: Is this added to the Github Actions run somehow? Something like https://github.com/marketplace/actions/python-coverage or similar?

webknossos/local_wk_setup.sh
```diff
@@ -1,7 +1,11 @@
 function export_vars {
+    # Fetch current version of webknossos.org this can be replaced with a fixed version for testing
+    WK_VERSION=$(curl -s https://webknossos.org/api/buildinfo | tr ',"' "\n" | sed -n '/version/{n;n;p;q;}')
```
hotzenklotz, line 3: The `buildInfo` is valid JSON. Please use the CLI tool `jq` for parsing instead, e.g. `curl -s https://webknossos.org/api/buildinfo | jq -r '.webknossos.version'`

.github/workflows/nightly.yml
```diff
@@ -37,10 +37,6 @@ jobs:
         git diff --no-ext-diff --quiet --exit-code
         [[ -z $(git status -s) ]]
 
-    - name: Use currently deployed Webknossos version
```
hotzenklotz, line 40: Looking at this, I would assume that the current nightly already run against the latest wk.org version, right? So your changes to do not affect this behavior and just move the code around?

hotzenklotz (changes requested):

.github/workflows/ci.yml
markbader: It does not restore the coverage file, but uses the path given in `COVERAGE_FILE` for storing the partial coverage report.

.github/workflows/ci.yml
markbader: Yes, I will remove this.

webknossos/local_wk_setup.sh
markbader: Sure, I will change this, looks much better.

.github/workflows/nightly.yml
markbader: Yes, the nightly run already used the current wk.org version, but the lokal_wk_setup script has to be updated manually. Now the latest wk.org version is also used when updating test cassettes.

.github/workflows/ci.yml
hotzenklotz: I would love to see some nice statistics/summary as part of the GA run. Have a look at WK CI and the output of the Docker build step: https://github.com/scalableminds/webknossos/actions/runs/14442701567

markbader: I added your proposed python-coverage package to the Actions runner. Do you have any preferences on threshold values for the shown statistics?

hotzenklotz: No, not really. I guess the 80/20 rule could be applied here as well, so 80% coverage should be good enough. Maybe, we don't need a hard threshold.

hotzenklotz (approved):

.github/workflows/ci.yml
markbader: I add a threshold of 0.8, and we will see how it works.

PR #1283, opened 2025-04-03
Add require unique name param to announce manual upload

### LINKED ISSUE #1282: Adapt reserve manual upload
## Detailed Description
- Affected library:  webknossos

## Features to implement
- [ ] reserveManualUpload should return two values directoryName and datasetId
- [ ] reserveManualUpload should take another argument

## Use Cases & Context
changes for voxelytics

markbader:
### Description:
- This PR adds a bool param to annaounce manual upload to make it fail when the dataset name is already taken. Additionally the request is answered with the dataset_id and the directory_name, these are passed through to the convenience methods.

### Issues:
- fixes #1282

### Todos:
 - [x] Updated Changelog


webknossos/webknossos/client/api_client/datastore_api_client.py
```diff
@@ -72,10 +73,13 @@ def dataset_reserve_manual_upload(
         self,
         dataset_announce: ApiDatasetAnnounceUpload,
         token: str | None,
-    ) -> None:
+        require_unique_name: bool = False,
+    ) -> ApiDatasetManualUploadSuccess:
         route = "/datasets/reserveManualUpload"
-        query: Query = {"token": token}
-        self._post_json(route, dataset_announce, query)
+        query: Query = {"token": token, "requireUniqueName": require_unique_name}
```
MichaelBuessemeyer, line 79: Could you please make `require_unique_name` part of `ApiDatasetAnnounceUpload`? Otherwise the backend needs to extract the info not only from the post body but also from the query string. I would prefer to have all the info in the body (excluding the token which is needed as a query parameter) :pray:

MichaelBuessemeyer:
The corresponding wk pr is: https://github.com/scalableminds/webknossos/pull/8476

MichaelBuessemeyer (approved):
Awesome thanks for implementing this :dancers: 

I tested this with
```python
def main() -> None:
    with wk.webknossos_context(
            token="<local token>", url="http://localhost:9000"
    ):
        new_name = "new_name"
        print(wk.Dataset.announce_manual_upload(new_name, "sample_organization", [], "67ee5554780100780102eda3", False))
        try:
            print(wk.Dataset.announce_manual_upload(new_name, "sample_organization", [], "67ee5554780100780102eda3", True))
        except Exception as e:
            print("failure expected")
        try:
            print(wk.Dataset.announce_manual_upload(new_name, "sample_organization", [], "67ee5554780100780102eda3", False))
            print("should be printed")
        except Exception as e:
            print("should not be printed")
        print(wk.Dataset.announce_manual_upload("another_new_name", "sample_organization", [], "67ee5554780100780102eda3", True))
        print(wk.Dataset.announce_manual_upload("another_new_name2", "sample_organization", [], "67ee5554780100780102eda3", False))
```

And everything worked as expected :tada:

Please only merge together with the corresponding wk pr or afterwards

PR #1284, opened 2025-04-08
Fix some typos with codespell

fm3:
I’m unsure about the `Thsi error` line, was this intentional?


markbader (approved):
Looks good. I don't remember that there was an intention behind this error. Might be a debugging relict.

PR #1286, opened 2025-04-10
Fix get_remote_datasets

markbader:
### Description:
- This PR fixes an issue with get_remote_datasets. It fixes a call of remote_dataset that puts the dataset_id in the name_or_url argument. Additionally it adds a __repr__ method to the lazy loaded RemoteDatasetRegistry and gives the context it is initialized in to each remote dataset.. It also changes behavior of _parse_remote.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/dataset.py
```diff
@@ -773,19 +749,18 @@ def open_remote(
...
         with context_manager:
             wk_context = _get_context()
-            api_dataset_info = wk_context.api_client.dataset_info(
-                dataset_id, sharing_token
-            )
             token = sharing_token or wk_context.datastore_token
+            api_dataset_info = wk_context.api_client.dataset_info(dataset_id, token)
+            directory_name = api_dataset_info.directory_name
+            organization_id = api_dataset_info.owning_organization
+            datastore_url = api_dataset_info.data_store.url
```
MichaelBuessemeyer, line 756: Nice, this is much cleaner :tada:

MichaelBuessemeyer (commented):
Hej @markbader, thanks a lot for taking care of this fix.

I found one small remark. Please find it in the comments. Besides that the code looks good :+1: 

For testing: Do you have any steps / a script I can try to check whether the fix works? I could not find any linked issue or slack thread or such from which to derive testing steps.

Mroeover, I noticed that in `dataset.py` the regex `_DATASET_URL_REGEX` has a mistake: In it, it defines `directory_name` as a matching group but this is in fact a potentially outdated `dataset_name` the libs should only rely on the `dataset_id` in such a link. For datasets that were never renamed, the `directory_name` part is actually the dataset's directory name but once a dataset was renamed, this is either the new name or some outdated name. Thus, one cannot be rely on this information. I'd say it is best to rename the matching group to something like `dataset_name` or so and highlight in a comment, that this might be outdated data and should not be trusted. The up to date name can always be retrieved from the fetched dataset object from the server when requesting with the id :D.

Ok long story short: The code never seems to use the match group `directory_name` :+1: neat :100:

markbader:
Hey @MichaelBuessemeyer, thanks for your feedback. I adjusted the comment of the _parse_remote() method.

You are right, the directory_name was only included as a group to give this part of the regex a name. Would you suggest changing it to something like `dataset_name` or leave this regex part unnamed?

Regarding testing: there was no issue related to this. I just found the bug when trying to access a remote dataset, because the disambiguation currently uses get_remote_datasets. Before this PR the values of the dict returned by get_remote_datasets were invalid, as they were calls of `Dataset.open_remote(<dataset_id>)` instead of `Dataset.open_remote(dataset_id=<dataset_id>)`. So the test I wrote would fail on the previous version because accessing the values of the lazy remote dataset dicts did not work.

MichaelBuessemeyer (approved):
> Ok I'll give the get_remote_datasets a quick check 👍

Nice looks good. Can confirm that the test works :tada:

PR #1287, opened 2025-04-14
Fix upsampling

markbader:
### Description:
- This PR fixes an issue with the upsample cube job.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


daniel-wer (approved):
Thank you for looking into this and cleaning up the code while you were at it! The upsample command that triggered the report works perfectly for me now :)

PR #1290, opened 2025-04-17
Always store path in mags

normanrz:
### Description:
- All mags now store their path in the `datasource-properties.json` file. This eliminates the need for symlinks and removes the need to check multiple possible mag paths (e.g. 1-1-1 or 1). It also unifies local and remote mags.
- `Dataset.add_symlink_layer`, `Layer.add_symlink_mag` and `Attachments.add_symlink_attachments` are deprecated and will be removed in a future release. Please use `Dataset.add_layer_as_ref`, `Layer.add_mag_as_ref` and `Attachments.add_attachment_as_ref` instead.
- Renamed methods `add_{copy,remote}_{layer,mag}` to `add_{layer,mag}_as_{copy,ref}`.
- Replaced `add_{copy_,}attachments` with `add_attachment_as_{copy,ref}` (note the singular form).
- `Dataset.shallow_copy_dataset` does not symlink layers, mags and attachments anymore. Instead mags and attachments are referenced by their path.
- Dataset-local mags and attachments use the `./` prefix for their path in compliance with [IETF RFC1808](https://datatracker.ietf.org/doc/html/rfc1808)

Followup to #1288

### Todos:
- [x] Updated Changelog
 - [x] Added / Updated Tests


philippotto:
> I wonder if we should unify this to `add_copy_*` and `add_ref_*`? cc @valentin-pinkau @philippotto

yes, `add_remote` does not make much sense when it is also used for non-remote stuff. additionally, I think, that the method names could be clearer. for example: `add_{mag,...}_as_{copy,ref}` or `add_{mag,...}_{copy,reference}`.

normanrz:
Hab alles zu `add_{mag,layer,attachment}_as_{copy,ref}` umbenannt.

markbader (commented):
Looks good, thanks for bringing this forward. I personally don't like the use of the deprecated `add_symlink_layer()` method in the tests. If the user is not encouraged to use these methods, we should rather adapt the test. Maybe you could adapt the test cases where the deprecated methods are used, to fit to our new recommended methods.

normanrz:
Waiting to merge until https://github.com/scalableminds/voxelytics/pull/4157 is ready.

PR #1297, opened 2025-04-30
switch to ported slurm docker cluster

robert-oleynik:
### Description:
- The old Docker Slurm Cluster image (see: https://github.com/scalableminds/dockerfiles/tree/master/slurm-docker-cluster) sporadically fails to build. Because of that we switch to the latest Debian version and our custom Slurm packages.

### Issues:
- fixes CI build failures for Docker Slurm Cluster image

### DIFF
--- a/cluster_tools/dockered-slurm/docker-compose.yml
+++ b/cluster_tools/dockered-slurm/docker-compose.yml
@@ -13,7 +13,7 @@ services:
       - ..:/cluster_tools
 
   slurmdbd:
-    image: scalableminds/slurm-docker-cluster:master__11274637426
+    image: scalableminds/slurm-docker-cluster:port_slurm_in_docker_to_debian__14751549755
     command: [ "slurmdbd" ]
     container_name: slurmdbd
     hostname: slurmdbd
@@ -29,7 +29,7 @@ services:
       - mysql
 
   slurmctld:
-    image: scalableminds/slurm-docker-cluster:master__11274637426
+    image: scalableminds/slurm-docker-cluster:port_slurm_in_docker_to_debian__14751549755
     command: [ "slurmctld" ]
     container_name: slurmctld
     environment:
@@ -50,7 +50,7 @@ services:
       - "slurmdbd"
 
   c1:
-    image: scalableminds/slurm-docker-cluster:master__11274637426
+    image: scalableminds/slurm-docker-cluster:port_slurm_in_docker_to_debian__14751549755
     command: [ "slurmd" ]
     hostname: c1
     container_name: c1
@@ -68,7 +68,7 @@ services:
       - "slurmctld"
 
   c2:
-    image: scalableminds/slurm-docker-cluster:master__11274637426
+    image: scalableminds/slurm-docker-cluster:port_slurm_in_docker_to_debian__14751549755
     command: [ "slurmd" ]
     hostname: c2
     container_name: c2


cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -433,18 +435,28 @@ def parse_key_value_pairs(
...
+                continue
+
+            # Parse stdout into a key-value object
+            memory_limit_investigation = self._investigate_memory_consumption(stdout)
+            if memory_limit_investigation:
+                return memory_limit_investigation
 
-        # Parse stdout into a key-value object
-        properties = parse_key_value_pairs(stdout, "\n", ":")
+            break
```
philippotto, line 456: 1. this doesn't do retries if sacct fails? would it do any harm to do so?
2. I find the trailing break in this for-loop unintuitive to read. my suggestion would be this:

```python
        max_retries = 10
        for _ in range(max_retries):
            stdout, _, exit_code = call(
                f"sacct -P --format=JobID,State,MaxRSS,ReqMem --unit K -j {job_id_with_index}"
            )

            if exit_code != 0 or len(stdout.splitlines()) <= 1:
                # Sleep and retry in the next loop
                time.sleep(0.2)
            else:
                # Parse stdout into a key-value object
                memory_limit_investigation = self._investigate_memory_consumption(stdout)
                if memory_limit_investigation:
                    return memory_limit_investigation
                break
```

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -466,30 +478,33 @@ def _investigate_time_limit(
...
-        if match is None:
-            return None
-
-        try:
-            percentage = float(match.group(1))
-        except ValueError:
-            return None
-
-        if percentage < 100:
+        stdout_lines = stdout.splitlines()
```
philippotto, line 483: could you add a comment that mentions the expected format for the stdout input? this makes it easier to follow the parsing logic.

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -466,30 +478,33 @@ def _investigate_time_limit(
...
+        max_rss = 0
+        req_mem = 0
+        if len(stdout_lines) > 0:
+            for line in stdout_lines[1:]:
+                params = line.split("|")
+                try:
+                    max_rss = max(max_rss, int(params[2][:-1]))
```
philippotto, line 490: `[:-1]` strips away the unit which is always expected to be one character long? maybe a regex for numbers would make this more robust? :thinking:

cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -466,30 +478,33 @@ def _investigate_time_limit(
...
+                except Exception:
+                    pass
+                try:
+                    req_mem = max(max_rss, int(params[3][:-1]))
+                except Exception:
+                    pass
```
philippotto, line 496: in which cases are these exceptions expected? i.e., what kind of stdouts lines make this necessary? maybe we can guard against these a bit more specific than a generic try-catch.

philippotto (changes requested):
great stuff :) I only left a couple of smaller comments.

.github/workflows/ci.yml
```diff
@@ -29,121 +29,168 @@ jobs:
...
               - 'cluster_tools/**'
 
-  cluster_tools:
+  cluster_tools_slurm:
     needs: changes
     if: ${{ needs.changes.outputs.cluster_tools == 'true' }}
     runs-on: ubuntu-latest
     timeout-minutes: 30
     strategy:
-      max-parallel: 4
```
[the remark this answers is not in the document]
robert-oleynik: It is not easy to set across multi matrix runs and GitHub can execute all of them in parallel anyway. Restricting the number of jobs to 4 will only increase the time until all jobs are complete.

PR #1298, opened 2025-05-05
Fix `_dump_mag_path` with paths of different file systems 

Tobias314:
### Description:
- only check for relative paths on same file system because otherwise upathlib fails

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/layer.py
+++ b/webknossos/webknossos/dataset/layer.py
@@ -53,6 +53,7 @@ from ..utils import (
     named_partial,
     resolve_if_fs_path,
     rmtree,
+    safe_is_relative_to,
     warn_deprecated,
 )
 from .defaults import (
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/dataset/layer.py
```diff
@@ -230,7 +230,10 @@ def _dump_mag_path(path: Path, dataset_path: Path) -> str:
     path = resolve_if_fs_path(path)
     if str(path).startswith(str(dataset_path)):
         return str(path).removeprefix(str(dataset_path)).lstrip("/")
-    if path.is_relative_to(dataset_path):
+    if (
```
normanrz, line 233: Can you extract that into a `safe_is_relative_to` function in `webknossos/utils.py`?

normanrz (approved):

PR #1303, opened 2025-05-13
Shallow copy for RemoteDatasets

markbader:
### Description:
- Removes an assertion from shallow copy that disables shallow copy of remote datasets.

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -2597,9 +2597,6 @@ class Dataset:
             copies of remote datasets or create shallow copies in remote locations.
         """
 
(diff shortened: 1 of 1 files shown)


Tobias314:
Sorry for jumping in here.
Can we also add full support for remote datasets in this PR?
I think the biggest challenge will be the call to `copy_directory_with_symlinks` since we cannot have symlinks with remote (e.g. S3) datasets. Not sure about the best solution in that case but simply copying files that cannot be referenced or symlinked might be a good fallback.

PR #1309, opened 2025-05-20
Follow up, correction of missing webknossos in usage

markbader:
### Description:
- In some usage fields the  `webknossos` command at the beginning was missing.


hotzenklotz (approved):
LGTM

PR #1310, opened 2025-05-20
Add support for in-memory datasets (e.g. useful for testing)

Tobias314:
### Description:
- add support for tensorstore in-memory driver to support in-memory datasets for zarr3 format
- add tensorstore context to allow re-opening of in-memory datasets
- add simple test to test a fully remote dataset by using using the in-memory file system as "remote" file system

### Todos:
 - [ ] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/_array.py
+++ b/webknossos/webknossos/dataset/_array.py
@@ -22,6 +22,8 @@ from ..geometry import BoundingBox, NDBoundingBox, Vec3Int, VecInt
 from ..utils import is_fs_path
 from .data_format import DataFormat
 
+TS_CONTEXT = tensorstore.Context()
+
 
 def _is_power_of_two(num: int) -> bool:
     return num & (num - 1) == 0
@@ -442,6 +444,14 @@ class TensorStoreArray(BaseArray):
             else:
                 kvstore_spec["aws_credentials"] = {"type": "default"}
             return kvstore_spec
+        elif isinstance(path, UPath) and path.protocol == "memory":
+            # use memory driver (in-memory file systems), e.g. useful for testing
+            # attention: this is not a persistent storage and it does not support
+            # multiprocessing since memory is not shared between processes
+            return {
+                "driver": "memory",
+                "path": path.path,
+            }
         else:
             return {
                 "driver": "file",
@@ -470,9 +480,13 @@ class TensorStoreArray(BaseArray):
                 },
                 open=True,
                 create=False,
+                context=TS_CONTEXT,
             ).result()  # check that everything exists
             return cls(path, _array)
         except Exception as exc:
+            import pdb
+
+            pdb.set_trace()
             raise ArrayException(f"Could not open array at {uri}.") from exc
 
     def read(self, bbox: NDBoundingBox) -> np.ndarray:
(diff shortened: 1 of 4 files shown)


webknossos/webknossos/dataset/_array.py
```diff
@@ -22,6 +22,8 @@
 from ..utils import is_fs_path
 from .data_format import DataFormat
 
+TS_CONTEXT = tensorstore.Context()
```
Tobias314, line 25: I needed to add a context, to allow re-opening the in-memory store. However, having a context should not hurt for non-in-memory datasets either, I hope.

webknossos/webknossos/dataset/_array.py
```diff
@@ -470,9 +480,13 @@ def _open(cls, path: Path) -> Self:
                 },
                 open=True,
                 create=False,
+                context=TS_CONTEXT,
             ).result()  # check that everything exists
             return cls(path, _array)
         except Exception as exc:
+            import pdb
+
+            pdb.set_trace()
```
markbader, line 489: This is probably an artifact from debugging. 
```suggestion
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -358,10 +358,9 @@ def __init__(
                 dataset_path_is_empty = next(self.path.iterdir(), None) is None
                 dataset_path_exists = True
             except NotADirectoryError:
-                dataset_path_exists = True
+                dataset_path_is_empty = True
```
markbader, line 361: Why did you change this line? You change causes that the Runtime error is not raised in the case of a NotADirectoryError.

markbader (commented):
Looks good to me, I just made two small remarks.

webknossos/webknossos/dataset/dataset.py
Tobias314: I ran into this error when creating a Dataset with a non-existing folder. I thought, the expected behavior is that the dataset is created. So we try to iterate the directory. Since the directory does not exist, we get an "NotADirectoryError". This tells us that the dataset path does not exist (`dataset_path_exists=False`) and since a non-existing directory cannot contain any files we also know that there are no conflicting files in the dataset directory (`dataset_path_is_empty=True`). So I thought the correct behavior ist to not throw an error and create the directory. Is that not what we want?

Tobias314: Ok, I tried to make the checks more explicit. I am not sure whether this is exactly what we had before but the way I see it we want to assert the following:
- either the given path does not exist yet (all fine, we will just create it as new folder)
- the given path exist but must be an empty directory.

In any other case (path exists but not a directory, directory exists but not empty) we throw a speaking error.

Does this make sense?

markbader: I think the expected error of a non-existing folder would be a FileNotFoundError. The NotADirectoryError should be raised when a directory-operation is called on something that is not a directory, e.g. a file. So there might be a file that is called exactly like the directory you are about to create. If the NotADirectoryError is also raised for non-existing folders, I agree with your change, but I think the intention of this line was to avoid overwriting files.

markbader (approved):
LGTM :)

PR #1313, opened 2025-05-23
Make folder id parameter of announce upload route optional according to beackend route

MichaelBuessemeyer:
### Description:

@MatthisCl noticed that the folderId parameter of the announce upload function is required although it is optional. This PR aims to make this parameter optional.

See https://github.com/scalableminds/webknossos/blob/da0d52d80a69427cc1b840723eff96f22904b6e7/webknossos-datastore/app/com/scalableminds/webknossos/datastore/services/uploading/UploadService.scala#L51 to compare with the backend route properties. There folderId is listed as optional.

### Issues:
- No issue exists for this

### Todos:
 - [ ] Updated Changelog
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/client/api_client/models.py
+++ b/webknossos/webknossos/client/api_client/models.py
@@ -131,7 +131,7 @@ class ApiDatasetAnnounceUpload:
     dataset_name: str
     organization: str
     initial_team_ids: list[str]
-    folder_id: str
+    folder_id: str | None
     require_unique_name: bool
 
 
--- a/webknossos/webknossos/dataset/dataset.py
+++ b/webknossos/webknossos/dataset/dataset.py
@@ -465,7 +465,7 @@ class Dataset:
         dataset_name: str,
         organization: str,
         initial_team_ids: list[str],
-        folder_id: str | RemoteFolder,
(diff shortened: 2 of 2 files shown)


MichaelBuessemeyer:
@MatthisCl FYI

MichaelBuessemeyer:
@markbader should I add tests for this?

markbader:
> should I add tests for this? I cannot find any existing ones for this route
I think testing this is a bit complicated, as webknossos is run in a docker container for testing. If you know an easy way to manually upload a dataset to docker, within the python tests, you could add a test. But this is not urgent, so we might do that in a follow-up PR.

MichaelBuessemeyer:
> I think testing this is a bit complicated, as webknossos is run in a docker container for testing. If you know an easy way to manually upload a dataset to docker, within the python tests, you could add a test.

Ah dang, right that's rather complicated. What we could do is just trigger the `announce_manual_upload` call and then load the dataset from the backend. It should be in an unsable state but sending the request without any additional parameters should work. But not sure how urgent this is.

> But this is not urgent, so we might do that in a follow-up PR.

Sounds good :+1:

PR #1316, opened 2025-06-02
Raise ValueError for unsupported dtypes.

### LINKED ISSUE #894: Check for supported dtypes when adding new mag
## Detailed Description

When adding a new mag / adding data to a layer, we should assert that the data's dtype is actually supported by WEBKNOSSOS and can be rendered. If not, provide a warning.

Supported dtypes can be found: https://docs.webknossos.org/webknossos/data_formats.html#wkw-datasets or at https://github.com/scalableminds/webknossos/issues/6581

## Use Cases & Context

This could prevent user errors, e.g. uploading color layer data as `uint64`.

markbader:
### Description:
- Adds a check to ensure, that the used dtypes are supported by webknossos.

### Issues:
- fixes #894 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -20,6 +20,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - When adding a layer, the used dtype is only valid if it is supported by webknossos. [#1316](https://github.com/scalableminds/webknossos-libs/pull/1316)
 
 ### Fixed
+- Fixed an issue that creates an Error when `add_mag()` was called just with a `chunk_shape`. [#1315](https://github.com/scalableminds/webknossos-libs/pull/1315)
 
 
 ## [2.3.8](https://github.com/scalableminds/webknossos-libs/releases/tag/v2.3.8) - 2025-05-28
--- a/webknossos/tests/dataset/test_layer.py
+++ b/webknossos/tests/dataset/test_layer.py
@@ -4,6 +4,7 @@ import numpy as np
 import tensorstore
 
 import webknossos as wk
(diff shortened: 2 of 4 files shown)


webknossos/tests/dataset/test_dataset.py
```diff
@@ -1554,41 +1555,6 @@ def test_adding_layer_with_valid_dtype_per_layer() -> None:
...
-            num_channels=1,
-            largest_segment_id=100000,
-        )
-        ds.add_layer(
-            "seg5",
-            SEGMENTATION_CATEGORY,
-            dtype_per_channel="float",
-            num_channels=3,
-            largest_segment_id=100000,
-        )
```
markbader, line 1591: I removed these layers from the test, as float and double are not supported for segmentation layers. Is there any reason this was tested before? @normanrz @philippotto

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1299,6 +1299,37 @@ def add_layer(
         else:
             dtype_per_channel = np.dtype("uint" + str(DEFAULT_BIT_DEPTH))
 
+        # assert that the dtype_per_channel are supported by webknossos
```
philippotto, line 1302:
```suggestion
        # assert that the dtype_per_channel is supported by webknossos
```

philippotto (approved):
lgtm :+1:

webknossos/tests/dataset/test_dataset.py
philippotto: I don't think there was a good reason for it.

PR #1317, opened 2025-06-04
Make slurm job canceling more robust

daniel-wer:
### Description:
- From time to time some slurm jobs keep running after killing voxelytics (and therefore invoking `inner_handle_kill` in `slurm.py`)
- I suspect this is either due to a race condition where jobs are in between pending and running when the function is called or because there is a moment during startup where the job doesn't react to the SIGINT signal
- To circumvent this, `scancel` is invoked again after a short grace period which definitely kills all given job ids, regardless of state

### Issues:
- fixes sporadic leftover slurm jobs

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/cluster_tools/cluster_tools/schedulers/slurm.py
+++ b/cluster_tools/cluster_tools/schedulers/slurm.py
@@ -262,8 +262,11 @@ class SlurmExecutor(ClusterExecutor):
             job_id_string = " ".join(unique_job_ids)
             # Send SIGINT signal to running jobs instead of terminating the jobs right away. This way, the jobs can
             # react to the signal, safely shutdown and signal (cancel) jobs they possibly scheduled, recursively.
+            # After a short waiting time kill all jobs that are still running (due to race conditions or because they
+            # didn't react to the SIGINT signal for some reason).
             _, stderr, _ = call(
-                f"scancel --state=PENDING {job_id_string}; scancel -s SIGINT --state=RUNNING {job_id_string}; scancel --state=SUSPENDED {job_id_string}"
+                f"scancel --state=PENDING {job_id_string}; scancel -s SIGINT --state=RUNNING {job_id_string};"
+                + "scancel --state=SUSPENDED {job_id_string}; sleep 2; scancel {job_id_string}"
             )
 
             maybe_error_or_warning = (


robert-oleynik (approved):
LGTM.

Note: Checking for jobs to kill may be cleaner, but this is a lot simpler (and should not return any errors)

daniel-wer:
I dove into this and fixed [#4035](https://github.com/scalableminds/voxelytics/issues/4035) while I was at it.

Turns out, when instantiating multiple cluster executors in the same process, the SIGINT/SIGTERM signal handler was overwritten multiple times which often caused the original signal handler of the main voxelytics process to no longer be called (i.e. if one of the executors no longer was referenced). To fix this and clean up the signal handling, I changed the signal handler overwriting/chaining to be implemented at a class level so it will only happen once and also won't be affected if previous executors are no longer referenced.

Regarding the sleep duration, I learned that it doesn't have to be <5s since the main voxelytics signal handler and the cluster executor signal handler are _chained_ and run one after the other (cluster executor first). So I opted to use the same timeout length that voxelytics is using (SIGTERM_WAIT_IN_S env variable - 5s by default).

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -38,18 +38,6 @@
 _S = TypeVar("_S")
 
 
-def _handle_kill_through_weakref(
```
philippotto, line 41: I think the same function still exists in `dask.py`. is this alright?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -158,26 +141,68 @@ def as_completed(cls, futs: list[Future[_T]]) -> Iterator[Future[_T]]:
...
+        try:
+            for hook in cls._shutdown_hooks:
+                hook()
+        except Exception as e:
+            print(f"Error during shutdown: {e}")
+
+        if callable(existing_signal_handler) and existing_signal_handler not in (
+            signal.SIG_DFL,
+            signal.SIG_IGN,
+            signal.default_int_handler,
```
philippotto, line 190: can you add a comment that explains this condition?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -158,26 +141,68 @@ def as_completed(cls, futs: list[Future[_T]]) -> Iterator[Future[_T]]:
...
+        signal.signal(
+            signal.SIGTERM,
+            partial(cls._handle_shutdown, existing_sigterm_handler),
+        )
+
+        cls._installed_signal_handler = True
+
+    @classmethod
+    def _register_shutdown_hook(cls, hook: Callable[[], None]) -> None:
+        cls._shutdown_hooks.append(hook)
```
philippotto, line 168: `_register_shutdown_hook` is only called in `__init__`. why is it a list then? couldnt it be an `Optional[Callable]` and `_installed_signal_handler` could also be absorbed, because either the hook was attached (is not none) or not.
additionally, I think that one could merge `_register_shutdown_hook` and `_install_signal_handlers` into one function, because one simply delegates to the other (plus appending to a list). I feel like this would be a bit easier to understand then.

philippotto (commented):
code looks good overall :+1: however, I still have some questions and also didn't test yet. maybe we can have a quick call (then you also don't need to reply to my questions here)?

> Turns out, when instantiating multiple cluster executors in the same process, the SIGINT/SIGTERM signal handler was overwritten multiple times which often caused the original signal handler of the main voxelytics process to no longer be called (i.e. if one of the executors no longer was referenced).

is this case tested? I don't see a spec which instantiates multiple executors 🤔 

> To fix this and clean up the signal handling, I changed the signal handler overwriting/chaining to be implemented at a class level so it will only happen once and also won't be affected if previous executors are no longer referenced.

I don't really understand this 🙈 questions for a call: why is a weakref no longer necessary? how is it ensured that it only happens once?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -130,15 +121,7 @@ def __init__(
...
-        # Clean up if a SIGINT signal is received. However, do not interfere with the
-        # existing signal handler of the process or the
-        # shutdown of the main process which sends SIGTERM signals to terminate all
-        # child processes.
-        existing_sigint_handler = signal.getsignal(signal.SIGINT)
-        signal.signal(
-            signal.SIGINT,
-            partial(_handle_kill_through_weakref, ref(self), existing_sigint_handler),
-        )
+        self._register_shutdown_hook(self.handle_kill)
```
daniel-wer, line 124: Deregister in shutdown and test that afterwards this list was cleaned up

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -158,26 +141,68 @@ def as_completed(cls, futs: list[Future[_T]]) -> Iterator[Future[_T]]:
...
+            signal.SIGTERM,
+            partial(cls._handle_shutdown, existing_sigterm_handler),
+        )
+
+        cls._installed_signal_handler = True
+
+    @classmethod
+    def _register_shutdown_hook(cls, hook: Callable[[], None]) -> None:
+        cls._shutdown_hooks.append(hook)
+        cls._install_signal_handlers()
```
daniel-wer, line 169: Rename to ensure_signal_handler_is_installed or so

cluster_tools/cluster_tools/schedulers/cluster_executor.py
daniel-wer: @normanrz Do we still want to maintain the dask executor?

If so, it needs to be adapted as well, yes.

normanrz: Nope. The dask executor can go.

cluster_tools/pyproject.toml
```diff
@@ -17,15 +17,16 @@ Changelog = "https://github.com/scalableminds/webknossos-libs/blob/master/cluste
 
 
 [project.optional-dependencies]
-kubernetes = ["distributed ~=2023.9.1"]
-dask = ["kubernetes ~=27.2.0"]
+dask = ["distributed ~=2023.9.1"]
+kubernetes = ["kubernetes ~=27.2.0"]
```
daniel-wer, line 21: The dependencies were switched by mistake :shrug:

cluster_tools/cluster_tools/schedulers/cluster_executor.py
daniel-wer: We decided not to remove the dask executor at this point to avoid a breaking change. Not fixing this bug for it is not a big deal. We should deprecate the dask and kubernetes executors in a followup PR.

cluster_tools/tests/test_slurm.py
```diff
@@ -168,85 +169,149 @@ def wait_until_first_job_was_submitted(
...
+        original_sigint_handler_was_called = True
+
+    signal.signal(
+        signal.SIGINT,
+        original_sigint_handler,
+    )
+
+    assert len(cluster_tools.SlurmExecutor._shutdown_hooks) == 0
+    # Reset class variable after changing signal handler so it will be set again when the next cluster executor is instantiated
+    cluster_tools.SlurmExecutor._installed_signal_handler = False
```
philippotto, line 269: won't this lead to the signal handler being set up twice? how about a class method à la `uninstall_signal_handler` which would do the uninstall + set the variable to false? however, it might be overkill if it's only necessary for this test.

philippotto (commented):
code looks good :+1: thank you so much for spending your time to get this right!
I didn't approve though, because you mentioned long delays during testing..

cluster_tools/tests/test_slurm.py
daniel-wer: In lines 262-264 we "reset" the signal handler, so the cluster executor signal handler that was installed by one of the previous tests is no longer installed. Therefore, we also need to make sure the `_installed_signal_handler` class variable is set to False again.
It's a special case for this test, because we pretend that a signal handler is already installed before the cluster executor is instantiated (to mimic what happens if voxelytics instantiates a cluster executor) and check whether it was actually called, then.

PR #1329, opened 2025-07-01
fix windows paths, maybe

normanrz:
### Description:
- abc

### Issues:
- fixes #...

### Todos:
 - [ ] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


markbader (approved):
Changes look plausible to me, but I couldn't test them on windows yet.
The uv version in ci.yml for cluster_tools and in publish_docs.yml is still 0.6.3 I think we should update it there too.

PR #1336, opened 2025-07-10
Allow supplying datastore_url in announce_manual_upload

fm3:
Manual uploads do not depend on the datastore having allowsUpload=true.

The users should be able to select the datastore for manual uploads.
 - [ ] Updated Changelog


valentin-pinkau (approved):
Awesome! LGMT!

PR #1340, opened 2025-07-22
Create volume annotations

brokkoli71:
### Description:
- Volume Annotations can be created and modified

### Issues:
- fixes #1318

### Todos:
- [x] How to test uploading volume annotation to wk?
- [x] Memory Store
- [x] is the created volume annotation zip correct?
- [x] fix slow tests

### Todos Documentation:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests

PR #1343, opened 2025-07-25
Configurable Zarr3 codecs

normanrz:
### Description:
Adds options to configure the Zarr3 arrays that are created by wklibs. This is required for generating volume annotations, see also #1340 

- Added `zarr3_codecs` and `zarr3_chunk_key_encoding` parameters to `MagView.add_mag` to allow for configuring Zarr3 datasets.
- The `sharding_indexed` codec is no longer used if `chunk_shape == shard_shape`.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


normanrz:
Adding `zarr3_codecs` and `zarr3_chunk_key_encoding` to all these functions is quite verbose. I wonder if we should overload the `compress` arg to allow for customization instead?

```python
class Zarr3Config:
    codecs: Sequence[Zarr3Codec] | None = None
    chunk_key_encoding: Zarr3ChunkKeyEncoding | None = None

def add_mag(...,
    compress: bool | Zarr3Config = True,
    ...
) -> MagView:
    ...

PR #1346, opened 2025-08-01
Fix bug in dump_path

normanrz:
### Description:
Adding refs layers to a dataset that has a prefix path of the source datasets didn't work properly. 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/utils.py
+++ b/webknossos/webknossos/utils.py
@@ -308,7 +308,7 @@ def is_remote_path(path: UPath) -> bool:
     return not is_fs_path(path)
 
 
-def resolve_if_fs_path(path: UPath) -> UPath:
+def cheap_resolve(path: UPath) -> UPath:
     from upath.implementations.http import HTTPPath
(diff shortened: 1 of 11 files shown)


webknossos/webknossos/utils.py
```diff
@@ -562,17 +562,23 @@ def enrich_path(path: str | PathLike | UPath, dataset_path: UPath) -> UPath:
...
+    if not path.endswith("/"):
+        return path + "/"
+    return path
+
+
 def dump_path(path: UPath, dataset_path: UPath | None) -> str:
     if is_fs_path(path) and not path.is_absolute():
         if dataset_path is None:
             raise ValueError("dataset_path must be provided when path is not absolute.")
         path = dataset_path / path
```
fm3, line 575: Other edge cases that come to mind: What happens if dataset_path also is not absolute? what if “..” is contained?

fm3 (approved):

PR #1349, opened 2025-08-05
Support reading wkw datasets with new mag key instead of wkwResolutions

fm3:
As of https://github.com/scalableminds/webknossos/pull/8708 webknossos writes out wkw layers with the key "mag" instead of the legacy "wkwResolutions".

That means that the libs need to be able to read these datasource-properties.jsons.

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/properties.py
+++ b/webknossos/webknossos/dataset/properties.py
@@ -368,10 +368,11 @@ def layer_properties_pre_structure(
         type_value: type[LayerProperties | SegmentationLayerProperties],
     ) -> LayerProperties | SegmentationLayerProperties:
         if d["dataFormat"] == "wkw":
-            d["mags"] = [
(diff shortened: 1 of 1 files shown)


hotzenklotz (approved):
LGTM

webknossos/webknossos/dataset/properties.py
```diff
@@ -368,10 +368,11 @@ def __layer_properties_pre_structure(
         type_value: type[LayerProperties | SegmentationLayerProperties],
     ) -> LayerProperties | SegmentationLayerProperties:
         if d["dataFormat"] == "wkw":
```
valentin-pinkau, line 370: So this is basically an evolution of all possible input dicts to the current schema?
And this is why we need to check if "wkwResolutions" is present, as it might not be, since the current version does not have it anymore?

fm3: Yes, that is my understanding also.

PR #1385, opened 2025-10-21
Fix test.py when refreshing snapshots (nightly tests)

fm3:
### Description:
- Follow-up fixes for python port of test.sh/test.py (correct docker image name, handle exceptions in retried requests)

### DIFF
--- a/webknossos/test.py
+++ b/webknossos/test.py
@@ -80,14 +80,7 @@ def start_wk_via_docker(wk_docker_dir: Path, wk_docker_tag: str) -> None:
     )
 
     # Wait for booting
-    wk_healthy = False
-    while not wk_healthy:
-        try:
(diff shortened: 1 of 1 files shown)


webknossos/test.py
```diff
@@ -80,7 +80,14 @@ def start_wk_via_docker(wk_docker_dir: Path, wk_docker_tag: str) -> None:
     )
 
     # Wait for booting
-    while not requests.get(f"{WK_URL}/api/v{WK_API_VERSION}/health").ok:
+    wk_healthy = False
+    while not wk_healthy:
+        try:
+            if requests.get(f"{WK_URL}/api/v{WK_API_VERSION}/health").ok:
+                wk_healthy = True
+        except Exception:
```
normanrz: Could this be narrowed to specific exception classes?

normanrz (commented):
Just a few comments.

PR #1386, opened 2025-10-27
RemoteDataset.used_storage_bytes and n5/neuroglancer read support

normanrz:
### Description:
- Adds `RemoteDataset.used_storage_bytes`. Comments on the naming welcome.
  Allows for queries such as:
  ```python
  import webknossos as wk
  datasets = wk.RemoteDataset.list()
  for ds_id in datasets.keys():
      try:
          ds = datasets[ds_id]
          print(f"{ds_id}, {ds.name}, {ds.used_storage_bytes}")
      except Exception as e:
          print(ds_id, e)
  ```
- Adds read support for n5 and neuroglancer_precomputed to be on-par with Webknossos.
- Removes checking of mag data if "numChannels" metadata is missing. Assumes numChannels=1.

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/cluster_tools/Changelog.md
+++ b/cluster_tools/Changelog.md
@@ -7,16 +7,22 @@ and this project adheres to [Semantic Versioning](http://semver.org/) `MAJOR.MIN
 For upgrade instructions, please check the respective *Breaking Changes* sections.
 
 ## Unreleased
-[Commits](https://github.com/scalableminds/webknossos-libs/compare/v2.5.0...HEAD)
+[Commits](https://github.com/scalableminds/webknossos-libs/compare/v3.0.0...HEAD)
 
 ### Breaking Changes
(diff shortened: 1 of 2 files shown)


webknossos/webknossos/dataset/layer/view/_array.py
```diff
@@ -797,6 +815,7 @@ def __setstate__(self, d: dict[str, Any]) -> None:
 
 class Zarr3Array(TensorStoreArray):
     data_format = DataFormat.Zarr3
+    _ts_data_format = "zarr3"
```
fm3: I’d rename it to _tensorstore_data_format to make it more glanceable what this is about. Or maybe even create a translation function somewhere from DataFormat to the tensorstore string and use that instead of a defining this redundantly here.

fm3 (approved):

PR #1388, opened 2025-10-30
In pims tiff reader, leave chunks black that yield KeyError when accessing zarr store

fm3:
Converting some tiff files resulted in KeyErrors in this as_zarr implementation, while imread would yield black at those spots.

We can fall back to black by skipping the missing chunks.

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
+++ b/webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
@@ -211,6 +211,7 @@ class PimsTiffReader(FramesSequenceND):
                     ]
                 except KeyError:
                     # chunk not present in zarr_store, leave black.
+                    # ruff: noqa: PERF203 try-catch is still faster than in for the ZarrStore
                     pass
 
             return out


webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
```diff
@@ -199,15 +199,13 @@ def get_frame_2D(self, **ind: int) -> np.ndarray:
...
-                    .ravel()
-                    .reshape(chunk_shape)
-                )
-                # write in output array
-                out[chunk_projection.out_selection] = chunk_data[
-                    chunk_projection.chunk_selection
-                ]
+                chunk_key = ".".join(map(str, chunk_projection.chunk_coords))
+                if chunk_key in zarr_store:
+                    chunk_data = zarr_store[chunk_key].ravel().reshape(chunk_shape)
```
normanrz:
```suggestion
                chunk_data_maybe = zarr_store.get(chunk_key)
                if chunk_data_maybe is not None:
                    chunk_data = chunk_data_maybe.ravel().reshape(chunk_shape)
```

normanrz: Don't know how expensive the `in` is in tifffile.

PR #1390, opened 2025-11-06
Add transfer mode support to RemoteAttachments. 

valentin-pinkau:
### Description:
- This PR adds SYMLINK transfer mode
- Introduces a new method on RemoteAttachments: upload_attachment
- Makes sure syncing to the remote state is applied to segmentation layers and their attachments as well
- Makes RemoteLayers read-only in case of zarr streaming.

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -15,10 +15,14 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Breaking Changes
 
 ### Added
+- Added symlink transfer mode. [#1390](https://github.com/scalableminds/webknossos-libs/pull/1390)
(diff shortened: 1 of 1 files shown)


webknossos/tests/dataset/test_dataset.py
```diff
@@ -2258,12 +2258,13 @@ def test_dataset_shallow_copy(data_format: DataFormat, output_path: UPath) -> No
     assert len(Dataset.open(copy_path).get_layer("color").mags) == 3, (
         "Expecting all mags from original dataset and new downsampled mag"
     )
-    assert (
+    assert str(
```
fm3: why stringify the paths for the comparison?

PR #1400, opened 2025-11-25
Make sharing_token usage more explicit, don’t pass context token as query param

### LINKED ISSUE #1397: Remove user token from logs
## Context
Currently, opening or failing to open a remote dataset (or annotation) with the wklibs puts the user token into the related logging message. The token appears in the URL as `sharingToken=`. This should not happen

## Steps to Reproduce the bug

- [ ] Cannot reproduce the bug anymore / needs deeper investigation.

1. Do something with a remote dataset that causes a log message to appear, e.g. running `RemoteDataset.open` from a context with a wrong token.

## Your Environment for bug
- wklibs version 3.0.3

fm3:
### Description:
- While dataset `sharing_token`s may occasionally be used as context tokens to access shared datasets, never should normal user tokens be passed as sharing_tokens. One problem this may cause is the token showing up in logging.
- This PR makes the use of sharing_tokens more explicit.
- In the methods `Dataset.trigger_dataset_import`, `Dataset.trigger_reload_in_datastore` and `RemoteDataset.trigger_reload_in_datastore` the parameter `token` was deprecated (use an authenticated webknossos context instead). In `RemoteDataset.download_mesh` and `RemoteSegmentationLayer.download_mesh` the parameter `token` was renamed to `sharing_token` to clarify that it is meant for a dataset sharing token. Otherwise, use an authenticated webknossos context instead.

### Issues:
- fixes #1397

### Todos:
 - [x] Updated Changelog


webknossos/tests/dataset/test_dataset_download_upload_remote.py
```diff
@@ -137,18 +136,7 @@ def test_remote_dataset(tmp_upath: UPath) -> None:
     assert remote_ds.get_color_layers()[0].get_finest_mag().read_only
 
     assert remote_ds.name == "test_remote_metadata"
-
-    # allow 10s for wk to scan the used storage bytes
```
fm3: the server’s finishUpload call now blocks until the storage is measured.

PR #1402, opened 2025-11-26
relax layer name constraints

normanrz:
### Description:
- Relax layer name constraints when opening datasets

### Todos:
 - [x] Updated Changelog


fm3 (approved):
Tests are probably fine, but a changelog entry might make sense

PR #1405, opened 2025-12-08
Fix convert zarr

### LINKED ISSUE #1404: Conversion to webknossos dataset from Zarr fails due to UPath being passed to ternsorstore.open()
## Context
- Affected library: Webknossos CLI

Trying to convert a Zarr dataset to WBK dataset using `webknossos convert-zarr`

## Current Behavior

The path to the Zarr dataset is converted to a UPath before being opened with Tensorstore.
UPath cannot be serialised to JSON and throws an error:
    return tensorstore.open(
TypeError: Object does not support conversion to JSON

When in the except clause the wrong exception is caught:
except tensorstore.TensorStoreError:
AttributeError: module 'tensorstore' has no attribute 'TensorStoreError'. 
### pulled from [webknossos convert-zarr](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/cli/convert_zarr.py#L39)
```
def _try_open_zarr(path: UPath) -> tensorstore.TensorStore:
    try:
        return tensorstore.open(
            {"driver": "zarr3", "kvstore": {"driver": "file", "path": path}}
        ).result()
    except tensorstore.TensorStoreError:
        return tensorstore.open(
            {"driver": "zarr", "kvstore": {"driver": "file", "path": path}}
        ).result()
source = "/Path/to/zarr/directory/"
source_u = UPath(source)
res = _try_open_zarr(source_u)
```
## Your Environment for bug

- webknossos version: 3.0.4
- Python version: 3.12
- OS: Rocky Linux

normanrz:
### Description:
- Fixes convert-zarr CLI tool

### Issues:
- fixes #1404 

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/cli/convert_zarr.py
```diff
@@ -37,14 +37,12 @@
...
-        return tensorstore.open(
-            {"driver": "zarr3", "kvstore": {"driver": "file", "path": path}}
-        ).result()
-    except tensorstore.TensorStoreError:
-        return tensorstore.open(
-            {"driver": "zarr", "kvstore": {"driver": "file", "path": path}}
-        ).result()
+    return tensorstore.open(
+        {
+            "driver": "auto",
```
fm3: I assume this merging of the two codepaths with "auto" is not part of the fix, but instead a refactoring using the auto feature of tensorstore now?

Issue #1398, opened 2025-11-24
Make RemoteDatasets editable


AndreaNathansen:
## Detailed Description
- Affected library: webknossos

With the new way of storing the datasource properties in a database and not a JSON file in the dataset directory, it isn't possible to open such datasets as normal `Dataset` anymore, only as `RemoteDataset`. Currently, they and their RemoteLayers cannot be edited, e.g., adding a new layer, modifying a layer or downsampling/upsampling of a layer. This should be changed, so that we can edit new datasets with wklibs again.

## Use Cases & Context

This is relevant for all recent datasets that do not have a `datasource_properties.json` anymore (and maybe also the ones on s3?) and therefore require being opened remotely. So, it's important if we want to still be able to modify datasets with the wklibs in the future.

PR #1416, opened 2026-01-15
Allow more edits for RemoteDataset

fm3:
### Description:
- New functions for RemoteDataset `add_layer`, `downsample` and RemoteLayer `add_mag_as_copy`, `delete_mag`, `name` (setter), `downsample`
- Corresponding WK PR: https://github.com/scalableminds/webknossos/pull/9193

### Issues:
- contributes to #1398

### TODO:
 - [x] add correct label to transfer_mode.transfer
 - [x] add downsample and redownsample calls
   - [x] downsample
   - [x] why does it not upload the other mags?
   - [x] align with other layers
 - [x] docstrings

### Steps to test:
I tested the different steps interactively like this, checking the results in wk (and on disk if something should be deleted). Note that the dataset id is that of the `l4_sample_remote` dataset, which is included in wk’s initial data.
```
    with wk.webknossos_context("http://localhost:9000", "secretSampleUserToken"):
        src_mag = (
            wk.Dataset.open("/home/f/scm/datasets/dsA").get_layer("color").get_mag("16")
        )
        remote = wk.Dataset.open_remote(
            dataset_id="68b80290d4000090f8f4aa62", use_zarr_streaming=False
        )
        remote.get_layer("color").delete_mag(16)
        remote_layer = remote.add_layer("third", "color", dtype_per_channel=np.uint16)
        remote_layer.add_mag_as_copy(src_mag)
        remote_layer.name = "third_renamed"
```

### DIFF
--- a/webknossos/tests/dataset/test_transfer_mode.py
+++ b/webknossos/tests/dataset/test_transfer_mode.py
@@ -16,7 +16,7 @@ def test_transfer_mode_move_and_symlink() -> None:
         src.mkdir(parents=True, exist_ok=True)
         (src / "a.txt").touch()
 
-        transfer_mode.transfer(src, dst)
+        transfer_mode.transfer(src, dst, progress_desc_label="test data")
 
         assert (dst / "a.txt").exists()
         assert (src / "a.txt").exists()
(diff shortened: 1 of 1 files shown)


webknossos/webknossos/dataset/abstract_dataset.py
```diff
@@ -115,10 +115,16 @@ def _load_dataset_properties(self) -> DatasetProperties:
         pass
 
     @abstractmethod
-    def _save_dataset_properties_impl(self) -> None:
+    def _save_dataset_properties_impl(
```
normanrz: It is odd to have this in the AbstractDataset, because it only relates to the RemoteDataset. Could you only add it there?

PR #1418, opened 2026-01-29
add proxy path support

normanrz:
### Description:
- Adds support for RemoteAccessMode.PROXY_PATH
```python
ds = wk.RemoteDataset.open("...", access_mode=wk.RemoteAccessMode.PROXY_PATH)

m = ds.get_layer("segmentation").get_finest_mag()
print(m.path)
print(m.read())

a = ds.get_layer("segmentation").attachments.agglomerates[0]
print(a.path)
print(ts.open({"driver":"zarr3", "kvstore":{"driver":"http", "base_url":f"{a.path}/agglomerate_to_segments", "headers": [f"x-auth-token: {a.path.storage_options['headers']['X-Auth-Token']}"]}}).result()[:].read().result())
```
  
- Moves some args to kwarg-only

To be used with https://github.com/scalableminds/webknossos/pull/9247

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/cluster_tools/Changelog.md
+++ b/cluster_tools/Changelog.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/) `MAJOR.MIN
 For upgrade instructions, please check the respective *Breaking Changes* sections.
 
 ## Unreleased
-[Commits](https://github.com/scalableminds/webknossos-libs/compare/v3.1.4...HEAD)
+[Commits](https://github.com/scalableminds/webknossos-libs/compare/v3.1.5...HEAD)
 
 ### Breaking Changes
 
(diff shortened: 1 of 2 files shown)


webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -69,6 +70,14 @@
     from webknossos.dataset import Dataset
 
 
+class RemoteAccessMode(Enum):
+    """The access mode determines how the dataset is accessed."""
```
fm3:
```suggestion
    """Determines how data of a remote dataset is accessed. Note that DIRECT_PATH can only be used if the client has access to the underlying storage."""
```

PR #1419, opened 2026-01-30
Handle c-axis more generally

normanrz:
### Description:
- Refactors bounding box handling to allow for channel axis at arbitrary locations
  - The interface of the BoundingBox class is kept intentionally stable
  - There is now a NormalizedBoundingBox class, which makes the c-axis explicit
  - The NormalizedBoundingBox is used everywhere internal to the library, including json de/serialization
  - The element class (dtype + num channels) and numchannels attributes are refactored as well
  - The BufferedSliceWriter/Reader now use axis-name based dimensions (e.g. `"z"` instead of `2`) 
- Cleans up method signatures (more kw-only)
- Refactors VecInt to use composition instead of subclassing tuple
- Adds a specification document for the datasource-properties.json to the docs

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -36,6 +36,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Added
 - Added support for proxy paths when accessing RemoteDatasets. Use `RemoteDataset.open(..., access_mode=RemoteAccessMode.PROXY_PATH)`. [#1418](https://github.com/scalableminds/webknossos-libs/pull/1418)
 - Added `parent` property to `RemoteFolder`. [#1418](https://github.com/scalableminds/webknossos-libs/pull/1418)
+- Added `RemoteAiModel` class, allowing to access information about trained AI models registered in WEBKNOSSOS. [#1409](https://github.com/scalableminds/webknossos-libs/pull/1409)
 
 ### Changed
 - The `RemoteDataset.open` method now accepts `access_mode` instead of `use_zarr_streaming`. `use_zarr_streaming` is still supported, but deprecated. [#1418](https://github.com/scalableminds/webknossos-libs/pull/1418)
--- a/webknossos/tests/test_utils.py
+++ b/webknossos/tests/test_utils.py
@@ -1,9 +1,11 @@
+from shutil import ignore_patterns
 from unittest.mock import Mock, patch
 
 import pytest
 from upath import UPath
 
-from webknossos.utils import call_with_retries, dump_path
+from tests.utils import TestTemporaryDirectoryNonLocal
+from webknossos.utils import call_with_retries, copytree, dump_path
 
 
 def test_call_with_retries_success() -> None:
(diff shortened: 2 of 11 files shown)


docs/src/webknossos-py/spec/datasource-properties.md
```diff
@@ -0,0 +1,342 @@
...
+
+---
+
+## DataFormat
+
+A string indicating the on-disk storage format.
+
+| Value | Description |
+|---|---|
+| `"wkw"` | WEBKNOSSOS Wrap format. Uses `wkwResolutions` for mags. |
```
fm3: can also use the new mags format

docs/src/webknossos-py/spec/datasource-properties.md
```diff
@@ -0,0 +1,326 @@
...
+| `team` | `string` | Yes | Organization identifier. |
+
+### VoxelSize
+
+Describes the physical size of a single voxel.
+
+| Field | Type | Required | Default | Description |
+|---|---|---|---|---|
+| `factor` | `[number, number, number]` | Yes | | Voxel dimensions as `[x, y, z]`. |
+| `unit` | [LengthUnit](#lengthunit) | No | `"nanometer"` | Unit of the voxel size factor. |
```
fm3:
```suggestion
| `unit` | [LengthUnit](#lengthunit) | Yes | `"nanometer"` | Unit of the voxel size factor. |
```
in this format the unit is currently required by the backend. Only the legacy format [number, number, number] assumes the default nanometer.

webknossos/tests/dataset/test_dataset.py
```diff
@@ -403,7 +398,7 @@ def test_create_default_mag(data_format: DataFormat) -> None:
     else:
         assert mag_view.info.shard_shape.xyz == Vec3Int.full(1024)
         assert mag_view.info.chunks_per_shard.xyz == Vec3Int.full(32)
-    assert mag_view.info.num_channels == 1
+    assert mag_view.info.shape.size.c == 1
```
fm3: maybe rename shape to bounding_box if it has a topleft (even if it that currently always zero).

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -215,14 +254,28 @@ def _flush_buffer(self) -> None:
...
+            elif len(data.shape) == 3:
+                assert "c" in self._bbox.axes
+                axis_map = {"c": 0, width_axis: 1, height_axis: 2}
+            else:
+                raise ValueError(f"Unsupported data shape: {data.shape}")
+
+            # The order of the axes in the input data might not match the order of the axes in the bbox
+            data = np.transpose(
+                data, [axis_map[a] for a in self._bbox.axes if a in axis_map]
+            )
```
fm3: is this still expensive if the map is identity? should there be a shortcut case?

webknossos/webknossos/dataset/layer/view/view.py
```diff
@@ -377,23 +425,24 @@ def write(
...
         shard_bbox = bbox.align_with_mag(shard_shape, ceil=True)
-        if shard_bbox.intersected_with(self.bounding_box.in_mag(self._mag)) != bbox:
+        self_bbox = self.normalized_bounding_box
+        if shard_bbox.intersected_with(self_bbox.in_mag(self._mag)) != bbox:
             raise ValueError(
                 f"The bounding box to write {bbox} is not aligned with the shard shape {shard_shape}. "
                 + "Performance will be degraded as existing shard data has to be read, combined and "
                 + "written as whole shards. Additionally, writing without shard alignment data can lead to "
-                + f"issues when writing in parallel. Bounding box: {self.bounding_box}",
+                + f"issues when writing in parallel. Bounding box: {self_bbox}",
```
fm3: While we’re here, maybe the wording of this error can be changed. “Performance will be degraded” reads like it fits a warning, not a ValueError 🤔

webknossos/webknossos/dataset/layer/view/mag_view.py
```diff
@@ -143,31 +149,18 @@ def create(
...
             zarr3_array_info = Zarr3ArrayInfo(
                 data_format=DataFormat.Zarr3,
                 voxel_type=layer.dtype_per_channel,
-                num_channels=layer.num_channels,
                 chunk_shape=chunk_shape,
                 shard_shape=shard_shape,
-                axis_order=axis_order,
-                shape=shape,
-                dimension_names=dimension_names,
+                shape=bbox,
```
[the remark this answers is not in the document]
normanrz: jup

webknossos/webknossos/dataset/layer/view/view.py
fm3: Maybe take it from the warning in the wkw-and-not-compressed case?

“Either ensure that you write shard-aligned chunks OR pass allow_unaligned=True. When using the latter, take care to not write concurrently. Also, performance will be degraded etc etc”

normanrz: Ok. Changed in https://github.com/scalableminds/webknossos-libs/pull/1419/commits/09791e774629dd50a897824731108e57c843db37

PR #1433, opened 2026-03-06
Remote convert cli

normanrz:
### Description:
- Added convert, downsample and upsample for remote datasets

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/cli/convert.py
+++ b/webknossos/webknossos/cli/convert.py
@@ -278,11 +278,12 @@ def main(
                 folder_obj: None | RemoteFolder = None
                 if folder is not None:
                     folder_obj = RemoteFolder.get_by_path(folder)
-                dataset.upload(
+                remote_dataset = dataset.upload(
                     new_dataset_name=name,
                     folder=folder_obj,
                     transfer_mode=transfer_mode,
                 )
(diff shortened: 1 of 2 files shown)


webknossos/webknossos/dataset/layer/remote_layer.py
```diff
@@ -203,6 +203,73 @@ def downsample(
...
+        )
+
+        with TemporaryDirectory() as tmpdir:
+            tmp_dataset = Dataset(
+                dataset_path=tmpdir,
+                voxel_size_with_unit=self.dataset.voxel_size_with_unit,
+            )
+            tmp_layer = tmp_dataset.add_layer_like(self, self.name)
+            tmp_layer.upsample(
+                from_mag=from_mag,
```
fm3: the tmp_layer doesn’t have the from_mag at this time, does it? How can this work? Don’t we have to pass in the MagView from self?

normanrz: right. a test would've caught that :D i added from_mag_view to Layer.upsample and implemented it here.

PR #1442, opened 2026-03-26
Batching executor

normanrz:
### Description:
- Add BatchingExecutor

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/docs/mkdocs.yml
+++ b/docs/mkdocs.yml
@@ -288,6 +288,9 @@ nav:
         - DataFormat: api/webknossos/dataset_properties/data_format.md
         - LengthUnit: api/webknossos/dataset_properties/length_unit.md
         - LayerToLink: api/webknossos/dataset/layer/layer_to_link.md
+      - Annotation:
+        - Annotation: api/webknossos/annotation/annotation.md
+        - AnnotationInfo: api/webknossos/annotation/annotation_info.md
       - Skeleton:
         - Skeleton: api/webknossos/skeleton/skeleton.md
         - Group: api/webknossos/skeleton/group.md


cluster_tools/cluster_tools/executors/batching.py
```diff
@@ -0,0 +1,128 @@
...
+                offset += size
+                try:
+                    results = batch_future.result()
+                    for f, r in zip(item_futures, results):
+                        f.set_result(r)
+                except Exception as e:
+                    for f in item_futures:
+                        f.set_exception(e)
+
+        threading.Thread(target=resolve_all, daemon=True).start()
```
tmbo: OPT: there isn't really any life cycle management of the thread here (e.g. on shutdown we don't wait for the completion of this thread, which might mean that the results are not set after shutdown if the results completed, but the resolver thread didn't). i don't think this is a big issue, as the window where this would be a problem is small.

if we do not rely on batches finishing in order, we could alternatively attach a done callback to the batch_future in L83 resolving the item futures with the results from the batch once it completes. a bit easier to reason about as it resolves the item futures in the thread that handles the batch result.

cluster_tools/cluster_tools/executors/batching.py
```diff
@@ -0,0 +1,128 @@
...
+
+class BatchingExecutor:
+    """
+    Wraps another executor and groups items into batches before submission.
+    Each call to the underlying executor processes batch_size items at once.
+    """
+
+    def __init__(self, executor: Executor, *, batch_size: int) -> None:
+        self._executor = executor
+        self.batch_size = batch_size
```
tmbo: I'd validate somewhere that the batchsize is >= 1

cluster_tools/tests/test_batching.py
```diff
@@ -0,0 +1,200 @@
+from concurrent.futures import CancelledError, Future
```
daniel-wer, line 1: Nice tests, but I don't think they were ran in the CI afaics. They need to be included in the ci.yml, e.g. [here](https://github.com/scalableminds/webknossos-libs/blob/master/.github/workflows/ci.yml#L90) (and rename that step to something more general).

cluster_tools/cluster_tools/executors/batching.py
```diff
@@ -0,0 +1,174 @@
...
+            def on_batch_done(
+                bf: Future[list[_T]], ifs: list[Future[_T]] = item_futures
+            ) -> None:
+                try:
+                    results = bf.result()
+                    for f, r in zip(ifs, results):
+                        f.set_result(r)
+                except Exception as e:
+                    for f in ifs:
+                        f.set_exception(e)
```
daniel-wer: No blocker, but for me the conciseness is a bit too much in this function. Spelling out at least `bf` and `ifs` would make it more readable imo.

PR #1451, opened 2026-04-08
Fix upsampled Mag-1 bounding box in BufferedSliceWriter

### LINKED ISSUE #1449: `BufferedSliceWriter` upsamples Mag-1 bounding boxes
When writing to a `MagView` with mag > 1 (e.g. 2-2-1) with a `BufferedSliceWriter`, it uses the topleft of the Mag-1 bbox and upscales it to Mag-1, resulting in incorrect coordinates.

Example:
- MagView with Mag 2-2-1
- Absolute Mag-1 bounding box: 5000_5000_0_1000_1000_10
- write 500x500 slice (since we are writing in Mag 2-2-1)
- creates new Mag 2-2-1 bounding box from original topleft and shape of slice: 5000_5000_0_500_500_1
- bounding box of slice to write will be upscaled to Mag-1: 10000_10000_0_1000_1000_1 ⚡
- see [CI log](https://github.com/scalableminds/webknossos-libs/actions/runs/24132374093/job/70411917733?pr=1451) for error message

gwauge:
### Description:
- fix BufferedSliceWriter upsampled mag1 bbox

### Issues:
- fixes #1449
- see [CI log](https://github.com/scalableminds/webknossos-libs/actions/runs/24132374093/job/70411917733?pr=1451) for error message using newly created test

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/tests/dataset/test_buffered_slice_utils.py
+++ b/webknossos/tests/dataset/test_buffered_slice_utils.py
@@ -379,9 +379,13 @@ def test_buffered_slice_writer_resize_error(tmp_upath: UPath) -> None:
 
 
 def test_buffered_slice_writer_mag2_absolute_bbox(tmp_upath: UPath) -> None:
-    target_mag = Mag("2-2-1")  # write to a non mag1 mag using an absolute bbox with offset
+    target_mag = Mag(
+        "2-2-1"
+    )  # write to a non mag1 mag using an absolute bbox with offset
 
-    mag1_bbox = BoundingBox((128, 128, 0), (32, 32, 8))  # intentionally start at some offset
+    mag1_bbox = BoundingBox(
+        (128, 128, 0), (32, 32, 8)
+    )  # intentionally start at some offset
     mag2_bbox = mag1_bbox.align_with_mag(target_mag).in_mag(target_mag)
 
     # Allocate some data
--- a/webknossos/tests/dataset/test_buffered_slice_utils.py
+++ b/webknossos/tests/dataset/test_buffered_slice_utils.py
@@ -376,3 +376,34 @@ def test_buffered_slice_writer_resize_error(tmp_upath: UPath) -> None:
             for z in range(0, shape[2]):
                 section = data[:, :, z]
                 writer.send(section)
+
+
+def test_buffered_slice_writer_mag2_absolute_bbox(tmp_upath: UPath) -> None:
+    target_mag = Mag("2-2-1")  # write to a non mag1 mag using an absolute bbox with offset
+
+    mag1_bbox = BoundingBox((128, 128, 0), (32, 32, 8))  # intentionally start at some offset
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
```diff
@@ -162,10 +162,14 @@ def _flush_buffer(self) -> None:
             )
             buffer_depth = min(self._buffer_size, len(self._slices_to_write))
 
+            # downsample bbox to the mag of the view,
+            # so that the topleft is in the mags coordinate system
+            mag = self._view.mag
+            downsampled_bbox = self._bbox.align_with_mag(mag).in_mag(mag)
```
Tobias314: `align_with_mag` can slightly alter the bbox coordinates. Do we want this here?

What about omitting this and just calling `in_mag`. This will throw an error if the given bbox is not already mag-aligned. This might make it easier to find bugs early on which can be caused by callers providing wrong coordinates (e.g. off-by-one bugs etc.).

Actually, we could already check the bbox (and throw an error) when creating the buffered writer.

webknossos/tests/dataset/test_buffered_slice_utils.py
```diff
@@ -376,3 +376,36 @@ def test_buffered_slice_writer_resize_error(tmp_upath: UPath) -> None:
...
                 writer.send(section)
+
+
+def test_buffered_slice_writer_mag2_absolute_bbox(tmp_upath: UPath) -> None:
+    # write to a non mag1 mag using an absolute bbox with offset
+    target_mag = Mag("2-2-1")
+
+    # intentionally start at some odd offset
+    mag1_bbox = BoundingBox((128, 128, 0), (32, 32, 8))
+    mag2_bbox = mag1_bbox.align_with_mag(target_mag).in_mag(target_mag)
```
Tobias314: Why, odd? Isn't the offset mag-aligned. Maybe just remove the "odd" and the `align_with_mag` call?

If we decide that the slice writer throws an error if given a bbox that is not mag-aligned, we could also think about asserting this error in a test.

Tobias314 (commented):
Nice, see my comments about making the bbox logic more strict to fail early for non-mag-aligned bboxes.

webknossos/webknossos/dataset/_utils/buffered_slice_writer.py
gwauge: Would you check this in `MagView.get_buffered_slice_reader()` then?

Tobias314: Or in the `__init__` of the BufferedSliceWriter.

Tobias314 (approved):
Nice

PR #1456, opened 2026-04-22
Add layer view configurations

hotzenklotz:
### Description:
- Add read-only access to layer view configurations (`alpha`, histogram `min`/`max`etc) for `RemoteDatasets` via API client

### Issues:
- fixes #...

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests


fm3 (approved):
LGTM. If this property is accessed a lot, which I can see happening sooner or later, it may make sense to add caching to avoid excessive requests to the server.

PR #1461, opened 2026-04-29
add read_cxyz and write_cxyz

normanrz:
### Description:
- Added `View.{read,write}_cxyz` to normalize any non-default axis orderings

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/cli/export_as_tiff.py
+++ b/webknossos/webknossos/cli/export_as_tiff.py
@@ -5,7 +5,6 @@ from functools import partial
 from math import ceil
 from typing import Annotated, Any
 
-import fastremap
 import numpy as np
 import typer
 from cluster_tools import Executor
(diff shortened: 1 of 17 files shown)


webknossos/webknossos/cli/export_as_tiff.py
```diff
@@ -71,6 +70,8 @@ def _slice_to_image(data_slice: np.ndarray, downsample: int = 1) -> np.ndarray:
 
 
 def _apply_mapping(data: np.ndarray, mapping_array: TensorStore) -> np.ndarray:
+    import fastremap
```
fm3, line 74: below you use try-except. Should this be done here too? Could the string be extracted to something like a WkImportError so that it could be used in all such spots?

fm3 (approved):

webknossos/webknossos/cli/export_as_tiff.py
normanrz: This function is only called after fastremap was confirmed to be installed. I added a comment and WkImportError.

PR #1464, opened 2026-05-12
thread safe MultiprocessingLoggingHandler

Tobias314:
### Description

This PR makes `MultiprocessExecutor` safe to use from multiple threads simultaneously. Previously, concurrent use from different threads could cause race conditions in logging setup and broken signal/atexit handling.

#### Thread-safe logging (`MultiprocessingLoggingHandler`)

- Added a module-level `LOGGER_LOCK` to serialize modification to the root logger (e.g. adding/removing handlers, adjusting log levels) since the root logger is a singleton shared across threads.
- `_MultiprocessingLoggingHandlerPool` now reads `_owner_thread_id` which can be set for log handler. Handler which do not belong to the current thread are skipped. This prevents one thread's executor from wrapping or tearing down another thread's task-specific file handler and is required for parallel task execution via multiple thread in voxelytics.
- Thread-owned handlers (those with `_owner_thread_id` set) are **no longer closed by the pool** when the usage counter reaches zero — their owning thread (e.g. `attach_logging_handler`) is responsible for closing them.
- `decrement_usage` now holds the lock for the entire counter-decrement + handler-swap, eliminating a TOCTOU race where two threads could both see `_usage_counter == 0`.

#### Thread-safe signal/atexit handling (`ClusterExecutor`)

- `signal.signal()` only works from the main thread. Instantiating a `ClusterExecutor` from a worker thread (e.g. a vx task run in a separate non-main thread) would silently mark signal handlers as installed and permanently block any future main-thread installation.
- Now, if the executor is created from a worker thread, it logs a warning and **does not mark the signal handler as installed**, so a later main-thread instantiation (or an explicit `install_signal_handlers()` call) can still set it up.
- Added a public `install_signal_handlers()` class method that can be called once from the main thread at startup. **Good practice would be to call this `install_signal_handlers()` once in the beginning from the main thread before spawning any threads**
- Added `atexit.register(_run_shutdown_hooks)` so that shutdown hooks also run on normal process exit, not only on SIGINT/SIGTERM — e.g. for worker-thread executors that did not install signal handlers.
   - I think atexit is not equivalent to the signal handlers since it cannot listen to SIGTERM but having both the signal handlers plus the atexit handler should be somewhat robust and even work in most cases even if  `install_signal_handlers()`  was not called from the main thread
- Extracted `_run_shutdown_hooks()` as a shared helper method used by both the signal handler and the atexit handler, and also fixed pre-existing bugs?:
   - do not cancel all hooks if one fails
   - copy the list of hooks before starting iterating it since it might change during the iteration (hooks unregister themselves)
### Context

Required for https://github.com/scalableminds/voxelytics/pull/4555, which runs multiple `MultiprocessExecutor` instances concurrently from a thread pool.

### Todos
 - [x] Updated Changelog


cluster_tools/cluster_tools/_utils/multiprocessing_logging_handler.py
```diff
@@ -84,17 +88,22 @@ def increment_usage(self) -> None:
...
+            if self._usage_counter == 0:
+                # unwrap inner handler:
+                root_logger = getLogger()
+                root_logger.removeHandler(self)
+                root_logger.addHandler(self.wrapped_handler)
+
+                # Thread-owned handlers (e.g. task-specific file handlers) are closed by their
+                # owner (attach_logging_handler), not by the pool.
+                if not getattr(self.wrapped_handler, "_owner_thread_id", None):
+                    self.wrapped_handler.close()
```
daniel-wer, line 102: Was this needed for some specific reason or is it just best practice?

cluster_tools/cluster_tools/schedulers/cluster_executor.py
```diff
@@ -162,9 +183,28 @@ def _ensure_signal_handlers_are_installed(cls) -> None:
...
+
     @classmethod
     def _register_shutdown_hook(cls, hook: Callable[[], None]) -> None:
-        cls._shutdown_hooks.append(hook)
+        with cls._state_lock:
+            cls._shutdown_hooks.append(hook)
+            # New active executors exist; allow hooks to run again if signaled.
+            # This handles the case where a previous signal already ran hooks
+            # (setting _shutdown_hooks_ran=True) but new executors are now active.
+            cls._shutdown_hooks_ran = False
```
daniel-wer, line 211: Do we really want to allow to register new shutdown hooks after shutdown was already triggered before? It seems like something has gone wrong in this case 🤔

daniel-wer (commented):
Looks reasonable to me, thanks for working on this! Did you test whether the atexit handling actually works? I just want to make sure it's actually triggered and that the program doesn't stall before somehow.

I'll be looking at the voxelytics PR next.

cluster_tools/cluster_tools/schedulers/cluster_executor.py
Tobias314: Should not happen during normal execution but was required for the tests to work. I think the reason is that multiple tests run in the same process.
So I though this is an easy way to make this work in the tests while also making the code more robust against weird edge cases where new shutdown hooks are registered for some reason.

daniel-wer: Please add the information that this should only be relevant during testing to the comment to avoid future confusion.

cluster_tools/cluster_tools/_utils/multiprocessing_logging_handler.py
Tobias314: I think this is needed because for each task/thread vx attaches a custom task-logging-handler (with _owner_thread_id set). Once the task is done, vx tries to close this handler. However, without this check the handler is already closed earlier when the  _MultiprocessingLoggingHandler is shut down.

I think an underlying issue might be that the _MultiprocessingLoggingHandler actually shuts down handlers. Shouldn't it only "unwrap" them?
But this already was the case before this PR so I am not sure whether changing this would have other implications leading to problems.

daniel-wer: Yes makes sense :+1:

daniel-wer (approved):
LGTM 👍 Please adapt the comment as indicated in the unresolved feedback thread before merging.

PR #1474, opened 2026-06-12
Add Layer.transform and Layer.transform_affine

Tobias314:
## Description

Adds two new methods to `Layer` for resampling a layer's voxel data into another layer (upstreamed from an internal `volume_stitching_tools` implementation):

- **`Layer.transform(output_layer, inverse_transform, ...)`** — for every voxel of the output bounding box, an arbitrary `inverse_transform` callable (mapping (N, 3) Mag(1) output coordinates to input coordinates, NaN rows = no source) looks up the source position and the data is copied with nearest-neighbor sampling. The output layer may live in a different dataset; an optional `input_mask_layer` restricts copying to voxels where the mask is > 0 in all channels. Operates on a single mag (defaults to the finest available), returns the written `BoundingBox`. Output bounding boxes with negative topleft are shifted into positive space (`translate_to_positive`).
- **`Layer.transform_affine(output_layer, affine_matrix, ...)`** — wrapper accepting a 4x4 homogeneous forward matrix; it is inverted internally and, when no `output_bbox` is given, the output bounding box is computed from the transformed input bbox corners.

Key properties:

- **Executor support**: processing is chunked via `View.for_each_chunk` and parallelized with cluster_tools executors (multiprocessing, slurm, sequential), following the `downsample_mag` pattern.
- **Shard-aligned parallelism**: the default chunk per job is exactly one shard and explicit `chunk_shape`s must be shard multiples, so parallel jobs never write to the same shard.
- **Picklable job path**: cluster_tools uses plain pickle, so no lambdas/closures are used internally (the `translate_to_positive` shift is passed as a job argument; the affine callable is a picklable class). `inverse_transform` is documented to require picklability.
- **scipy-compatible sampling**: nearest-neighbor conventions match `scipy.ndimage.affine_transform` with `order=0`, `mode="constant"` (ties round up, coordinates valid up to the last sample position per axis).

## Tests

New `tests/dataset/test_transform_layer.py` with 9 tests: identity roundtrip (incl. bbox-truncated border chunks), translation, 90° affine rotation vs `np.rot90`, 3x affine scale with auto-computed bbox, full-array bit-identical comparison of a rotation+anisotropic-scale+translation affine against `scipy.ndimage.affine_transform`, multi-channel mask layer, Mag(2) transform, multiprocessing executor (pickling regression test), and negative output bbox handling.

## Todos

- [x] Updated Changelog
- [x] Added / updated tests
- [x] Considered adding documentation


webknossos/webknossos/dataset/layer/_transform_utils.py
```diff
@@ -0,0 +1,199 @@
...
+    relevant_input_bbox = BoundingBox(
+        topleft=input_topleft,
+        size=np.round(input_coords.max(axis=0) + 2 * mag_np - input_topleft).astype(
+            np.int64
+        ),
+    ).align_with_mag(mag, ceil=True)
+
+    input_data = input_mag_view.read(absolute_bounding_box=relevant_input_bbox)
+
+    output_coords = ((output_coords - chunk_topleft) / mag_np).astype(np.int64)
```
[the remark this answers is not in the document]
Tobias314: Claude and I thinks that's nonsense. Explanation by Claude:

>   1. output_coords are exact integers. They come from np.arange(topleft, bottomright, step=mag, dtype=np.float64) where topleft and mag are integers. Every value is topleft + k·mag — integer-valued float64 well below 2^53, computed from exact operations
  (integer-valued start + integer-valued product), so there's no representation error anywhere. The bot's "1.0 stored as 0.9999…" scenario requires an inexact operation to have happened first; none has.
>  2. The subtraction is exact. chunk_topleft is an integer array; differences of exactly-represented integers are exact.
>  3. The division is exact. output_coords - chunk_topleft is always an exact multiple of mag_np per axis: tile boundaries from chunk_bbox.chunk(buffer·mag, buffer·mag) lie at absolute multiples of buffer·mag (or at the chunk topleft itself), and the
  chunk topleft is mag-aligned because the output bbox is align_with_mag-ed and job-chunk borders are multiples of shard·mag. IEEE-754 division is correctly rounded, so when the true quotient is an exactly representable integer, the result is that
  integer exactly — and astype(np.int64) truncation of an exact integer is lossless.

Tobias314:
I have looked over the code and it looks good.

Not sure, whether we want to discusss the function interface. Some parameters like `translate_to_positive` are a bit specific but were found to be useful while using similar code in other projects.

I think, having this functionality in a private module allows us to use/test this while still allowing us to change the interface if needed.

Tricky details are the coordinate rounding, especially near bbox borders and how to fill voxels that could not be transformed. But I think the current implementation that does the rounding similar to scipy and that uses either a constant fill value or keeps existing values for every voxel  that could not be computed (e.g. due to it being masked out or mapping outside of the input volume) sounds reasonable.

One could also think about changing the dtype used for the coordinates that are transformed. Currently, it seems to be float64.

The test cases look good. I think some specific edge cases like some of the exceptions that can be thrown for wrong inputs are not covered. But I am not sure whether we want to blow up the tests with this, especially given, that this is a private module for now where the interface might change.

webknossos/webknossos/dataset/layer/_transform_utils.py
```diff
@@ -0,0 +1,486 @@
...
+
+
+def determine_transform_buffer_shape(input_chunk_shape: Vec3Int) -> Vec3Int:
+    # Roughly 256**3 voxels per processing tile (bounding the coordinate arrays
+    # to a few hundred MB), rounded up to a multiple of the input storage chunk
+    # shape since reads decompress whole input chunks anyway.
+    return input_chunk_shape * Vec3Int.full(256).ceildiv(input_chunk_shape).pairmax(1)
+
+
+class AffineTransform:
```
normanrz: Should we define a `AbstractTransform` base class that has `__call__` (or `apply`) and `inverse` methods?

webknossos/webknossos/dataset/layer/_transform_utils.py
```diff
@@ -0,0 +1,486 @@
...
+        output_view.for_each_chunk(
+            func,
+            chunk_shape=chunk_shape,
+            executor=actual_executor,
+            progress_desc=progress_desc,
+        )
+    return output_bbox
+
+
+def transform_affine(
```
normanrz: Thinking about it, this seems pretty redundant if it can be fully replaced by `transform(input, output, AffineTransform(np.array([...])))`.

normanrz (changes requested):

webknossos/webknossos/dataset/layer/_transform_utils.py
Tobias314: Ok, done. I slightly changed the interface to have both `transform` and `inverse_transform`. Reason: we need the inverse transform internally. For `transform` one has to give an `AbstractTransform` which defines an inverse. Currently, that probably means giving an `AffineTransform`.

Alternative, one can give any coordinate transforming callable as `inverse_transform`. This is especially useful, if one only has the inverse transform and does not want to define a "forward transform".

E.g. during some custom non-linear warping, its easier to just program the inverse transform code while not requiring to think about the code for the "forward" transform.

webknossos/webknossos/dataset/layer/_transform_utils.py
```diff
@@ -0,0 +1,463 @@
...
+    `inverse`. `transform` uses a transform's `inverse` to look up, for every output voxel,
+    the corresponding position in the input layer.
+    """
+
+    @abstractmethod
+    def apply(self, points: np.ndarray) -> np.ndarray:
+        """Applies the forward transform to an (N, 3) array of points."""
+        ...
+
+    def __call__(self, points: np.ndarray) -> np.ndarray:
```
normanrz, line 54: Maybe we should remove the `__call__`? `apply` and `inverse` seem more idomatic to me.

normanrz (approved):

webknossos/webknossos/dataset/layer/_transform_utils.py
normanrz: Let's keep the `__call__` as skimage also does that.

PR #1477, opened 2026-06-18
Add ChunkedImages abstraction for .ims and MRC conversion

normanrz:
## Summary

Adds a `ChunkedImages` abstraction for volumetric, chunk-based input formats — used to convert Imaris `.ims` and MRC files — as an alternative to the existing pims-based `PimsImages` path.

`PimsImages` reads slice-by-slice and can't always know its true x/y extent ahead of time, so `add_layer_from_images` inflates the layer's bounding box to a placeholder and corrects it after conversion. Chunk-based formats know their exact bounding box from metadata alone, so `ChunkedImages` implementations skip that placeholder dance entirely and read/write whole shard-sized blocks directly instead of going through `BufferedSliceWriter`.

- **`ChunkedImages`** (`webknossos/dataset/_utils/chunked_images.py`): abstract base + a small registry (`register_chunked_images`, `try_open_chunked_images`, `get_valid_chunked_image_suffixes`) so new formats (zarr, n5, neuroglancer_precomputed are natural future candidates) just need a new subclass — no changes to `dataset.py`.
- **`ImsChunkedImages`** (`ims_chunked_images.py`): reads shard-sized 3D/4D blocks directly from the underlying HDF5 file via `h5py`. Supports single- and multi-channel files, and multi-timepoint files (chunked along a `"t"` axis) without requiring an explicit `timepoint=`.
- **`MrcChunkedImages`** (`mrc_chunked_images.py`): reads blocks directly from `mrcfile`'s memory-mapped array — even simpler than IMS since MRC has no internal chunking, channels, or timepoints.
- `PimsImsReader` and `PimsMrcReader` (the previous pims-plugin-based readers) are **fully retired** — `.ims` and MRC files are read exclusively through `ChunkedImages`, never through pims/bioformats, including `Dataset.from_images()`'s directory discovery and z-dimension probing.

## Design notes

- `ChunkedImages.expected_bbox` is exact (from file metadata), so `add_layer_from_images` sets `layer.bounding_box` directly instead of the `SAFE_LARGE_XY` placeholder + post-hoc `dimwise_max(shapes)` correction used for the generic pims path. That correction was actually a latent bug for chunk-based formats: it's only correct when the image fits within a single output shard in x/y — for a multi-shard-wide image it would have shrunk the final bbox down to one shard's size.
- Both readers reopen their file handle per chunk (`h5py.File`/`mrcfile.mmap`) rather than keeping one open, since the underlying handles aren't picklable and mustn't cross a multiprocessing boundary.
- `flip_z` mirrors the reader's *entire* z-extent (matching `PimsImages.copy_to_view`'s semantics of reversing the whole sequence before batching), not just the current chunk in isolation — this required reading a mirrored source range per chunk and reversing it back into output order.
- Combining a multi-timepoint IMS file (t > 1, no explicit `timepoint=`) with 3+ channels isn't supported, since `expected_bbox` doesn't carry a `"c"` axis; this now raises a clear `ValueError` asking for an explicit `timepoint=` instead of silently misbehaving.

## Test plan

- [x] `.ims` conversion: single/multi-channel selection, flip/swap correctness (verified against an independent h5py reference), multi-shard-bbox correctness, multi-timepoint conversion, and the multi-channel+multi-timepoint guard
- [x] `Dataset.from_images()` directory discovery and z-dimension probing correctly pick up `.ims`/MRC via the `ChunkedImages` registry (no dependency on pims registration)
- [x] Multi-channel `.ims` file (`allow_multiple_layers=True`, used internally by `Dataset.from_images()`) creates one layer per channel with distinct data
- [x] MRC: basic conversion, flip/swap, multi-shard-bbox, metadata, and per-chunk mmap-reopening behavior
- [x] Full `test_add_layer_from_images.py` + `test_from_images.py` suites pass (aside from a pre-existing, unrelated JVM/DICOM failure in this environment)


normanrz:
Close in favor of #1495

PR #1480, opened 2026-06-23
Allow Not Tailing All Log Lines on Completed Futures

Tobias314:
### Description:
- calling .follow() on a already cancelled Tail object can lead to errors if the log file does not exist (e.g. on a slow NFS the current node might not immediately see a file written  from another node)
- Instead of failing, we just print a warning now
- Might loose some logs when a job is canceled before its log file is detected but might be better than failing, and assuring full logs might be hard with the current implementation + inconsistent NFS.

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/cluster_tools/Changelog.md
+++ b/cluster_tools/Changelog.md
@@ -14,7 +14,6 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Added
 
 ### Changed
-- Stop 
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


cluster_tools/Changelog.md
```diff
@@ -14,9 +14,10 @@ For upgrade instructions, please check the respective *Breaking Changes* section
 ### Added
 
 ### Changed
+- Stop
```
daniel-wer:
```suggestion
```

daniel-wer (approved):

AndreaNathansen (approved):
LGTM!

PR #1481, opened 2026-06-24
adds fix for virtual ImageJ stacks

normanrz:
### Description:
- Fixed a conversion issue with Tiff files produced by ImageJ virtual stacks.

### Issues:
- `uv run webknossos convert data/imagej_virtual_stack data/test --jobs 2 --voxel-size 3,3,3 --no-downsample`

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
+++ b/webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
@@ -51,6 +51,11 @@ class PimsTiffReader(FramesSequenceND):
                 axis for axis in self._tiff_axes if axis not in self._other_axes
             )
 
+            # ImageJ virtual stacks store all frames contiguously after a single IFD
+            # (series.is_truncated). In that case series.pages only exposes 1 real
+            # page and we must seek by computed byte offsets instead.
+            self._series_is_truncated = _tiff.is_truncated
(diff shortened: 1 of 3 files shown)


webknossos/webknossos/dataset/_utils/pims_tiff_reader.py
```diff
@@ -96,9 +107,24 @@ def get_frame_2D(self, **ind: int) -> np.ndarray:
...
+                            f"Expected {frame_bytes} bytes, got {len(raw)}."
+                        )
+                    page_data: np.ndarray = (
+                        np.frombuffer(raw, dtype=raw_dtype)
+                        .reshape(self._shape)
+                        .astype(self._dtype)
+                    )
+                else:
+                    page = pages[page_idx]
+                    assert page is not None, f"Page {page_idx} not found in TIFF file."
```
fm3: While we’re here: since this depends on user input, raising an exception would be more idiomatic than using assert

fm3 (approved):
Looking good, thanks for fixing this! Is `is_truncated` really always an ImageJ stack? Or should the comment rather say something like “truncated tiff series (as produced for example by ImageJ)”?

PR #1483, opened 2026-07-02
Fix SlurmExecutor srun jobs failing on some Slurm >=22.05 clusters due to missing resource inheritance

Tobias314:
## Summary
- Since Slurm 22.05, resources requested via `#SBATCH` (e.g. `--cpus-per-task`) are no longer automatically inherited by `srun` calls inside the sbatch script, which can cause jobs to fail with errors such as `srun: fatal: cpus-per-task set by two different environment variables SLURM_CPUS_PER_TASK=3 != SLURM_TRES_PER_TASK=cpu=1` (probably due to some cluster-defined default which do not match the --cpus-per-task specified for the sbatch)
- `SlurmExecutor.inner_submit` now also passes some configured `job_resources` (e.g. `cpus-per-task`) as explicit flags to the `srun` call, in addition to the existing `#SBATCH` lines.

## Test plan (as executed by Claude)
- [x] Reproduced the bug against a live Slurm 25.05.1 cluster (via `cluster_tools/dockered-slurm`): confirmed the generated script previously emitted a bare `srun ...` line with no resource flags.
- [x] Verified the fix: the generated script now emits `srun --cpus-per-task=1 ...`, restating the resource explicitly.
- [x] Ran a real job submission end-to-end with `job_resources={"cpus-per-task": "1"}` — succeeds.
- [x] Ran the full CI slurm test suite (`test_all.py`, `test_slurm.py`) against the fix: 30 passed, 2 skipped (expected/flaky-long-running), 2 deselected (require modified slurm config).
- [x] `ruff check` and `mypy` pass on the changed file.
- [x] Added a `Changelog.md` entry under `Unreleased > Fixed`.

### DIFF
--- a/cluster_tools/cluster_tools/schedulers/slurm.py
+++ b/cluster_tools/cluster_tools/schedulers/slurm.py
@@ -54,6 +54,22 @@ SLURM_STATES = {
     "Unclear": ["SUSPENDED", "REVOKED", "SIGNALING", "SPECIAL_EXIT", "STAGE_OUT"],
 }
 
+# Since Slurm 22.05, some resource flags are not automatically inherited from the
+# enclosing sbatch job by srun calls within the script anymore and must be restated
+# explicitly, or srun computes its own (potentially conflicting) resource request
+# leading to failures like:
+# "srun: fatal: cpus-per-task set by two different environment variables ..."
+# Other job_resources (e.g. --mem, --gres) should be inherited by srun automatically.
+# Therefore, we currently do not add them to the srun call
+SLURM_SRUN_NON_INHERITED_RESOURCE_KEYS = {
+    "cpus-per-task",
+    "ntasks",
+    "gpus",
+    "gpus-per-node",
+    "gpus-per-socket",
+    "gpus-per-task",
+}
+
 SLURM_QUEUE_CHECK_INTERVAL = 1 if "pytest" in sys.modules else 60
 
 T = TypeVar("T")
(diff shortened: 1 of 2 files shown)


cluster_tools/cluster_tools/schedulers/slurm.py
```diff
@@ -54,6 +54,22 @@
     "Unclear": ["SUSPENDED", "REVOKED", "SIGNALING", "SPECIAL_EXIT", "STAGE_OUT"],
 }
 
+# Since Slurm 22.05, some resource flags are not automatically inherited from the
+# enclosing sbatch job by srun calls within the script anymore and must be restated
+# explicitly, or srun computes its own (potentially conflicting) resource request
+# leading to failures like:
+# "srun: fatal: cpus-per-task set by two different environment variables ..."
+# Other job_resources (e.g. --mem, --gres) should be inherited by srun automatically.
+# Therefore, we currently do not add them to the srun call
```
daniel-wer: Just a note which we might want to add to this comment

> It is worth noting that this change was quite disruptive, particularly for MPI implementations. Because of this, in Slurm 23.11, SchedMD largely walked back this requirement. They reverted SRUN_CPUS_PER_TASK back to SLURM_CPUS_PER_TASK and introduced an --external-launcher option to srun to prevent task launchers from being bound to a single CPU.

daniel-wer (approved):

robert-oleynik (approved):

PR #1485, opened 2026-07-02
Remove storage leakage

### LINKED ISSUE #1476: annotation.download leaks storage
## Context
Affected library: webknossos
Affected classes: `Annotation` (`annotation.py`), indirectly `VolumeLayer` (`volume_layer.py`)

   Two code paths in `Annotation` create temporary zip files on disk but never clean them up:

   1. `Annotation.download()` — consolidates volume layer zips into a single temp file after loading
   2. `Annotation.add_volume_layer()` — creates a per-layer temp zip to back the new `VolumeLayer`

   The `NamedTemporaryFile` object is garbage-collected right away, deleting the file it created. `ZipFile` then recreates a new file at the same path — but this new file has no owner and is never deleted.

   ## Expected Behavior

   Temporary zip files backing `VolumeLayer` objects should be cleaned up when the owning `Annotation` is no longer in use (i.e. garbage collected or explicitly freed). Long-running scripts that download or create many annotations should not accumulate files in `/tmp`.

   ## Current Behavior

   Each `Annotation.download()` call and each `Annotation.add_volume_layer()` call leaves one orphaned `.zip` file in `/tmp`. These files are never deleted during the process lifetime, causing unbounded storage growth in `/tmp` for scripts that process many annotations.

MatthisCl:
### Description:
Downloading an annotation with volume data, or adding/editing a volume layer, silently wrote temporary zip files to the OS temp directory via tempfile.NamedTemporaryFile, even though the data was already available in memory.

see slack discussion: https://scm.slack.com/archives/C02Q35Q28P5/p1781788849421729

**Before:**
- Annotation.download() called _write_volume_layers() right after downloading -> consolidated all volume layers into one shared on-disk zip file created via NamedTemporaryFile.

- Annotation.add_volume_layer() created a NamedTemporaryFile on disk just to hold an empty placeholder zip for the new layer.

- VolumeLayer._write_dir_to_zip() reopened the volume layer's existing on-disk root zip file (self.zip.root.filename) and overwrote it in place with the updated data.

**Now:**
- Annotation.download() no longer needs a consolidation step at all — each volume layer already holds a valid in-memory ZipPath from _load_from_zip(). _write_volume_layers() was removed as dead code.

- Annotation.add_volume_layer() builds the empty placeholder zip in an in-memory io.BytesIO buffer instead.

- VolumeLayer._write_dir_to_zip() allocates a fresh io.BytesIO buffer per call and repoints self.zip at it, rather than overwriting the previous root file in place (which could have been shared with other volume layers).

### Issues:
- fixes #1476 

### Todos:
 - [x] Updated Changelog

### DIFF
--- a/webknossos/webknossos/annotation/volume_layer.py
+++ b/webknossos/webknossos/annotation/volume_layer.py
@@ -130,17 +130,18 @@ class VolumeLayer:
                         continue
                     volume_layer_zipfile.write(full_path, arcname)
 
-        volume_zip_buffer.seek(0)
+        # Repoint self.zip to a fresh in-memory root instead of rewriting the
+        # previous root, which may be shared with other volume layers.
+        annotation_zip_buffer = io.BytesIO()
         with ZipFile(
-            self.zip.root.filename,
+            annotation_zip_buffer,
             mode="w",
             compression=ZIP_DEFLATED,
             compresslevel=Z_BEST_SPEED,
         ) as annotation_zip:
-            annotation_zip.writestr(self.zip.at, volume_zip_buffer.read())
+            annotation_zip.writestr(self.zip.at, volume_zip_buffer.getvalue())
 
-        # updating self.zip.root.__lookup to include the new file
-        self.zip = ZipPath(self.zip.root.filename, self.zip.at)
+        self.zip = ZipPath(ZipFile(annotation_zip_buffer), self.zip.at)
         assert self.zip.exists()
 
     @contextmanager
--- a/cluster_tools/Changelog.md
+++ b/cluster_tools/Changelog.md
@@ -7,7 +7,7 @@ and this project adheres to [Semantic Versioning](http://semver.org/) `MAJOR.MIN
 For upgrade instructions, please check the respective *Breaking Changes* sections.
 
 ## Unreleased
-[Commits](https://github.com/scalableminds/webknossos-libs/compare/v3.5.3...HEAD)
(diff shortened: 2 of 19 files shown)


webknossos/webknossos/annotation/volume_layer.py
```diff
@@ -130,17 +130,18 @@ def _write_dir_to_zip(self, source: str) -> None:
...
             mode="w",
             compression=ZIP_DEFLATED,
             compresslevel=Z_BEST_SPEED,
         ) as annotation_zip:
-            annotation_zip.writestr(self.zip.at, volume_zip_buffer.read())
+            annotation_zip.writestr(self.zip.at, volume_zip_buffer.getvalue())
 
-        # updating self.zip.root.__lookup to include the new file
-        self.zip = ZipPath(self.zip.root.filename, self.zip.at)
+        self.zip = ZipPath(ZipFile(annotation_zip_buffer), self.zip.at)
```
philippotto, line 144: Can the old value in self.zip also contain a ZipPath(ZipFile()) object? should the ZipFile be closed explicitly?

philippotto (approved):
lgtm. see my one comment, though.

webknossos/webknossos/annotation/volume_layer.py
MatthisCl: Yes it can, for 2. I don't think so because the old ZipFile is shared with the volume layers and they will still read from there

philippotto: ok, then, let's merge :)

PR #1489, opened 2026-07-22
Faster BoundingBox Chunking Returning Topleft Chunk Corners as Integer Tuples

Tobias314:
### Description:
- the old `BoundingBox.chunk(...)` implementation allocated BoundingBox objects for all result chunks which led to a lot of overhead due to object creation e.g. when creating chunks in a tight loop
- the new code provides an implementation which works only on and returns lightweight integer tuples, skipping heavy allocations -> often, callers who require the speedup and are fine with not getting full BoundingBox objects back ->  required for https://github.com/scalableminds/voxelytics/pull/4673
- internal, the implementation of `chunk(...)` now also uses the lightweight method to compute the chunk positions, before allocating BoundingBox object for them
- slightly clarified/changed behavior on how to chunk nd-bounding-boxes depending on axes names to prevent unexpected corner cases allowed by the old code (see below)

### Changed behavior: how a `chunk_shape` is mapped to the box axes

A `chunk_shape` is now resolved in three tiers:

1. a `VecInt` **with axis names** is matched to the box's axes by name (reordered if needed) and used verbatim — the unambiguous form
2. **three unnamed entries** (tuple or `Vec3Int`) are the xyz shorthand: sizes for x, y and z **by name**, all other axes get size 1, except that `chunk()` lets the channel axis span all channels
3. **one entry per axis** is used verbatim

Mapping by name matters: a `("z", "y", "x")` box (zarr/NGFF order) chunked with `(8, 16, 32)` must give chunks of size `(32, 16, 8)`. All existing call sites pass a `Vec3Int`, so they stay on tier 2 — verified by diffing full `chunk()` output against master for 7 box layouts × 3 shapes × 2 alignments (63/63 identical, according to Claude).

**Edge case (behavior change):** for a box with exactly three axes that are *not* x, y and z, three unnamed entries fit both tier 2 and tier 3. Previously the third entry was silently dropped:

```python
bbox = NDBoundingBox(topleft=(10, 20, 5), size=(30, 30, 30), axes=("x", "y", "t"))
bbox.chunk((16, 16, 16))
# before: 120 chunks of size (16, 16, 1) — the 16 for t is ignored, t steps by 1
# now:    ValueError, pass VecInt(16, 16, 16, axes=("x", "y", "t")) to step t by 16
```

Boxes with four or more axes are unambiguous and unchanged, and nothing in the library builds a three-non-xyz-axis box.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -18,7 +18,6 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Added `NDBoundingBox.iter_chunk_starts` / `BoundingBox.iter_chunk_starts`, a fast, allocation-free variant of `chunk()` that yields the raw integer topleft coordinate of each chunk (no per-chunk `Vec3Int`/`BoundingBox` allocation). Also added `iter_overlapping_grid_cells`, which yields the toplefts of the origin-aligned grid cells that overlap the box (optionally clipped to another box). [#1489](https://github.com/scalableminds/webknossos-libs/pull/1489)
 
 ### Changed
-- `BoundingBox.chunk` / `NDBoundingBox.chunk` now share their chunk-grid math with the new `iter_chunk_starts` iterator. Their output is unchanged. [#1489](https://github.com/scalableminds/webknossos-libs/pull/1489)
 
 ### Fixed
 
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -15,8 +15,10 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 ### Breaking Changes
 
 ### Added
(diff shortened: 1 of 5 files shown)


normanrz:
I wonder if we had the same practical win, if we just added `BoundingBox.iter_chunk_starts() -> Iterator[tuple[int, int, int]]`. That would be a much smaller change.

Tobias314:
> I wonder if we had the same practical win, if we just added `BoundingBox.iter_chunk_starts() -> Iterator[tuple[int, int, int]]`. That would be a much smaller change.

Fine for me. The only downside might be that NDBoundingBox and BoundingBox share less code and we maintain separate chunking implementations for both. But that is currently also the case.

Actually, the calling code only strictly needs the `iter_overlapping_grid_cells` function. So if we want to keep changes minimal we could just implement this for BoundingBox. I will do a separate PR so that its easier to compare and decide.

Edit: see https://github.com/scalableminds/webknossos-libs/pull/1494

Tobias314:
Closing this in favor of https://github.com/scalableminds/webknossos-libs/pull/1494

PR #1491, opened 2026-07-30
Add coordinate transformations for local and remote layers

normanrz:
### Description:

Adds `Layer.coordinate_transformations`, which reads and sets the transformations that place a layer into the coordinate space of its dataset, for both local and remote layers. The two kinds are modelled as classes mirroring `CoordinateTransformation.scala`:

- `AffineCoordinateTransformation` — a 4x4 homogeneous matrix
- `ThinPlateSplineCoordinateTransformation` — `source`/`target` landmark arrays

**This also fixes a data-loss bug.** `LayerProperties` had no matching attribute, and the cattrs structure hooks are registered without `forbid_extra_keys`, so `coordinateTransformations` was silently dropped on read and lost on the next write of any dataset that had them set.

Two things came along with it:

- **A builder API,** so transformations can be composed instead of writing 4x4 matrices by hand — `from_rotation`, plus `chain` / `translate` / `scale` / `rotate` / `flip`, all returning new instances. For thin plate splines, `from_pairs` and a `pairs` property to work with `(source, target)` pairs rather than two parallel arrays.
- **`Vec3Float` in `webknossos.geometry`,** the float counterpart to `Vec3Int` (the class + `Vec3FloatLike`). `tuple[float, float, float]` had been redeclared as a local alias in seven modules; they now share one. The float-valued entry points then follow the same accept-`-Like`-and-coerce rule the library already applies to `Vec3IntLike` in 186 signatures.

Also fixes `add_layer_as_copy` / `copy_dataset`, which dropped both the new transformations and the pre-existing `default_view_configuration` of the copied layers.

### Issues:
- fixes #...

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

<details>

### Conventions I had to pin down

**Matrix layout** is row-major with the translation in the last column (`matrix[row][column]`). I verified this against the frontend rather than assuming: `dataset_layer_transformation_accessor.ts` reads a z-rotation's `sin` at `matrix[1][0]`, and `transformPointUnscaled` transposes the row-flattened array back into a column-major `M4x4` before applying it. So `[[1,0,0,10],[0,1,0,20],[0,0,1,30],[0,0,0,1]]` translates by `(10, 20, 30)`.

**Rotations take degrees**, matching WEBKNOSSOS (`rotationInDegrees`, `DatasetViewConfiguration.rotation`) rather than numpy's radians default. Their orientation satisfies the frontend's own lookup table (`sinusLocationOfRotationInMatrix`), so a transformation built here reads back correctly in the dataset settings UI. There is a test asserting `from_rotation("z", 90).matrix[1][0] == 1.0`.

**`chain` applies the receiver first**, so `identity().rotate("z", 90).translate((5,0,0))` rotates and then translates, reading left-to-right in execution order. Both directions are pinned by tests, since the opposite convention is equally defensible and silently inverting it would be nasty.

**The transformations are immutable values.** Their arrays are copied on construction and marked read-only, and the attributes use `attr.setters.frozen`, so `layer.coordinate_transformations[0].matrix[0, 3] = 9` raises instead of silently editing the layer's in-memory properties and being flushed to disk by the next unrelated write. `__copy__` / `__deepcopy__` return `self`, which keeps that invariant across every copying path — `deepcopy` of a read-only numpy array otherwise hands back a writable one. The getter returns a tuple, so nothing about the result is mutable; the stored properties keep a list, since that is what is serialized, and the setter accepts any sequence.

`attr.setters.frozen` is used rather than `@attr.frozen`, because the latter generates a `__hash__` that raises `TypeError` for the `cmp_using` numpy fields. This way the classes stay honestly unhashable.

### Not to be confused with `_transform_utils`

#1474 added `AffineTransform` / `transform()`, which *resample voxel data*. These classes are pure metadata affecting only rendering. They are deliberately kept independent — no `apply()`, no coupling to the private and still-unstable `_transform_utils`.

⚠️ The tests for these classes nevertheless live in `tests/dataset/test_transform_layer.py`, next to the `transform()` tests. Happy to move them into a `test_coordinate_transformations.py` of their own if that boundary should hold in the tests too.

### Boundaries I drew (happy to move any of these)

- **`_nml` and `api_client/models.py` keep their own `Vector3`.** Both are wire-format layers whose types mirror a file format rather than the domain model, and `_nml` deliberately depends on nothing beyond stdlib + loxun.
- **`Vec3Int.from_vec3_float` keeps its inline annotation** — `vec3_float.py` imports `vec3_int`, so the reverse would be circular.
- **`webknossos.annotation.Vector3` no longer exists.** It was a module attribute, never in `__all__` or the docs and never reachable from the top-level namespace, but `from webknossos.annotation import Vector3` did work before. This is the one externally observable consequence of the sweep.
- **`SegmentationLayerProperties.mappings` is still not copied** by `add_layer_as_copy`, although it is dropped by the same code path as the metadata I fixed. `mappings` names files under the layer directory that the copy does not transfer, so carrying the names over would create dangling references.
- **No mkdocs nav entry for `vec3_float`** — the API doc generator only emits pages for modules defining classes, so that module produces none and an entry would dangle.

### Two behaviour changes from the `Vec3FloatLike` widening

These are the only non-additive parts of the original change:

1. **Voxel sizes given as ints now serialize as floats.** `VoxelSize.factor` used a converter named `float_tpl` that, despite the name, never cast anything — so `voxel_size=(1, 1, 1)` wrote `"factor": [1, 1, 1]`. It now writes `[1.0, 1.0, 1.0]`. WEBKNOSSOS parses both as `Vec3Double`, and I checked the repo: exactly one fixture has a scale (already floats) and exactly one test does JSON equality on properties, so nothing here depends on it. The visible effect is diff noise the first time an existing dataset is re-saved.
2. **Malformed values are rejected at the boundary.** `Node.rotation`, `Annotation.edit_position` and `Annotation.edit_rotation` previously stored whatever they were handed and failed later during serialization; they now raise `ValueError` on construction. Strictly better, but code that got away with e.g. a 2-tuple rotation will now fail earlier.

Worth noting that mypy caught a real defect during this change: `RemoteAnnotation.__init__` was assigning a widened parameter straight to a strict attribute with no conversion, which is exactly the failure mode a `-Like` annotation invites. It is now an explicit `as_vec3_float_or_none` call.

### Naming of the coercion helpers

`parse_*` reads as "build this from a string", which is what it means everywhere else in the package — `parse_bbox`, `parse_mag`, `parse_path`, `parse_voxel_size` and `parse_vec2int` all take a `str`, as do the eleven `_parse(cls, element)` NML readers. The three vector helpers were the odd ones out, taking a `-Like` union and coercing it, and for `Vec3Float` the name also collided with `Vec3Float.from_str`. They are now `as_vec3_float_or_none`, `as_vec_int_or_none` and `as_vec_int`, after `np.asarray` and the `_as_matrix` / `_as_points` converters in `coordinate_transformations.py`. `as_vec3_float_or_none` is no longer re-exported from `webknossos.geometry`, so it is not public API.

They stay module-level functions rather than becoming classmethods, because mypy's attrs plugin only reads the `__init__` parameter type off a converter that is a named function, a type or a lambda. Handing it `Vec3Float.new_or_none` makes the parameter `Any` and silently stops it from checking what callers pass — which is exactly the check that caught the `RemoteAnnotation` defect above.

### Performance of the vector classes (out of scope, say the word and I will split it out)

⚠️ The last two commits touch `Vec3Int` and `VecInt`, which are not otherwise part of this PR. They are easy to move onto their own branch, as they sit at the tip.

Turning `Vec3Float` from a plain tuple into a class costs allocation, which prompted a look at all three vector classes:

- **`__slots__` on `Vec3Float`, `Vec3Int` and `VecInt`.** All three keep their components in a `_data` tuple, so the per-instance `__dict__` was pure overhead. Measured over 200k instances this saves 40 bytes each — 168 → 128 bytes for a `Vec3Float`, 184 → 144 for a `Vec3Int`. Construction time is unchanged; the cost there is the argument dispatch, not the allocation. `Vec3Int` has to declare `__slots__ = ()` explicitly, as a subclass without it gets a `__dict__` back. The one observable consequence is that instances can no longer be weakly referenced; nothing in the package takes weak references to them.
- **A fast path in `Vec3Int.__new__`.** Constructing a `Vec3Int` took 2.2µs against 0.37µs for a `Vec3Float`, although `Vec3Int.from_xyz` needs only 0.15µs — the cost was entirely dispatch. The general path sorts the arguments by axis name (a no-op for the default axes, 0.48µs a time) and then runs a second, overlapping `isinstance` ladder in `VecInt.__new__`: seven `isinstance` calls per construction, five of them against `Iterable`, which goes through the much slower `collections.abc` check. With the same fast path `Vec3Float` already had, `Vec3Int(1, 2, 3)` goes 2.216µs → 0.338µs and `Vec3Int((1, 2, 3))` 1.510µs → 0.394µs. The guard is `type(x) is int` rather than `isinstance`, on purpose: `bool` is a subclass of `int`, so `isinstance` would let `Vec3Int(True, 2, 3)` store `True` instead of `1`.

### Testing

`ruff` and `mypy` are clean. `tests/dataset` + `tests/geometry` + `tests/test_skeleton.py` + `tests/test_annotation.py`: **693 passed, 45 failed, 16 skipped**, with the same count and distribution as on the parent commit. The failures are all environmental: 39 in `test_remote_dataset.py` and 3 in `test_dataset.py` need a live WEBKNOSSOS at `localhost:9000`, 5 in `test_annotation.py` call out to `webknossos.org` for a test-only organization, and `test_from_dicom_images` needs a JVM that my machine does not have.

⚠️ **The two remote tests added to `test_remote_dataset.py` have never been executed** — that file needs the same live server, which I did not have. They follow the file's existing patterns but are unproven; please confirm they pass in CI. They are the only check that the server actually round-trips the nested `List[List[Double]]` matrix format.

</details>


webknossos/tests/geometry/test_vec3_float.py
```diff
@@ -0,0 +1,64 @@
...
+from webknossos import Annotation, Skeleton, VoxelSize
+from webknossos.geometry import (
+    Vec3Int,
+    parse_vec3_float,
+    parse_vec3_float_or_none,
+)
+
+
+def test_parse_vec3_float() -> None:
+    assert parse_vec3_float((1.5, 2, 3)) == (1.5, 2.0, 3.0)
```
fm3: When a function is called `parse` I expect the input to be a string. Is there a specific reason not to use the same pattern as `Vec3Int` with a `__new__` that takes `args: Union["Vec3IntLike", Iterable[str], int]`? Ok I see that Vec3Float is not a class but just a type alias. Is there a benefit to that? And if so, should we do the same for Vec3Int? Also, even if that doesn’t apply, I’d still vote to find a different name for this function

webknossos/webknossos/dataset_properties/coordinate_transformations.py
```diff
@@ -0,0 +1,261 @@
...
+            f"The correspondences must have shape (N, 3), got {points.shape}."
+        )
+    return points
+
+
+class CoordinateTransformation(ABC):
+    """Base class of the coordinate transformations of a layer.
+
+    See `AffineCoordinateTransformation` and `ThinPlateSplineCoordinateTransformation`
+    for the concrete transformations.
```
fm3: I’d remove these two lines, for fear of them becoming outdated. There are already quite a few spots we need to adapt when introducing a new type of coordinate transformation

fm3 (changes requested):
Very nice stuff, looking good! 

Should the transformations classes also get some kind of apply method, actually transforming input points or voxels?

Other than that, just two minor points:

webknossos/webknossos/dataset_properties/coordinate_transformations.py
normanrz:
```suggestion
    """Base class of the coordinate transformations of a layer."""
```

normanrz:
> Should the transformations classes also get some kind of apply method, actually transforming input points or voxels?

I'd like to make a followup where we integrate this with #1474, but I don't know how the API should look like, yet.

webknossos/tests/geometry/test_vec3_float.py
normanrz: `Vec3Float` is now a full class

webknossos/webknossos/dataset/layer/abstract_layer.py
```diff
@@ -274,6 +299,36 @@ def default_view_configuration(
...
+        renders the layer, the voxel data itself is not modified.
+
+        Returns:
+            tuple[CoordinateTransformation, ...]: The transformations, empty if there
+                are none.
+
+        Note:
+            The result is a tuple of immutable transformations, so nothing about it can
+            be changed in place. Assign a new sequence to change the transformations of
+            this layer.
```
normanrz:
```suggestion
                are none.
```

webknossos/tests/geometry/test_vec3_float.py
normanrz: `parse_vec3_float_or_none` -> `as_vec3_float_or_none`

normanrz:
I also had some micro-optimizations done to `Vec{Int,3Int,3Float}`, e.g. slots and fast-paths for common object constructions.

webknossos/webknossos/administration/task.py
```diff
@@ -262,7 +262,7 @@ class Task:
     task_type: TaskType
     experience: TaskExperience
     edit_position: Vec3Int
-    edit_rotation: tuple[float, float, float]
+    edit_rotation: Vec3Float = attr.field(converter=Vec3Float)
```
fm3, line 265: why is the same not needed for Vec3Int in the line above?

fm3 (approved):
I’m liking this much better! Just one question popped up

PR #1492, opened 2026-08-03
Support RemoteAccessMode per mag

normanrz:
### Description:
The access mode of a remote dataset used to be a dataset-wide choice made in `RemoteDataset.open`. It collapsed in `__init__` into `zarr_streaming_path` plus `_use_zarr_streaming`, after which `ZARR_STREAMING` and `PROXY_PATH` were indistinguishable — and it decided considerably more than the mag paths.

Individual mags can now pick their own mode:

```python
ds = wk.RemoteDataset.open(url)
layer = ds.get_layer("color")

mag1 = layer.get_mag(1, access_mode=wk.RemoteAccessMode.DIRECT_PATH)
mag2 = layer.get_mag(2, access_mode=wk.RemoteAccessMode.PROXY_PATH)
```

`RemoteDataset.open(access_mode=...)` still sets the default that all mags inherit. Passing `access_mode` to `get_mag` returns an uncached `RemoteMagView`, so `layer.mags` keeps reflecting the dataset default.

This is the groundwork for a follow-up where the client determines on its own whether `direct_path` is reachable and falls back to `proxy_path` / `zarr_streaming` per mag.

`RemoteDataset`s can now be edited (e.g. adding layers) no matter the access_mode they were opened in.

#### What `access_mode` actually controlled

Verified against the server responses recorded in the cassettes that were removed in 429c0c9a:

| | `DIRECT_PATH` | `PROXY_PATH` | `ZARR_STREAMING` |
|---|---|---|---|
| properties source | api `dataSource` | `{prefix}/datasets/{id}/proxy/…` | `{prefix}/zarr/{id}/…` |
| `layer.data_format` | underlying (`zarr3`) | underlying (`zarr3`) | **always `zarr`** |
| mag path in properties | absolute | `./layers/color/mags/1` | `./color/1` |
| attachments in properties | present, absolute | — | **omitted entirely** |
| layers writable | yes | no | no |
| annotations | rejected | rejected | only supported mode |
| unusable data source | `RuntimeError` | tolerated | tolerated |

`PROXY_PATH` is a byte-level proxy of the direct path, so it preserves the underlying format; `ZARR_STREAMING` re-serves the data as Zarr v2.

#### Direct paths are now the stored ones

`RemoteDataset` loads the api `dataSource` as its properties whenever it is usable, so `MagViewProperties.path` always holds the underlying direct path and `mag.direct_path` is reachable no matter how the dataset was opened. The zarr streaming and proxy paths are computed from the datastore url (layouts live in one place in `remote_access_mode.py`) and are persisted nowhere.

The served `datasource-properties.json` remains the source in the two cases where no direct paths exist: annotations, whose volume layers are not part of the dataset's data source, and datasets with an `ApiUnusableDataSource`. Requesting an unavailable mode raises with an explanatory message.

Since streaming mode no longer needs to fetch the served properties, it now makes one HTTP request fewer than before.

#### Other changes

- `Dataset.add_layer_as_ref` derives the new layer's `data_format` from the mags it references rather than from the foreign layer. Referencing a zarr-streamed remote layer otherwise built a `zarr3` local layer pointing at `zarr` arrays — three existing tests caught this.
- `RemoteDataset.reopen` now keeps the `read_only` flag of the original dataset.
- The read-only hint on `RemoteLayer` no longer points at the deprecated `use_zarr_streaming=False`.

### Breaking changes
- In `ZARR_STREAMING` mode, `RemoteLayer.data_format` now reports the format of the underlying storage (e.g. `zarr3`) instead of `zarr`. Use `RemoteMagView.data_format` for the format a given mag actually serves.
- The arguments of `RemoteDataset.__init__` changed. Use `RemoteDataset.open`, as its docstring already directs.
- `RemoteDataset.zarr_streaming_path` is deprecated in favour of the per-mag paths. It still returns the same values.
- `RemoteSegmentationLayer.attachments` is now populated in zarr streaming mode, where it used to be empty.

### Notes for the reviewer
- **`with_access_mode` sits on `RemoteAttachments`, not on individual `Attachment` objects.** An `Attachment` has no back-reference to its layer, so it cannot resolve a path on its own. Attachments default to direct-path resolution in every mode, which is why they are visible under streaming for the first time.
- **The proxy route for attachments** is `{proxy_base}/layers/{layer}/attachments/{type}/{name}`. This is now confirmed against the backend and covered by `test_attachments_are_available_in_every_access_mode`. The mag routes for both proxy and streaming are also confirmed from recorded responses.
- The abstract hook `AbstractLayer._determine_read_only_and_path_for_mag` became `_setup_mag_from_properties`, so remote layers can construct their own view type. This touches local `Layer` too.

### Issues:
- fixes #...

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -333,6 +314,115 @@ def reopen(self, *, access_mode: RemoteAccessMode) -> "RemoteDataset":
...
+        assert mag_properties.path is not None, (
+            f"Remote mags must have a path: {mag_properties}"
+        )
+        return enrich_path(mag_properties.path, base_path)
+
+    def _attachment_path(
+        self, attachment_properties: AttachmentProperties, access_mode: RemoteAccessMode
+    ) -> UPath:
+        """Resolves the path of an attachment for the given access mode."""
+        if access_mode == RemoteAccessMode.DIRECT_PATH:
```
normanrz: attachments are available through PROXY_PATH

webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -143,20 +137,29 @@ class RemoteDataset(AbstractDataset[RemoteLayer, RemoteSegmentationLayer]):
...
+            annotation_id: Optional annotation id, if the dataset is viewed through an annotation
             context: Context manager for WEBKNOSSOS connection
+            read_only: Whether metadata of the dataset may be modified
+            access_mode: The default access mode that all mags of this dataset inherit
+            datastore_url_prefix: Base url of the datastore api, used to compute
+                zarr streaming and proxy paths
+            dataset_properties: Already loaded properties of the remote dataset. Loaded
+                on demand if omitted.
+            properties_are_direct: Whether `dataset_properties` stems from the api data
+                source, i.e. whether it contains direct paths.
```
normanrz: What is this needed for?

webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -333,6 +314,115 @@ def reopen(self, *, access_mode: RemoteAccessMode) -> "RemoteDataset":
...
+        """The default access mode of this dataset.
+
+        All mags of this dataset are accessed this way unless a different mode is
+        requested explicitly, e.g. via `RemoteLayer.get_mag(mag, access_mode=...)`.
+        """
+        return self._access_mode
+
+    @property
+    def _use_zarr_streaming(self) -> bool:
+        return self._access_mode != RemoteAccessMode.DIRECT_PATH
```
normanrz: why do we still need this?

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
```diff
@@ -0,0 +1,98 @@
...
+
+    @property
+    def proxy_path(self) -> UPath:
+        """The path that proxies the underlying storage through the WEBKNOSSOS datastore."""
+        return self._path_for(RemoteAccessMode.PROXY_PATH)
+
+    @property
+    def zarr_streaming_path(self) -> UPath:
+        """The path where the WEBKNOSSOS datastore re-serves this mag as Zarr."""
+        return self._path_for(RemoteAccessMode.ZARR_STREAMING)
```
normanrz: Why are these properties necessary?

webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -143,20 +137,29 @@ class RemoteDataset(AbstractDataset[RemoteLayer, RemoteSegmentationLayer]):
...
         """Initialize a remote dataset instance.
 
         Args:
-            zarr_streaming_path: Path to the zarr streaming directory
-            dataset_properties: Properties of the remote dataset
             dataset_id: dataset id of the remote dataset
+            annotation_id: Optional annotation id, if the dataset is viewed through an annotation
             context: Context manager for WEBKNOSSOS connection
+            read_only: Whether metadata of the dataset may be modified
+            access_mode: The default access mode that all mags of this dataset inherit
```
normanrz: Why is this not None by default?

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Fixed — implemented, using a computed proxy route. I checked against the local WEBKNOSSOS server: the proxy does serve attachments, at `layers/{layer}/attachments/{type_name}/{name}` (e.g. `.../proxy/layers/segmentation/attachments/agglomerate/map_all/zarr.json`), confirmed by fetching a live proxied `datasource-properties.json` and the attachment's `zarr.json` through it. `ZARR_STREAMING` still isn't supported for attachments, since they're not part of the served `datasource-properties.json` at all (verified the same way). Added a test asserting the proxy path resolves and serves the same data_format as the direct path.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Good catch — it wasn't needed. `properties_are_direct` was always fully determined by whether `dataset_properties` was passed (`open()` set them in lockstep in all three branches), so I removed the parameter and derive `self._properties_are_direct = dataset_properties is not None` inside `__init__` instead.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: It's still needed as a helper (it's read from 3 places: layer read-only, new-layer read-only in `_apply_server_dataset_properties`, and the write-guard in `_save_dataset_properties_impl`), but the name was stale — it's true for `PROXY_PATH` too, not just `ZARR_STREAMING`, ever since proxy support was added. Renamed to `_metadata_is_read_only` with a docstring, and updated the two messages that said "zarr streaming" specifically to name the actual access mode.

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
normanrz: They let a caller inspect all three candidate paths for a mag without constructing separate `RemoteMagView` instances (each of which does its own path resolution) — which is exactly the input the planned follow-up needs (probing which of direct/proxy/streaming is reachable, per mag, and falling back automatically). Added a note to `direct_path`'s docstring explaining this, and while I was at it replaced the try/except there with a direct check of `RemoteDataset._properties_are_direct` instead of catching `ValueError`.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Intentional — `__init__` is documented as internal ("do not call directly, use `RemoteDataset.open()`"), and `open()` is the only caller. `open()` already does the None-defaulting (to `ZARR_STREAMING`) and the deprecated `use_zarr_streaming` bridging before it calls `cls(...)`, so by the time `__init__` runs, `access_mode` is always a concrete, resolved value. Keeping it required here means `__init__` can't construct an instance in an ambiguous "mode not yet decided" state — that resolution happens exactly once, at the public entry point.

normanrz:
Pushed changes addressing all five review comments:

- **Attachments via `PROXY_PATH`**: implemented, using `layers/{layer}/attachments/{type_name}/{name}`, confirmed against the local WEBKNOSSOS server. `ZARR_STREAMING` remains unsupported for attachments (they're not in the served `datasource-properties.json`).
- **`properties_are_direct` param**: removed — it was fully determined by whether `dataset_properties` was passed, so it's now derived inside `__init__`.
- **`_use_zarr_streaming`**: renamed to `_metadata_is_read_only` with a docstring, since it covers `PROXY_PATH` too and the old name was misleading; updated the two error messages that referenced "zarr streaming" specifically.
- **`RemoteMagView.direct_path`/`proxy_path`/`zarr_streaming_path`**: documented why they exist (letting a caller compare all three candidate paths for a mag without extra instances — the input the planned auto-fallback needs), and replaced the `direct_path` try/except with a direct check of `_properties_are_direct`.
- **`access_mode` with no default in `__init__`**: kept as-is with rationale — it's an internal constructor only called from `open()`, which already resolves the default/deprecation bridging before calling it.

Also added a test asserting attachments resolve correctly through `PROXY_PATH` and serve the same data as `DIRECT_PATH`. Full `test_remote_dataset.py` + `test_annotation.py` + attachment suites pass (101/101) against the local server; ruff and mypy clean.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Why does the metadata have to be read-only for DIRECT_PATH?

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
normanrz: What about making this a property: paths = dict[RemoteAccessMode, UPath]?

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Does it still make sense that the default is ZARR_STREAMING? Can it perform all actions (attachments, metadata edits, etc) that way?

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Good catch — it didn't have to be, that was a bug. `_initialize_layer_from_properties` was unconditionally overwriting the incoming `read_only` argument with `_metadata_is_read_only`, so `RemoteDataset.open(..., read_only=True, access_mode=RemoteAccessMode.DIRECT_PATH)` silently left layers writable — the explicit `read_only=True` was discarded. Fixed to combine them (`read_only or self._metadata_is_read_only`) and added a test (`test_read_only_forces_layers_read_only_with_direct_path`) covering both directions.

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
normanrz: Done — replaced `direct_path`/`proxy_path`/`zarr_streaming_path` with a single `paths: dict[RemoteAccessMode, UPath]`. A mode is simply omitted from the dict when it's unavailable for that mag, which cleanly subsumes both the "direct_path is None" case and the "PROXY_PATH unavailable for annotations" case without needing a nullable return type or a bespoke exception per property. Updated the tests and the docs example.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: I'd keep it — that tradeoff predates this PR (only `DIRECT_PATH` has ever supported metadata writes; `ZARR_STREAMING` was already read-only-for-writes before proxy support existed) and isn't something this change alters, so I don't think it should be decided as a side effect of making access mode per-mag.

To answer directly: under the `ZARR_STREAMING` default, reading works for everything — layers, mags, and now attachments too (they used to be empty under streaming before this PR, since the served `datasource-properties.json` omits them; now they resolve via their direct path regardless of the dataset's default). What doesn't work under `ZARR_STREAMING` (or `PROXY_PATH`) is *writing* — metadata edits and attachment/mag mutations, which require `DIRECT_PATH`. That split (works everywhere for reads, needs `DIRECT_PATH` for writes) seems like a reasonable default to me: it's the one that works without requiring the caller to have credentials for the underlying storage, and users who need to write already have to opt into `DIRECT_PATH` explicitly and know they need access to that storage. Happy to discuss changing the default separately if you'd like, but I'd treat it as its own decision rather than bundling it into this PR.

normanrz:
Pushed another round addressing the follow-up comments:

- **Fixed a real bug** the read-only question surfaced: `RemoteDataset.open(..., read_only=True, access_mode=RemoteAccessMode.DIRECT_PATH)` silently ignored `read_only` and left layers writable, because layer construction unconditionally overwrote the passed-in flag instead of combining it with the access-mode-derived one. Fixed and covered by a new test.
- **Replaced the three path properties with `RemoteMagView.paths: dict[RemoteAccessMode, UPath]`**, as suggested — cleaner than a nullable `direct_path` plus two non-nullable siblings, and it uniformly handles "mode unavailable for this mag" by omitting the key. Updated tests and docs.
- **Replied on the default-access-mode question** rather than changing it — that tradeoff (only `DIRECT_PATH` supports writes) predates this PR and reading works everywhere including attachments now, so I'd treat changing the default as a separate decision.

102/102 tests pass (remote dataset, annotation, attachments) against the local server; ruff and mypy clean.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: can we make it so that both zarr and proxy modes get to do writing?

normanrz: Yes — implemented. It turns out the technical reason writes were blocked for `ZARR_STREAMING`/`PROXY_PATH` had already disappeared earlier in this PR: since `RemoteDataset` now loads its properties from the api data source whenever one is usable (regardless of access mode), the properties backing a `ZARR_STREAMING`- or `PROXY_PATH`-opened dataset are now the *same* api-sourced object as a `DIRECT_PATH`-opened one. Writing metadata is just `client.dataset_update` PUTting that object back to the server — an API call, unrelated to how the array bytes are read.

So I changed the write-gate from "is the access mode DIRECT_PATH" to "do the properties stem from the api data source" (`_properties_are_direct`). That is a strictly better characterization: it correctly keeps writes blocked for the two cases where it is genuinely unsafe — an annotation's volume layers, and a dataset with an unusable data source — both of which fall back to a served (non-api) datasource-properties.json, so writing would not be consistent with what is being read. Under every other case (now the common case for all three access modes), metadata writes work.

This also affects the layer-level read_only flag the same way, which gates layer-level mutations like add_mag_as_ref and attachment management. Updated the two existing tests that encoded the old restriction and added a new one covering the annotation case that is still correctly blocked.

normanrz:
Pushed a merge from master (only `Changelog.md` conflicted — resolved by keeping this PR's entries and dropping two bullets that had already shipped in v3.6.0) plus one more substantive change from the follow-up discussion:

**Metadata writes now work under `ZARR_STREAMING` and `PROXY_PATH`, not just `DIRECT_PATH`.** The technical reason they were blocked had already disappeared earlier in this PR — since `RemoteDataset` loads its properties from the api data source whenever one is usable, regardless of access mode, all three modes now share the same api-sourced properties object. The write-gate is now "do the properties stem from the api data source" (`_properties_are_direct`) rather than "is access mode DIRECT_PATH", which correctly still blocks writes for the two cases where properties come from a served (non-api) JSON instead: an annotation's volume layers, and a dataset with an unusable data source.

Also fixed `SegmentStatisticsAttachment` (added on master) to route through the same `_attachment_path`/proxy mechanism as the other attachment types, since it had bypassed it.

Updated the two existing tests that encoded the old restriction, added a new one for the still-blocked annotation case, and re-ran the full remote/annotation/attachment/layer/transform suites (128 tests) plus ruff/mypy — all green.

normanrz:
Merge conflict resolved and pushed (`2e11af05`). GitHub now reports the PR as mergeable; CI is running against the merged state.

normanrz:
Merged `release-4.0` in as well (`cb78bdd9`). One conflict in `Changelog.md` — release-4.0 rewrote the Unreleased section for its Bio-Formats removal / image-conversion overhaul, so I interleaved both sets of entries under the shared Breaking/Added/Changed/Fixed headers rather than picking one side.

Verified the auto-merged intersection points (`abstract_layer.py`'s `_setup_mag_from_properties` hook, `dataset.py`'s `add_layer_as_ref` data-format fix, the export lists) survived intact. Re-ran the remote/annotation/attachment/layer/transform suites (128 tests) plus the image-conversion suites (167 tests, previously blocked locally by a missing JVM for Bio-Formats — moot now that release-4.0 removed it) — all pass. Ruff and mypy clean.

One cleanup note for the log: my working tree had a leftover extracted `WT1_wkw/` test-data directory from an earlier local test run; release-4.0 deletes the `WT1_wkw.tar.gz` fixture it came from with no replacement, so `git add -A` would have resurrected and committed that stale extraction as new files. Caught and removed before committing.

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
```diff
@@ -0,0 +1,95 @@
...
+            data_format=data_format_for_access_mode(access_mode, layer.data_format),
+        )
+
+    @property
+    def access_mode(self) -> RemoteAccessMode:
+        """How the data of this mag is accessed."""
+        return self._access_mode
+
+    @property
+    def paths(self) -> dict[RemoteAccessMode, UPath]:
```
normanrz: make it a readonly `Mapping[RemoteAccessMode, UPath]`

webknossos/webknossos/dataset/layer/remote_layer.py
```diff
@@ -62,11 +67,48 @@ def as_segmentation_layer(self) -> "RemoteSegmentationLayer":
...
+            IndexError: If specified magnification does not exist
+            ValueError: If the requested access mode is not available for this dataset
+
+        Examples:
+            ```
+            mag1 = layer.get_mag(1, access_mode=RemoteAccessMode.DIRECT_PATH)
+            mag2 = layer.get_mag(2, access_mode=RemoteAccessMode.PROXY_PATH)
+            ```
+        """
+        mag_view = super().get_mag(mag)
```
normanrz: Doesn't this return a regular `MagView`?

webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -149,20 +147,28 @@ class RemoteDataset(AbstractDataset[RemoteLayer, RemoteSegmentationLayer]):
...
+            datastore_url_prefix: Base url of the datastore api, used to compute
```
normanrz: Instead of storing the datastore_url_prefix in the RemoteDataset, could that not be retrieved from the relevant api client? Maybe the api client should also hold the paths for computing the mag and attachment paths so that we can model them based on the API versions?

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
```diff
@@ -0,0 +1,95 @@
...
+from ....dataset_properties import MagViewProperties
+from ....geometry import Mag
+from ...remote_access_mode import RemoteAccessMode, data_format_for_access_mode
+from .mag_view import MagView
+
+if TYPE_CHECKING:
+    from ..remote_layer import RemoteLayer
+
+
+class RemoteMagView(MagView["RemoteLayer"]):
```
normanrz, line 16: Please add a `with_access_mode()` method like on `RemoteAttachments`.

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
normanrz: Done — now returns a read-only Mapping (a MappingProxyType wrapper), not a plain dict.

webknossos/webknossos/dataset/layer/remote_layer.py
normanrz: At runtime it was correctly a RemoteMagView (super().get_mag(mag) does self.mags[mag], and RemoteLayer.mags is overridden to hold RemoteMagView instances), but you're right that it's typed as the base MagView, since AbstractLayer.get_mag()'s signature just says -> MagView. That's why the isinstance assert was there, purely to narrow the type for mypy.

Replaced it with a direct self.mags[mag] lookup (RemoteLayer.mags is already typed as Mapping[Mag, RemoteMagView]), so it's correctly typed without needing a runtime assert at all.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Done, both parts.

RemoteDataset now stores the datastore's base url instead of a precomputed url_prefix string, and looks up the DatastoreApiClient (hence url_prefix) from the context on demand — the same pattern already used for the auth token in _base_path, for the same reason (don't freeze something context-derived at open() time).

And moved the actual route construction (zarr streaming / proxy base urls, mag paths, attachment paths) onto DatastoreApiClient as methods, e.g. zarr_streaming_dataset_url, proxy_mag_path, proxy_attachment_path. They replace the standalone mag_url_suffix/attachment_url_suffix functions that used to live in remote_access_mode.py. Now a route change for a future api version only has to be reflected on the client, e.g. by overriding on a version-specific subclass like DatastoreApiClientV13 (which already exists for a couple of upload routes) — RemoteDataset no longer needs to know the route layout itself, just which client method to call for which access mode.

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
normanrz: Added RemoteMagView.with_access_mode(), mirroring RemoteAttachments.with_access_mode() exactly. RemoteLayer.get_mag() now delegates to it for a non-default access mode instead of constructing a RemoteMagView directly.

normanrz:
Pushed `20dca240` addressing the four new comments:

- **`RemoteMagView.paths` is now a read-only Mapping** (`MappingProxyType`), not a plain `dict`.
- **`RemoteLayer.get_mag()` no longer relies on `super().get_mag()` + an `isinstance` assert to work around the base class's untyped return.** It looks up `self.mags[mag]` directly, which is already correctly typed as `Mapping[Mag, RemoteMagView]`.
- **`RemoteDataset` no longer stores a precomputed `datastore_url_prefix` string.** It stores the datastore's base url and looks up the `DatastoreApiClient` (and thus `url_prefix`) from the context on demand — same pattern already used for the auth token, same reasoning (don't freeze something context-derived at `open()` time).
- **Moved the actual mag/attachment route construction onto `DatastoreApiClient`** (`zarr_streaming_dataset_url`, `proxy_mag_path`, `proxy_attachment_path`, etc.), replacing the standalone functions that lived in `remote_access_mode.py`. A route change for a future api version now only needs to be reflected there, overridable per version the same way `DatastoreApiClientV13` already overrides a couple of upload routes.
- **Added `RemoteMagView.with_access_mode()`**, mirroring `RemoteAttachments.with_access_mode()`.

Added a unit test for the new `DatastoreApiClient` path-building methods (`tests/test_api_client.py`) and extended the per-mag test with coverage for `paths` immutability and `with_access_mode()` equivalence to `get_mag(access_mode=...)`. Re-ran the remote/annotation/attachment/layer/transform/api-client suites (136 tests) — all pass; ruff and mypy clean.

normanrz:
Manually tested with:
```python
import webknossos as wk

target = wk.RemoteDataset.open("https://webknossos.org/datasets/idr0062A-69f20ac0010000ea07806ac0/view#135,137,118,0,1.3")
target.layers # Dapi is in there

target.get_layer("Dapi").get_mag(1).read() # zarr streaming
target.get_layer("Dapi").get_mag(1,access_mode=wk.RemoteAccessMode.PROXY_PATH).read()
target.get_layer("Dapi").get_mag(1,access_mode=wk.RemoteAccessMode.DIRECT_PATH).read()

source = wk.RemoteDataset.open("https://webknossos.org/datasets/6001240-69134faa010000da00e2c05b/view#135,137,118,0,1.3")
target.add_layer_as_ref(source.get_layer("Dapi"), new_layer_name="Dapi2")
target.layers # Dapi2 is in there
```

normanrz:
Pushed `182ecf29`, fixing a correctness bug reported from manual testing against a real external dataset (idr0062A on webknossos.org):

**The bug**: `layer.get_mag(1, access_mode=RemoteAccessMode.ZARR_STREAMING).read()` silently returned all zeros, while `PROXY_PATH`/`DIRECT_PATH` returned real data.

**Root cause**: this PR made `RemoteDataset` always load `self._properties` from the api data source whenever usable, regardless of access mode, so direct paths/attachments would always be discoverable. For this dataset, the api's bounding box/axis order for the affected layer genuinely disagrees with what the `ZARR_STREAMING` endpoint itself serves (`topleft.c=1` vs `0`, swapped x/z axis order) — likely because the datastore re-indexes a per-channel-split layer's own served array independently of how the api describes the channel's position in the original multi-channel source. Reading with the wrong bounding box against the right array requested an out-of-range region, and zarr's fill-value convention silently turned that into zeros instead of an error.

**Fix**: `self._properties` (bounding box, axis order, mag paths) is now always sourced from the endpoint matching the dataset's own access mode, exactly as before this PR. Direct paths and attachments stay discoverable regardless of access mode via a separate, lazily-fetched api lookup — but that lookup is used only for path/attachment resolution, never for bounding-box/axis metadata. `RemoteLayer.data_format` is resolved the same way (this also fixed a second bug the local suite caught: a `DIRECT_PATH` mag picking the wrong array-reader class for an already-converted zarr3 dataset).

**Necessary reversions**: two conveniences added earlier in this PR depended on the now-disproven assumption that api and served properties always agree, so they're reverted: metadata writes and attachment listing again require/prefer `DIRECT_PATH`-sourced properties rather than working under every access mode. (Attachments remain visible under every mode — they were never part of the served properties to begin with, so that part was always safe. Only the metadata-*write* capability under `ZARR_STREAMING`/`PROXY_PATH` is reverted, since writing a served representation back as the api data source risked corrupting it.)

Verified directly against the reported dataset (now returns real data, matching pre-PR behavior exactly) and against the full local suite (136 tests, remote/annotation/attachments/layer/transform/api-client) — all pass. Ruff and mypy clean.

normanrz:
Pushed `5fa62409` — this fixes a deeper version of the same class of bug my previous commit only partially addressed.

**What was still wrong**: the earlier fix restored correct bounding-box/axis-order for a dataset's own *default* access mode, but a mag read via a *different* mode (`layer.get_mag(mag, access_mode=...)`, this PR's whole point) still reused the default mode's bounding box. For a real dataset (`idr0062A`) reported manually, this meant `ZARR_STREAMING`, `PROXY_PATH` and `DIRECT_PATH` returned *different pixel values* for the same mag — its "Dapi" layer is one channel split out of a two-channel raw source, and the datastore's `ZARR_STREAMING` representation already slices out just that channel (so its bounding box correctly starts at channel 0), while the raw/proxied array still has both channels (so its bounding box correctly starts at channel 1, per the api's `channelIndex` metadata). Mixing one mode's bounding box against a different mode's array silently read the wrong channel.

**Fix**: every property needed to construct a `RemoteMagView` — path, bounding box, axis order, `data_format` — now always comes from a single document: the `datasource-properties.json` actually served by the access mode being constructed (or the api data source for `DIRECT_PATH`), never mixed across modes. This replaces the formula-based path computation from the previous round (which couldn't account for per-dataset served-representation differences, e.g. channel splitting) with fetching each mode's own document on first use, cached per (dataset, mode). Reading through the dataset's own default mode still costs no extra request; a non-default mode now costs one extra request the first time it's used.

**Verified**: all three access modes now return byte-for-byte identical data for `idr0062A`'s "Dapi" layer (confirmed by transposing to a common axis order and comparing — the axis-order difference itself is legitimate/pre-existing, unrelated to this bug). Added `test_explored_remote_access_modes_return_same_data`, which calls `RemoteDataset.explore_and_add_remote` on the actual reported dataset URL and asserts all three modes agree — this is a live regression test against the real external server, not a synthetic fixture.

Also reverted the `RemoteLayer.data_format` override from the previous round (no longer needed — `RemoteMagView.data_format` is now always self-correct per mode) and removed the now-dead `DatastoreApiClient` path-formula methods.

Full local suite (137 tests, including the new external regression test) passes; ruff/mypy clean.

webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -264,66 +271,44 @@ def open(
...
+                dataset_id=dataset_id,
+                annotation_id=annotation_id,
+                context=context_manager,
+                read_only=read_only,
+                access_mode=access_mode,
+                datastore_url=datastore_url,
+                dataset_properties=dataset_properties,
+            )
 
     def reopen(self, *, access_mode: RemoteAccessMode) -> "RemoteDataset":
```
normanrz, line 291: This should be deprecated in favor of a new `with_access_mode` function

webknossos/webknossos/dataset/remote_dataset.py
normanrz, line 1: The docstrings in this module are way too long. Shorten them. For private methods, they should only have a single sentence each.

webknossos/webknossos/dataset/layer/remote_layer.py
```diff
@@ -37,17 +39,26 @@ def __init__(
...
-        read_only = True
-        return read_only, mag_path
+        mag = Mag(mag_properties.mag)
+        self._assert_mag_does_not_exist_yet(mag)
+        try:
+            # Remote mags are always read-only, regardless of the access mode.
+            mag_view = RemoteMagView(
+                self, mag, access_mode=self._dataset.access_mode, read_only=True
+            )
+        except ValueError:
```
normanrz, line 58: Silently continuing here seems dangerous

webknossos/Changelog.md
normanrz, line 1: The changelog is way too detailed, which makes the users lose track of the core change: The access mode can now be set on the level of mags and attachments. This allows that remote datasets can be edited (e.g. adding/removing layers/mags) no matter the access mode.

webknossos/webknossos/dataset/remote_dataset.py
normanrz: [Claude] Added `RemoteDataset.with_access_mode()`, mirroring `RemoteMagView`/`RemoteAttachments.with_access_mode()`. Unlike `reopen()`, it reuses the already-fetched properties/context instead of making a new request. Kept `reopen()` itself as-is (not deprecated) per follow-up discussion, since it's still useful for an actual re-fetch from the server. (`c25e6362`)

webknossos/webknossos/dataset/remote_dataset.py
normanrz: [Claude] Shortened the docstrings this PR introduced in this module, especially the private methods (`_metadata_is_read_only`, `_get_datastore_api_client`, `_get_dataset_properties_for_mode`, `_base_path`, `_mag_path`, `_attachment_path`, `_save_dataset_properties_impl`), down to a single sentence each, and trimmed `__init__`/`with_access_mode`. (`c25e6362`)

webknossos/webknossos/dataset/layer/remote_layer.py
normanrz: [Claude] Now emits a `UserWarning` naming the mag/layer before skipping it, instead of silently dropping it from `layer.mags`. (`c25e6362`)

webknossos/Changelog.md
normanrz: [Claude] Condensed the changelog entries for this PR down to a two-sentence summary of the core change plus the `with_access_mode`/`zarr_streaming_path` API notes and the one genuine pre-existing bugfix (`RemoteDataset.reopen` now keeps `read_only`). Dropped the internal dev-time bugfix narrative for bugs that were introduced and fixed within this same unreleased PR, since they never shipped. (`c25e6362`)

webknossos/webknossos/client/api_client/datastore_api_client.py
```diff
@@ -39,6 +39,23 @@ def url_prefix(self) -> str:
...
         return f"{self.url_prefix}/datasets/upload/dataset"
 
+    # Routes to the datasource-properties.json documents that describe a dataset's
+    # served representation for a given access mode. Kept here, so that a route change
+    # for a future api version only has to be reflected in one place. Everything within
+    # such a document (mag/attachment paths, bounding box, axis order) always comes
+    # from the document itself, never computed: the datastore can derive a materially
+    # different representation than another mode's document describes (e.g. splitting
+    # a multi-channel source into single-channel layers), so paths within a document
+    # are not safely predictable from a formula.
```
fm3:
```suggestion
```
tbh I don’t think this comment adds much. This code is simple enough to understand from reading, and someone wonders why this exists here, they can check the callers.

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
```diff
@@ -0,0 +1,120 @@
...
+        """
+        Do not use this constructor manually. Instead use `RemoteLayer.get_mag()`.
+        """
+        self._access_mode = access_mode
+        # The properties document actually served by access_mode is the only source
+        # that reliably describes what it serves: the datastore may derive a
+        # materially different representation for one mode than another mode's
+        # document describes (e.g. splitting a multi-channel source into
+        # single-channel layers), so bounding_box, data_format and the mag path all
+        # come from this single document, never mixed across modes.
```
fm3: comment seems a little convoluted, should probably be rewritten by a human

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
```diff
@@ -0,0 +1,120 @@
...
+        if it is not available for this mag (e.g. `DIRECT_PATH` when the server does not
+        expose it, or `PROXY_PATH`/`ZARR_STREAMING` for an annotation's volume layers).
+        """
+        result: dict[RemoteAccessMode, UPath] = {}
+        for access_mode in RemoteAccessMode:
+            try:
+                result[access_mode] = self.layer.dataset._mag_path(
+                    self.layer.name, self._mag, access_mode
+                )
+            except ValueError:  # noqa: PERF203 only 3 iterations, clarity wins here
```
fm3, line 111: where are the 3 iterations here?

webknossos/webknossos/dataset/layer/abstract_layer.py
```diff
@@ -100,10 +97,14 @@ def as_segmentation_layer(self) -> "AbstractSegmentationLayer":
...
-    ) -> tuple[bool, UPath]:
-        pass
+    def _setup_mag_from_properties(
+        self, mag_properties: MagViewProperties, read_only: bool
+    ) -> None:
+        """Initialize a single mag when opening the layer.
+
+        Subclasses resolve the mag's path (and, for remote layers, its access mode)
+        and register the resulting `MagView` in `self._mags`.
+        """
```
fm3, line 107: can the “pass” be omitted if there is a docstring?

webknossos/webknossos/dataset/layer/remote_layer.py
```diff
@@ -62,11 +80,52 @@ def as_segmentation_layer(self) -> "RemoteSegmentationLayer":
...
+                dataset's access mode.
+
+        Returns:
+            RemoteMagView: View of data at the specified magnification
 
-    def get_finest_mag(self) -> MagView["RemoteLayer"]:
-        return super().get_finest_mag()
+        Raises:
+            IndexError: If specified magnification does not exist
+            ValueError: If the requested access mode is not available for this dataset
```
fm3:
```suggestion
            ValueError: If the requested access mode is not available for this mag
```
(it may be available for other mags of this dataset and this still happens, right?)

webknossos/webknossos/dataset/layer/remote_layer.py
```diff
@@ -62,11 +80,52 @@ def as_segmentation_layer(self) -> "RemoteSegmentationLayer":
...
+
+        Examples:
+            ```
+            mag1 = layer.get_mag(1, access_mode=RemoteAccessMode.DIRECT_PATH)
+            mag2 = layer.get_mag(2, access_mode=RemoteAccessMode.PROXY_PATH)
```
fm3:
```suggestion
            mag1 = layer.get_mag(1)
            mag2 = layer.get_mag(2, access_mode=RemoteAccessMode.DIRECT_PATH)
            mag4 = layer.get_mag(4, access_mode=RemoteAccessMode.PROXY_PATH)
```
Maybe the first example should not pass an access_mode. That will probably still be what most users should do.

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1651,7 +1651,19 @@ def add_layer_as_ref(
         )
 
         new_layer = self.add_layer_like(foreign_layer, new_layer_name)
-        for mag_view in foreign_layer.mags.values():
+        mag_views = list(foreign_layer.mags.values())
+        if len(mag_views) > 0:
+            # A remote layer's data_format describes its underlying storage, but
+            # referencing its mags references the paths of the layer's access mode,
+            # which may serve a different format (zarr streaming always serves Zarr).
```
fm3:
```suggestion
```
I think the value error msg explains this better than the comment.

webknossos/webknossos/dataset/remote_access_mode.py
```diff
@@ -0,0 +1,9 @@
+from enum import Enum
+
+
+class RemoteAccessMode(Enum):
+    """Determines how data of a remote dataset is accessed. Note that DIRECT_PATH can only be used if the client has access to the underlying storage."""
```
fm3:
```suggestion
    """Determines how data of a remote dataset is accessed.
    
`ZARR_STREAMING`: The WEBKNOSSOS datastore re-serves the data as Zarr. Works everywhere and is the default. It is also the only mode that can read annotations. 
`PROXY_PATH`: The WEBKNOSSOS datastore proxies the bytes of the underlying storage, preserving its data format. 
`DIRECT_PATH`: The underlying storage is read directly. This is the fastest option, but it requires that your machine has access to that storage (and credentials for it). 
```

For once I suggest a *longer* comment 😮

webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -264,88 +249,231 @@ def open(
...
         with self._context:
             return RemoteDataset.open(
                 dataset_id=self.dataset_id,
                 annotation_id_or_url=self.annotation_id,
                 access_mode=access_mode,
+                read_only=self.read_only,
             )
 
+    def with_access_mode(self, access_mode: RemoteAccessMode) -> "RemoteDataset":
+        """Returns this dataset with a different default access mode, without a new request.
```
fm3:
```suggestion
        """Returns this dataset with a different default access mode.
```

webknossos/Changelog.md
```diff
@@ -13,6 +13,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 [Commits](https://github.com/scalableminds/webknossos-libs/compare/v3.7.0...HEAD)
 
 ### Breaking Changes
+- The arguments of `RemoteDataset.__init__` changed. Use `RemoteDataset.open` instead, as documented.
```
fm3: please add pr links for every new changelog line

webknossos/Changelog.md
```diff
@@ -21,6 +22,8 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Removed the `timepoint` argument of `Dataset.add_layer_from_images`. Timeseries are now converted into a single layer with a `t` axis covering all timepoints, so selecting one only discarded data. Remove the argument from calls, and slice the `t` axis afterwards if you need a single timepoint. [#1477](https://github.com/scalableminds/webknossos-libs/pull/1477)
 
 ### Added
+- The access mode can now be set on the level of mags and attachments, e.g. `RemoteLayer.get_mag(mag, access_mode=RemoteAccessMode.PROXY_PATH)`, `RemoteMagView.with_access_mode()`, `RemoteAttachments.with_access_mode()`. This allows remote datasets to be edited (e.g. adding/removing layers/mags) no matter the access mode.
```
fm3: note that default is inherited from dataset

fm3 (approved):
Very nice 😎 Please have a look at my comments, all pretty small stuff

webknossos/webknossos/dataset/layer/view/remote_mag_view.py
normanrz: only 3 RemoteAccessModes

webknossos/webknossos/dataset/layer/abstract_layer.py
normanrz: seems so, but explicit pass probably doesn't hurt

```suggestion
        """
        pass
```

PR #1495, opened 2026-08-10
Add ChunkedImages abstraction for .ims and MRC, remove Bio-Formats

normanrz:
- Adds support for converting Imaris .ims files
- Adds a `ChunkedImages` abstraction for volumetric input formats, uses it for Imaris `.ims` and MRC
- Removes Bio-Formats from the conversion, which was effectively unused

---

## `ChunkedImages`

`PimsImages` reads slice-by-slice and can't know its true x/y extent up front, so conversion inflates the layer bbox to a 10-billion-wide placeholder and corrects it afterwards from observed slice shapes. Formats that are natively volumetric know their exact extent from metadata, so `ChunkedImages` reads whole shard-sized blocks and writes them directly — no placeholder, no post-hoc correction, no slice-based writer.

Adding a format is now a subclass plus `@register_chunked_images`; the subclass implements only `_read_source_box(timepoint, z, y, x) -> (c, z, y, x)` and the base owns all axis bookkeeping (mag, `swap_xy`, flips, channel and timepoint placement). zarr/n5/neuroglancer-precomputed should slot in the same way.

`.ims` gains support (new), and MRC moves over from `PimsMrcReader`.

## Bio-Formats removal

The Bio-Formats feature has probably not been used at all: It is not in the worker and not in the `webknossos convert` CLI.

Removes the `use_bioformats` argument, the reader, the `webknossos[bioformats]` extra and its JPype1/JVM dependency, and the DICOM example.

**Formats that were only reachable through Bio-Formats can no longer be converted** — among them `.dcm`/`.dicom`, `.nd2`, `.lif`, `.lsm`, `.zvi`, `.nii`, `.nrrd`, `.stk`. Formats with a dedicated reader (TIFF, CZI, DM3/DM4, `.ims`, MRC, common 2D images) are unaffected.

## Other API changes

- **`timepoint` removed** from `add_layer_from_images`. All timepoints now land in one layer on a `t` axis, so pinning one only discarded data. Timepoints are no longer split into separate layers either; `allow_multiple_layers` still splits channels.
- **Bounding boxes are honest**: `expected_bbox` reports every axis the source has, including `c` when more than one channel is written — the invariant `NormalizedBoundingBox` already assumed. A multi-channel, multi-timepoint file converts to a single `(t, c, x, y, z)` layer instead of raising.

## Fixes

Several are regressions introduced earlier in this branch, caught in review:

- `flip_x`/`flip_y` mirrored each shard individually rather than the whole image, scrambling multi-shard output into mirrored tiles. `flip_z` already avoided this; all three axes now share one implementation.
- `mag > 1` crashed with *"zero-size array to reduction operation maximum"* once the image spanned several shards, and **silently corrupted data** when converting compressed layers in parallel (chunks came out smaller than a shard, so workers overlapped within one shard).
- A `pathlib.Path` fell through to `PimsImages`, which has no reader for these formats, so it failed outright. Paths are now normalized to `UPath` at the API boundary.
- `PimsImages.expected_bbox` reported the source's raw channel count for n-dimensional multi-channel images, contradicting the layer's `num_channels` — a spurious *"images are larger than expected"* warning and a bbox disagreeing with the data. Pre-existing on master.
- `from_images` now names the missing extra (e.g. `.ims` without `webknossos[ims]`) instead of just omitting the format from the supported list.

## Testing

Conversion suite: 82 passed, 9 skipped. Beyond the crash/data assertions, the regression tests target what the old ones structurally couldn't see:

- flip and mag tests are parametrized over `shard_shape`, since single-shard input hides both bug classes;
- the parallel-corruption test uses a real multiprocessing executor, since `SequentialExecutor` cannot expose it;
- a test asserts the reader suffix table stays in sync with each reader's `class_exts()`.

Each was checked to fail with its fix reverted.

Verified by brute force where the space is small enough: all 16 flip×swap combinations multi-shard for both readers, and 24 mag×shard×flip combinations for MRC.

## Test data

Moved the large local fixtures out of `webknossos/testdata/` onto the `wklibs-samples` bucket on `static.webknossos.org` (same pattern already used for `brain_crop3.ims`), following up on review discussion:

- `WT1_wkw.tar.gz`, `4D/`, `various_tiff_formats/`, `tiff_with_different_shapes/`, `temca2/` — ~104MB removed from the repo.
- All remote/synthetic fixture helpers consolidated into a new `tests/data_fixtures.py`.
- Downloads are cached in a repo-local, gitignored `.cache/` dir, so repeated local test runs reuse what's already there instead of re-fetching every time.


webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -48,7 +52,12 @@ def test_compare_tifffile(tmp_upath: UPath) -> None:
         np.testing.assert_array_equal(data[:, :, z_index], comparison_slice)
 
 
-def test_mrc_from_images(tmp_upath: UPath) -> None:
+# Chunk-based formats have no pims reader to fall back on, so a path type that
+# try_open_chunked_images fails to recognize doesn't degrade — it fails outright
+# with "could not autodetect how to load a file of type mrc". pathlib.Path is
+# the easy one to miss, since it is not a UPath subclass.
+@pytest.mark.parametrize("path_type", [str, Path, UPath])
```
normanrz:
```suggestion
# Testing different path types in addition to the mrc conversion
@pytest.mark.parametrize("path_type", [str, Path, UPath])
```

webknossos/webknossos/dataset/_utils/chunked_images.py
```diff
@@ -0,0 +1,392 @@
...
+from ...geometry.bounding_box import BoundingBox
+from ...geometry.nd_bounding_box import NDBoundingBox
+from ...geometry.vec_int import VecInt
+from ..layer.view import MagView
+
+
+class ChunkedImages(ABC):
+    """
+    Base class for volumetric, chunk-based input formats (Imaris .ims and
+    MRC now; zarr, n5, neuroglancer_precomputed in the future).
```
normanrz:
```suggestion
    Base class for volumetric, chunk-based input formats (e.g. ims, mrc, zarr, n5).
```

webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -70,6 +76,377 @@ def test_mrc_from_images(tmp_upath: UPath) -> None:
...
+
+    read_data = layer.get_finest_mag().read()[0]
+    reference = _read_ims_reference(ims_path, 0)
+    expected = reference.transpose(2, 1, 0)
+    assert layer.bounding_box.size.to_tuple() == expected.shape
+    np.testing.assert_array_equal(read_data, expected)
+
+
+# The fixture is 673x635x51, so (256, 256, 32) spans several shards in every
+# axis while (1024, 1024, 64) fits in one. Each flip mirrors the whole source
```
fm3:
```suggestion
# The fixture is 673x635x51, so with shard shape (256, 256, 32) it spans several shards in every
# axis while with (1024, 1024, 64) it fits in one. Each flip mirrors the whole source
```

webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -70,6 +76,377 @@ def test_mrc_from_images(tmp_upath: UPath) -> None:
...
+    # channels into separate layers is what allow_multiple_layers=True does
+    # instead — see test_ims_from_images_multi_timepoint_multi_channel_creates_multiple_layers.
+    ims_path = tmp_upath / "synthetic_multi_t_multi_c.ims"
+    create_synthetic_multi_timepoint_ims(
+        ims_path, num_timepoints=2, num_channels=3, z=4, y=8, x=10
+    )
+    ims_chunked_images = importlib.import_module(
+        "webknossos.dataset._utils.ims_chunked_images"
+    )
+    monkeypatch.setattr(
```
fm3: why is this needed? to avoid verbose logging? should it also happen in the production code? or only in the test?

webknossos/tests/utils.py
```diff
@@ -14,3 +17,43 @@ def TestTemporaryDirectoryNonLocal() -> Generator[UPath, None, None]:
     temp_dir = UPath(f"memory:///{random_prefix}")
     temp_dir.mkdir(parents=True, exist_ok=True)
     yield temp_dir
+
+
+IMS_FIXTURE_URL = "https://static.webknossos.org/data/wklibs-samples/brain_crop3.ims"
```
fm3: should we move all test data out of the repository to this bucket? Or is it just for large files?

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
```diff
@@ -0,0 +1,391 @@
...
+    # read range below from these, so they are always pre-swap_xy: _x is the
+    # source's x, even when swap_xy makes it the output's y.
+    _x: int
+    _y: int
+    _z: int
+    # Timepoint handling. Formats without a time dimension leave these alone
+    # and are read at timepoint 0; a subclass with several timepoints and none
+    # pinned sets _include_t_axis, which puts a "t" axis on expected_bbox and
+    # makes each chunk carry its own timepoint.
+    _t: int = 1
```
fm3, line 46: do we plan on supporting other axes than t in the future?

webknossos/webknossos/dataset/_utils/chunked_images.py
```diff
@@ -0,0 +1,391 @@
...
+        self._channel = channel
+        self._swap_xy = swap_xy
+        self._flip_x = flip_x
+        self._flip_y = flip_y
+        self._flip_z = flip_z
+        self._is_segmentation = is_segmentation
+
+    @classmethod
+    @abstractmethod
+    def class_exts(cls) -> set[str]:
```
fm3: Maybe name this `supported_file_extensions`? With `class_exts` I have to read the docstring

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
```diff
@@ -0,0 +1,391 @@
...
+
+from ...geometry.bounding_box import BoundingBox
+from ...geometry.nd_bounding_box import NDBoundingBox
+from ...geometry.vec_int import VecInt
+from ..layer.view import MagView
+
+
+class ChunkedImages(ABC):
+    """
+    Base class for volumetric, chunk-based input formats (e.g. ims, mrc, zarr, n5).
```
fm3: Is this only used for reading input data when converting to our own format? Or is it also used when handling our own wk datasets? Since zarr is in the list here, I would make this even more explicit to avoid future confusion

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
```diff
@@ -0,0 +1,391 @@
...
+        The exact bounding box of the data to convert, in the source's
+        native Mag(1) space. Unlike PimsImages.expected_bbox, this never
+        needs placeholder inflation, since chunk-based formats know their
+        true extents from metadata alone.
+
+        Reports every axis the source actually has, including "c" when more
+        than one channel is written (which is where NormalizedBoundingBox
+        reads num_channels from) and "t" for unpinned multi-timepoint data.
+        Always well-defined — it never has to reject an axis combination.
+        """
```
fm3, line 115: I feel like these docstrings are a little too verbose. I fear that on future changes the likelihood that they become stale is large because the person doing changes might not re-read them completely. This stuff is already explained in several spots.

webknossos/webknossos/dataset/_utils/chunked_images.py
```diff
@@ -0,0 +1,391 @@
...
+        valid_suffixes.update(cls.class_exts())
+    return valid_suffixes
+
+
+def try_open_chunked_images(
+    # Deliberately excludes pims.FramesSequence, which callers may also hold:
+    # pims ships no stubs, so that type is Any and a union containing it makes
+    # the annotation accept everything — including the un-normalized str/Path
+    # this signature exists to reject. Callers passing a FramesSequence-typed
+    # value still type-check for exactly that reason.
```
fm3: I don’t really understand this comment 😕

webknossos/webknossos/dataset/_utils/image_conversion.py
```diff
@@ -0,0 +1,863 @@
+"""Implementation of Dataset.from_images() / Dataset.add_layer_from_images(),
+factored out of dataset.py since this functionality (mapping input image
+files to layers, then converting them via either the ChunkedImages or
+PimsImages reader path) is large and self-contained.
```
fm3: This describes the diff rather than the new code. I’d remove this comment

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -0,0 +1,863 @@
...
+        if suffix in unavailable:
+            found[suffix] = unavailable[suffix]
+    if not found:
+        return None
+
+    extras = sorted({extra for extra in found.values()})
+    return (
+        f". Found {', '.join('.' + s for s in sorted(found))} files, which need an "
+        + "optional dependency that is not installed — install it with "
+        + f"`pip install {' '.join(f'webknossos[{extra}]' for extra in extras)}`"
```
fm3, line 201: Is recommending pip install in this context still best practice in times of uv?

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -0,0 +1,863 @@
...
+
+        with warnings.catch_warnings():
+            # We need to catch and ignore a warning here about comparing persisted properties.
+            # The problem is that there is an `axisOrder` property that goes in the datasource-properties.json
+            # file, but is only stored in a mag object. However, at first we don't have any mags
+            # so there is no place to store them. When we add the first mag and update the properties,
+            # there is a check that reads the properties first and compares them to the current state.
+            # This check fails because the axisOrder property hasn't been stored in the properties file.
+            # It is safe to ignore this warning because it is only an intermediate problem in this function.
+            # At the end of this function, the properties are complete and consistent.
```
fm3, line 774: will the check be done again at the end?

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -0,0 +1,863 @@
...
+                    )
+                    dataset._save_dataset_properties()
+
+        except Exception:
+            # The used heuristic was not able to guess the layer category, the previous value is kept
+            pass
+        if first_layer is None:
+            first_layer = layer
+    assert first_layer is not None
+    return first_layer
```
fm3, line 863: Is it just me or is this function way too long? Probably not introduced in this PR, but I would really like this to be split up at some point

webknossos/webknossos/dataset/_utils/ims_chunked_images.py
```diff
@@ -0,0 +1,126 @@
...
+    return shape, dtype  # type: ignore[return-value]
+
+
+@register_chunked_images
+class ImsChunkedImages(ChunkedImages):
+    """
+    ChunkedImages implementation for Imaris .ims files. Reads shard-sized 3D
+    blocks directly from the underlying HDF5 file via h5py and writes them
+    to mag_view directly — no slice-by-slice pims reading, no BufferedSliceWriter.
+    This is the only supported way .ims files are read for conversion.
```
fm3:
```suggestion
    ChunkedImages implementation for Imaris .ims files. Reads shard-sized 3D
    blocks directly from the underlying HDF5 file via h5py and writes them
    to mag_view directly.
```

webknossos/webknossos/dataset/_utils/mrc_chunked_images.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+@register_chunked_images
+class MrcChunkedImages(ChunkedImages):
+    """
+    ChunkedImages implementation for MRC files. MRC data is stored as a
+    single contiguous array (no internal chunking, unlike HDF5-based
+    formats), so shard-sized blocks are read directly via mrcfile's
+    memory-mapped array and written to mag_view directly — no
+    slice-by-slice pims reading, no BufferedSliceWriter.
```
fm3:
```suggestion
    ChunkedImages implementation for MRC files. MRC data is stored as a
    single contiguous array, so shard-sized blocks are read directly via mrcfile's
    memory-mapped array and written to mag_view directly.
```

webknossos/Changelog.md
```diff
@@ -13,12 +13,22 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 [Commits](https://github.com/scalableminds/webknossos-libs/compare/v3.7.0...HEAD)
 
 ### Breaking Changes
+- Removed Bio-Formats support, including the `use_bioformats` argument of `Dataset.from_images`/`add_layer_from_images`/`RemoteDataset.from_images`, the `webknossos[bioformats]` extra (and its `JPype1`/JVM dependency), and the DICOM upload example. Formats that were only readable through Bio-Formats — among them `.dcm`/`.dicom`, `.nd2`, `.lif`, `.lsm`, `.zvi`, `.nii`, `.nrrd` and `.stk` — can no longer be converted. Formats with a dedicated reader (TIFF, CZI, DM3/DM4, `.ims`, MRC, and the common 2D image formats) are unaffected. [#1477](https://github.com/scalableminds/webknossos-libs/pull/1477)
```
fm3: removing the example isn’t exactly a breaking change, I’d skip that here

fm3 (changes requested):
Pretty cool stuff, I think this is moving in a good direction :-) Happy that you reduced the size of dataset.py as well.

There are still a lot of very long functions and there is a lot of complexity of course. I hope we can simplify further.

I did see nrrd, dicom and lif occasionally in failed conversion jobs. I think it might be worth creating converters for those again.

Just for my understanding: You mentioned in the PR description that channels now stay “honest” and include everything from the input. But wk does not actually support multiple channels (except for the 3×uint8 special case) – is that where you should use allow_multiple_layers? Should it be named split_channels_to_layers or something? –Ok I see now that from_images always sets that flag to true. Is it still possible to create the uint24 layers with it for rgb rendering?

Other than that, please have a look at my questions/suggestions!

webknossos/tests/utils.py
normanrz: Yeah, I think we should move them all.

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
normanrz: Probably, but ims and mrc don't have it.

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
normanrz: This is only for conversion into Datasets. I think I'll move these modules to `webknossos/dataset/_image_conversion`.

webknossos/Changelog.md
normanrz:
```suggestion
- Removed Bio-Formats support, including the `use_bioformats` argument of `Dataset.from_images`/`add_layer_from_images`/`RemoteDataset.from_images`, and the `webknossos[bioformats]` extra (and its `JPype1`/JVM dependency). Formats that were only readable through Bio-Formats — among them `.dcm`/`.dicom`, `.nd2`, `.lif`, `.lsm`, `.zvi`, `.nii`, `.nrrd` and `.stk` — can no longer be converted. Formats with a dedicated reader (TIFF, CZI, DM3/DM4, `.ims`, MRC, and the common 2D image formats) are unaffected. [#1477](https://github.com/scalableminds/webknossos-libs/pull/1477)
```

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: I think pip is still is the default

webknossos/tests/utils.py
normanrz: Done in https://github.com/scalableminds/webknossos-libs/pull/1495/commits/80a09b476082e1b3ce66f67f04a1f64bf3675564

normanrz:
> I did see nrrd, dicom and lif occasionally in failed conversion jobs. I think it might be worth creating converters for those again.

See #1507 

> Just for my understanding: You mentioned in the PR description that channels now stay “honest” and include everything from the input. But wk does not actually support multiple channels (except for the 3×uint8 special case) – is that where you should use allow_multiple_layers? Should it be named split_channels_to_layers or something? –Ok I see now that from_images always sets that flag to true. Is it still possible to create the uint24 layers with it for rgb rendering?

Yes, `allow_multiple_layers` should be used. I think this gets cleaned up a bit in #1496. Let's continue the discussion there.

webknossos/webknossos/dataset/_utils/chunked_images.py
normanrz: I flatly removed this. The structure of this changes in #1498 anyways.

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
normanrz: Fair point. Some of this changes in #1498. I'll go over the docstrings there one more time.

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: Huh. Not even 500 lines :D

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: Well, this part is verbatim copied from the original code. It is not actually checked in the end. I'll take a look if that would be feasible.

webknossos/webknossos/dataset/_utils/chunked_images.py
normanrz: Fixed in https://github.com/scalableminds/webknossos-libs/pull/1495/changes/8a348135db420de95ad8b4d7e379126767b10b13

webknossos/examples/convert_4d_tiff.py
normanrz, line 1: Should we drop this? The downloading code seems too weird to be put in an example.

webknossos/tests/data_fixtures.py
```diff
@@ -0,0 +1,96 @@
...
+            zip_file.extract(info, dest_dir)
+
+
+def download_wklibs_sample_archive(name: str) -> UPath:
+    """Downloads `{name}.zip` from the wklibs-samples bucket and extracts it
+    into CACHE_DIR, once ever (subsequent calls, including from later test
+    runs, reuse the cached extraction). Each archive's contents live at a
+    top-level file or folder named after the archive (e.g. `4D.zip` contains
+    `4D/`, `brain_crop3.ims.zip` contains `brain_crop3.ims`), which is what
+    this function returns the path to."""
```
normanrz:
```suggestion
    runs, reuse the cached extraction)."""
```

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
normanrz: Moved in https://github.com/scalableminds/webknossos-libs/pull/1495/changes/c29296ac69f2d15aed288b8f435b02993b85042d

webknossos/webknossos/dataset/_image_conversion/chunked_images.py
```diff
@@ -0,0 +1,387 @@
...
+
+from ...geometry.bounding_box import BoundingBox
+from ...geometry.nd_bounding_box import NDBoundingBox
+from ...geometry.vec_int import VecInt
+from ..layer.view import MagView
+
+
+class ChunkedImages(ABC):
+    """
+    Base class for volumetric, chunk-based input formats (e.g. ims, mrc, zarr, n5).
```
normanrz:
```suggestion
    Base class for reading volumetric, chunk-based input formats for conversion.
```

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: Added a proper check in https://github.com/scalableminds/webknossos-libs/pull/1495/changes/607b1d7a7a7af3b95b5270fe87cf033e62c90d06

webknossos/examples/convert_4d_tiff.py
fm3: Maybe extract the downloading to a utility function so that it can still run in our CI? Then the example code can focus on what to do with the downloaded input.
I can also live with dropping this one, if importing the util violates conventions of the examples

fm3 (approved):

PR #1496, opened 2026-08-10
Report image conversion input errors, and split non-RGB channels into layers

normanrz:
We want to catch errors from the image conversion process in the worker in order to surface useful error messages to the user. These typed error classes are the prerequisite for that.

---
Claude's PR description:

Stacked on #1495 — please review/merge that one first.

Two related things: conversion failures caused by the *input* become catchable exceptions, and multi-channel images are split into layers unless their channels are actually colour.

## Catchable errors

When conversion failed because of its input, it failed in whatever way the reader happened to fail, so a caller could not tell "this file is damaged" from "this format is not supported" from a genuine bug — the only handle was the message text. Removing Bio-Formats in #1495 makes several of these paths common, and a downstream codebase wants to turn each into a user-friendly message.

```
ValueError
└── ImageConversionError          # catch this for "the input is the problem"
    ├── UnsupportedImageFormatError
    ├── CorruptImageError
    └── UnsupportedImageDataError
```

`ImageConversionError` subclasses `ValueError` — what these call sites raised before — so `except ValueError` keeps working. All carry the offending `path`.

| exception | means | what the user should do |
| --- | --- | --- |
| `UnsupportedImageFormatError` | no reader handles this format | convert to a supported format — or, if `missing_extras` is non-empty, install that extra (`.ims` without `webknossos[ims]`); also carries `suffix` and `supported_suffixes` |
| `CorruptImageError` | supported format, unreadable file | re-upload — it is damaged or incomplete |
| `UnsupportedImageDataError` | read fine, cannot be stored as requested | convert with different arguments |

They live in a new, dependency-free `webknossos/dataset/errors.py`, so catching them costs no reader imports. `WkImportError` is unchanged: it is an import-time signal, swallowed by the reader registries (it *is* an `ImportError`), which is why an uninstalled format just disappears from the supported list — `missing_extras` is how that resurfaces at conversion time.

### Judgement calls

**Corrupt vs. missing.** `pims.open` flattens each handler's exception into a *message*, so a missing file, a permission error and a damaged file all arrive as the same `UnknownFormatError`. Rather than guess from exception types, the file is probed directly: a path that does not exist, or cannot be read at all, keeps its own error instead of being called corrupt.

**Unsupported format vs. corrupt.** Same root cause — `UnknownFormatError` is raised both when no handler claims the suffix and when every handler that claimed it errored out. So the format check goes by **suffix**, and only after every open strategy has failed, which also keeps files that pims opens without a recognized suffix (via `ImageSequence`/skimage) working.

## Channels

Several channels shared a layer whenever the dtype allowed it, which got both common cases wrong:

| input | before | now |
| --- | --- | --- |
| RGB `.png` via `from_images` | 3 grayscale layers | 1 RGB layer |
| RGBA `.png` via `from_images` | 4 grayscale layers | 1 RGB layer (alpha dropped) |
| 3×uint16 `.ims` via `add_layer_from_images` | 1 layer WEBKNOSSOS cannot display | `UnsupportedImageDataError` |
| 3×uint8 `.ims`/TIFF via `from_images` | 1 RGB layer | 3 layers |

Whether channels belong together is a property of the **format**, not the dtype. The everyday image formats (`png`, `jpg`, `bmp`, `gif`, `webp`, …) store colour, so three uint8 channels are RGB and belong in one layer. Every other source stores one acquisition per channel — `.ims`, MRC, TIFF, CZI, DM3/DM4 alike — and those become one layer each, however many there are.

- `Dataset.from_images` splits them; it converts whatever it is given.
- `Dataset.add_layer_from_images` raises `UnsupportedImageDataError` naming both ways out (`allow_multiple_layers=True`, or `channel=<index>`) rather than silently adding N layers when it was asked for one.

The decision uses the **source's** channel count rather than the reader's `num_channels`, which is pinned or truncated by then — a two-channel source reported one channel and discarded the second without a word.

## Also fixed

- `Dataset.from_images` on a single unsupported file raised `UnboundLocalError: cannot access local variable 'input_files'` (pre-existing on master).
- A corrupt MRC converted into an *empty layer*: `mrcfile` parses an unreadable header permissively and reports an empty extent, which is now detected up front.

## Testing

`tests/dataset/test_from_images.py` + `tests/dataset/test_add_layer_from_images.py`: 101 passed, 9 skipped.

New tests cover each exception and each channel rule: the single unsupported file (`UnboundLocalError` before the fix); corrupt `.tif`/`.mrc`/`.ims`; that a *missing* file is not called corrupt; a float image refused for a segmentation layer; RGB and RGBA png staying one layer through both entry points; `.ims` and 16-bit TIFF channels needing one layer each (parametrized over 2/3/4 channels and uint8/uint16, so the format rather than the dtype is what is pinned); `channel=` as the other way out; the missing-extra case (reader unregistered via monkeypatch, since the test env installs every extra); and the `ValueError` back-compat contract.


webknossos/Changelog.md
```diff
@@ -18,12 +18,20 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
...
 
 ### Changed
 - `.ims` and MRC file conversion now reads shard-sized blocks directly instead of slice-by-slice, improving conversion performance for large files. These formats are always read through a dedicated reader. [#1477](https://github.com/scalableminds/webknossos-libs/pull/1477)
+- Several channels are now only written into a single layer when they are actually colour: three uint8 channels of an image format that stores colour (`.png`, `.jpg`, `.bmp`, …), which WEBKNOSSOS displays as RGB. The channels of every other source — `.ims`, MRC, TIFF, CZI, DM3/DM4 — are separate acquisitions and become one layer each, whatever their number and dtype. `Dataset.from_images` splits them; `Dataset.add_layer_from_images` raises `UnsupportedImageDataError` unless `allow_multiple_layers=True` (write one layer per channel) or `channel=<index>` (convert one channel) is given, since it adds a single layer by contract. Previously the dtype alone decided, so three 16-bit channels were written into one layer that WEBKNOSSOS cannot display, while an RGB `.png` was split into three grayscale layers by `from_images`. [#1496](https://github.com/scalableminds/webknossos-libs/pull/1496)
```
fm3: Let’s avoid the colour spelling. I’d always just use “RGB” or “RGB color” here in place of “colour”, since we have “color” already as a fixed keyword (as opposed to segmentation). This goes for changelog, code comments and PR title.

webknossos/webknossos/dataset/errors.py
```diff
@@ -0,0 +1,100 @@
+"""Exceptions raised by the dataset API.
```
fm3:
```suggestion
"""Exceptions raised by the dataset conversion.
```

fm3: (at least for now I guess)

webknossos/webknossos/dataset/errors.py
```diff
@@ -0,0 +1,100 @@
...
+from upath import UPath
+
+
+class ImageConversionError(ValueError):
+    """Base class for the ways image conversion can fail because of its input.
+
+    Raised by `Dataset.from_images` and `Dataset.add_layer_from_images`.
+
+    Subclasses `ValueError`, which the call sites raised before these exceptions
+    existed, so `except ValueError` keeps working.
```
fm3:
```suggestion
    Raised by `Dataset.from_images` and `Dataset.add_layer_from_images`.
```
(describes the diff more than the new code)

webknossos/webknossos/dataset/errors.py
```diff
@@ -0,0 +1,100 @@
...
+class UnsupportedImageFormatError(ImageConversionError):
+    """Raised when image data cannot be converted because no reader handles its format.
+
+    Raised either because the input contains no file with a supported suffix,
+    or because every reader failed to recognize the file.
+
+    Attributes:
+        path: The offending input path, if there is a single one. `None` when
+            the images were passed as a list or as a `pims.FramesSequence`.
+        suffix: The offending file's suffix, lowercase and without the leading
```
fm3: in the upstream PR you used "ext" or "extension" for this, now it’s "suffix". Is the difference the dot? Or should it be unified?

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1161,7 +1170,27 @@ def add_layer_from_images(
         * `allow_multiple_layers`: set to `True` if channels may result in multiple layers being added (only the first is returned)
         * `max_layers`: only applies if `allow_multiple_layers=True`, limits the number of layers added via different channels
         * `truncate_rgba_to_rgb`: only applies if `allow_multiple_layers=True`, set to `False` to write four channels into layers instead of an RGB channel
+
+        Several channels are only written into a single layer when they are
+        actually colour: three uint8 channels of an image format that stores
```
fm3: again, avoid “colour”. Also, I think this comment can be shortened

webknossos/webknossos/dataset/_image_conversion/ims_image_source.py
```diff
@@ -75,7 +76,17 @@ def __init__(
...
+            file_shape, self.dtype = _read_ims_metadata_quietly(str(path))
+        except Exception as e:
+            # The reader speaks HDF5, so a damaged or truncated file surfaces
+            # as an OSError/KeyError from h5py deep inside it, which says
+            # nothing useful to whoever uploaded the file.
+            raise CorruptImageError(
+                f"Cannot open IMS file {path}. "
+                + "The file is likely corrupted or not a valid IMS file.",
+                path=path,
+            ) from e
```
fm3, line 71: should this also check for a shape that has zero in an axis, like the mrc codepath?

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -201,15 +207,68 @@ def _find_unavailable_input_formats(input_upath: UPath) -> str | None:
...
+    Whether the source's channels are colour channels of a single layer, rather
+    than separate channels that each deserve their own layer.
+
+    This needs the *source's* channel count, not the reader's `num_channels`:
+    the readers already pin or truncate channels (a two-channel source reports
+    one channel, pinned to the first), which is exactly the silent discarding
+    this decision exists to avoid.
+    """
+    if num_source_channels == 1:
+        return True
```
fm3, line 287: If there is just one source channel, it cannot be rgb, right? So either the function is named badly or this return is wrong

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -377,10 +436,18 @@ def from_images(
     if z_slices_sort_key is None:
         z_slices_sort_key = natsort_keygen()
 
-    if input_upath.is_file():
+    # Kept around for the error below, since input_upath moves to the parent
+    # directory for a single file that can be converted.
```
fm3: I don’t understand this comment, maybe it can be rephrased

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -603,17 +684,34 @@ def _reopen_image_source(
...
+            # being offered up as one layer each. The readers take care of the
+            # RGBA -> RGB truncation.
+            assert possible_layers is not None
+            del possible_layers["channel"]
+        elif not allow_multiple_layers:
+            raise UnsupportedImageDataError(
+                f"Cannot convert the {len(source_channels)} channels of these images "
+                + "into a single layer: WEBKNOSSOS only displays several channels "
+                + "within one layer as RGB, which means three uint8 channels of an "
+                + f"image format that stores colour ({_describe_rgb_formats()}). "
```
fm3: I think this should also be rephrased, reads pretty complicated

webknossos/webknossos/dataset/_image_conversion/pims_images.py
```diff
@@ -333,6 +339,110 @@ def _normalize_original_images(self) -> str | list[str]:
         else:
             return str(original_images)
 
+    def _error_path(self) -> UPath | None:
+        """The single input path to blame when opening fails, or None when
```
fm3:
```suggestion
        """The single input image path to blame when opening fails, or None when
```

webknossos/webknossos/dataset/_image_conversion/pims_images.py
```diff
@@ -333,6 +339,110 @@ def _normalize_original_images(self) -> str | list[str]:
...
+        *is* supported, which must keep surfacing as its original error.
+
+        The decision is made on the suffix, not on the exception type:
+        pims.open raises UnknownFormatError both when no handler claims the
+        suffix and when every handler that claimed it errored out, so a
+        corrupt TIFF is indistinguishable from an unreadable format by type.
+
+        Classifying afterwards, rather than rejecting unknown suffixes up
+        front, keeps files that pims opens without a recognized suffix (via
+        pims.ImageSequence/skimage) working as before: this only runs once
```
fm3:
```suggestion
        pims.ImageSequence/skimage) working: this only runs once
```

webknossos/webknossos/dataset/_image_conversion/pims_images.py
```diff
@@ -333,6 +339,110 @@ def _normalize_original_images(self) -> str | list[str]:
...
+        every open strategy has already failed.
+        """
+        # Imported inside the function so that the two reader modules stay
+        # independent at import time; this is only needed on the error path.
```
fm3: here “error path” means error codepath. But just a few lines below we have self._error_path(). Maybe we can avoid some confusion for future readers by making both more explicit

webknossos/webknossos/dataset/_image_conversion/pims_images.py
```diff
@@ -333,6 +339,110 @@ def _normalize_original_images(self) -> str | list[str]:
...
+            supported_suffixes=tuple(sorted(supported_suffixes)),
+            missing_extras=tuple(sorted(set(missing.values()))),
+        )
+
+    def _classify_read_failure(self) -> CorruptImageError | None:
+        """
+        Decides whether a failure to open a file of a *supported* format means
+        the file itself is unreadable — damaged or incompletely uploaded, by
+        far the most common cause — as opposed to a problem this class should
+        not put words in the user's mouth about.
```
fm3: very poetic ;-)

webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -328,6 +330,104 @@ def test_timepoint_argument_is_removed(tmp_upath: UPath) -> None:
         )
 
 
+def test_add_layer_from_images_unsupported_format(tmp_upath: UPath) -> None:
+    # No reader handles .dcm since Bio-Formats was removed. Downstream code
```
fm3:
```suggestion
    # No reader handles .dcm. Downstream code
```

webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -328,6 +330,104 @@ def test_timepoint_argument_is_removed(tmp_upath: UPath) -> None:
...
+        # An HDF5 signature with nothing behind it: the .ims reader gets far
+        # enough to fail deep inside h5py.
+        ("broken.ims", b"\x89HDF\r\n\x1a\n" + b"\x00" * 100, True),
+    ],
+)
+def test_add_layer_from_images_corrupt_file(
+    tmp_upath: UPath, filename: str, contents: bytes, wraps_reader_error: bool
+) -> None:
+    # A damaged file of a *supported* format is the most common upload
+    # failure, and needs a different message than an unsupported format.
```
fm3:
```suggestion
    # A damaged file of a *supported* format needs a different message than an unsupported format.
```

webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -363,45 +463,144 @@ def test_ims_from_images_multi_timepoint(
...
+    read = layer.get_finest_mag().read()
+    for channel in range(3):
+        assert (read[channel] == channel + 1).all()
+
+
+def test_multi_channel_16bit_tiff_needs_one_layer_per_channel(
+    tmp_upath: UPath,
+) -> None:
+    # TIFF is not an image format that stores colour, so even three channels
+    # are separate acquisitions rather than RGB.
```
fm3: I’ve stored rgb images as tiff many times. This comment is wrong. At most we could say “In microscopy context, tiff is usually not used for rgb channels”.

However, since the test is about uint16, wk can only do three channels anyway.

fm3 (changes requested):
Sweet, I really like the exception concept :-)

On the channel thing, I think it’s very good to be explicit, but I’m not sure about some of the choices:
 - The PR description says that `3×uint16 .ims via add_layer_from_images` will now result in `UnsupportedImageDataError` – why can’t it create three layers? Or maybe it’s just because the allow_multiple_layers thing always needs to be explicit for this function?
 - When uploading an 3×uint8 tif file, I would expect RGB output. A code comment says that tif isn’t used for RGB but I think that’s not generally true. But you know our users better, maybe in our context here it’s at least true most of the time. Your call.

Also, I think we need to avoid the word “colour” all around here, to avoid confusion with the existing keyword “color” as layer category.

Please have a look at the individual comments :)

webknossos/Changelog.md
normanrz: Fixed — replaced "colour" with "RGB" everywhere it appeared: this changelog entry, the code comments/docstrings in `image_conversion.py`/`dataset.py`, and the test comments. Also updated the PR title.

webknossos/webknossos/dataset/errors.py
normanrz: Applied, thanks.

webknossos/webknossos/dataset/_image_conversion/ims_image_source.py
normanrz: Good catch — added the same zero-in-any-axis guard as `MrcChunkedImages`, raising `CorruptImageError`, plus a regression test (`test_ims_empty_shape_is_corrupt`).

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: Reworded to "input_upath is replaced with its parent directory below for a single convertible file, so keep the original around to report in the error."

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: Simplified: "Cannot write these {n} channels into a single layer: WEBKNOSSOS only combines several channels into one layer as RGB (three uint8 channels of {formats}). Set allow_multiple_layers=True to write one layer per channel, or channel=<index> to convert a single channel."

webknossos/webknossos/dataset/_image_conversion/pims_images.py
normanrz: Applied.

webknossos/webknossos/dataset/_image_conversion/pims_images.py
normanrz: Applied.

webknossos/webknossos/dataset/_image_conversion/pims_images.py
normanrz: Reworded to something more literal: "...as opposed to some other failure this class has no specific enough evidence to explain."

webknossos/tests/dataset/test_add_layer_from_images.py
normanrz: Applied.

webknossos/tests/dataset/test_add_layer_from_images.py
normanrz: Applied.

webknossos/webknossos/dataset/errors.py
normanrz: Removed that sentence, you're right it was describing the diff rather than the code.

webknossos/webknossos/dataset/errors.py
normanrz: ~~No real difference, just inconsistent naming on my part. Kept `suffix` (matches the rest of the module: `supported_suffixes`, `get_valid_chunked_image_suffixes`, etc.) and added a clarifying "the offending file's extension" to the docstring so both terms are tied together for a reader coming from the other PR.~~

Fixed in https://github.com/scalableminds/webknossos-libs/pull/1496/commits/d5345ac689f22692f604bee934698df58b7e5653

webknossos/webknossos/dataset/dataset.py
normanrz: Reworded away from "colour" and shortened it. Also moved it out from between the `truncate_rgba_to_rgb` and `executor` bullets, where it was awkwardly interrupting the list — it now sits after the full bullet list instead.

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: You're right — that branch was both wrong (a single channel is never RGB) and actually unreachable: the only call site only invokes this with 2+ source channels (`get_possible_layers()` never offers a "channel" split for just one channel). Removed it and documented the 2+ invariant in the docstring instead.

webknossos/webknossos/dataset/_image_conversion/pims_images.py
normanrz: Renamed the method to `_blame_path()` and reworded the neighboring comment to say "only needed once opening has failed" instead of "error path", so the two no longer collide.

webknossos/tests/dataset/test_add_layer_from_images.py
normanrz: You're right, fixed. The comment now explains that `.tif` just isn't in this codebase's RGB-suffix allowlist (rather than claiming TIFF can't store RGB, which it can), and notes that the test uses uint16 anyway, where RGB never applies regardless of format.

normanrz:
> * The PR description says that `3×uint16 .ims via add_layer_from_images` will now result in `UnsupportedImageDataError` – why can’t it create three layers? Or maybe it’s just because the allow_multiple_layers thing always needs to be explicit for this function?
 
Yeah, that would require `allow_multiple_layers=True`.

> * When uploading an 3×uint8 tif file, I would expect RGB output. A code comment says that tif isn’t used for RGB but I think that’s not generally true. But you know our users better, maybe in our context here it’s at least true most of the time. Your call.

I think most of our users would want their multi-channel tiffs to be in different layers so that they get individual histograms, can turn them on/off, and set the color. That holds true for tiffs that happen to have three channels.

The thread comments were replied to by Claude (I later added instructions for Claude to mark them as such).

fm3 (approved):
Thanks for implementing the feedback! I only added a small commit with a change to a comment.

> I think most of our users would want their multi-channel tiffs to be in different layers so that they get individual histograms, can turn them on/off, and set the color. That holds true for tiffs that happen to have three channels.

Fair enough. As discussed in person, it would be nice if there could still be an option the users could set to actually get their uint8 rgb tiffs into uint24 layers (without having to manually write magViews). Doesn’t have to block this PR.

Also, when writing multi-channel to multi-layer, would it be possible to already set colors for them in the view config (heuristically or determined from input metadata)? That way, users wouldn’t at first see everything in grayscale. Of course this can be a new issue as well

PR #1498, opened 2026-08-10
Internalize the pims dependency

normanrz:
This is a fairly big refactoring motivated by the lack of maintenance in pims. This PR internalizes all of pims and restructures the class hierarchies so that it works more elegantly with sliced (e.g. tiff stack) and chunked (e.g. ims, mrc) formats. All of this is under the hood and not really user-facing.

---

Claude's PR description:

Stacked on #1496 (which is stacked on #1495) — please review/merge those first. Replaces `scm-pims` with an in-library implementation of the parts we actually use, then cleans up what that exposed.

`scm-pims` is our own fork of an effectively unmaintained upstream, so every Python/numpy bump needs a fork release before this library can move. The surface we depend on is thin: the `FramesSequenceND` base class our readers subclass, `Slicerator`'s lazy slicing, `pims.open()`'s suffix dispatch, and the generic 2D-image readers. Everything else pims ships was already unused — Bio-Formats went away in #1495.

Sizes, so nobody has to guess: `_utils/` grows from 5290 to 5841 lines, which is the cost of owning the code instead of depending on a fork. The two largest modules shrink — `pims_images.py` (802) → `sliced_image_source.py` (537), and `image_conversion.py` 971 → 740 — and the last three commits take 330 lines back out of `_utils/` on their own.

## The shape of it

Two reading strategies, one interface:

| | knows its extents up front | reads |
| --- | --- | --- |
| `SlicedImageSource` | no — placeholder bbox, corrected after | one 2D slice at a time, via a `SliceReader` |
| `ChunkedImageSource` | yes, from metadata | whole shard-aligned blocks |

`ImageSource` is what `image_conversion.py` drives, and `open_image_source(images, options)` is the only way either is created — so nothing outside the registry asks which kind it is holding. Format readers register for their suffixes: `SingleImageSliceReader`, `MultiImageSliceReader`, `StackedFileSliceReader`, `TiffSliceReader`, `Dm3SliceReader`/`Dm4SliceReader` on the sliced side; `CziImageSource`, `ImsImageSource`, `MrcImageSource` on the chunked side.

image-conversion-classes

`ChunkedImageSource` is a base class — its subclasses supply only `_read_source_box()`, and all the mag, swap, flip, channel and timepoint bookkeeping lives in the parent. `SlicedImageSource` instead *composes* a `SliceReader`, re-opening one per chunk because sources are pickled across processes and must not hold file handles. That is why `TiffSliceReader` and `CziImageSource` are siblings in role but unrelated in the type graph: the suffix a reader claims decides which branch it lands on, and the registry is the only place that mapping exists.

## 1. The replacement

Three modules under `webknossos/dataset/_image_conversion/`:

- **`slice_reader.py`** — `SliceReader` (named axes, `bundle_axes`/`iter_axes`/`default_coords`, and the adapter that narrows a reader's `get_slice` to the axes asked for), plus `_SlicedView` replacing `slicerator.Slicerator`. `pims.Frame` is gone — slices are plain ndarrays, which also removes the `display.py`/jinja2/PIL-rich-repr chain.
- **`image_source_registry.py`** — `open_image_source()` and the suffix registry replacing `pims.open()`, with `@register_slice_reader` / `@register_chunked_image_source` decorators mirroring the existing chunked-source registry rather than pims's subclass walk.
- **`common_slice_readers.py`** — `SingleImageSliceReader` / `MultiImageSliceReader` / `StackedFileSliceReader`, backed by `imageio` (what pims used), so decoded arrays are unchanged.

`SlicedImageSource` additionally sheds the `ImageIOReader` monkeypatch and all `_init_c_axis` handling (~30 lines across three call sites), which existed solely to patch that reader.

## 2. Vocabulary

The reader layer had three conflicting senses of "sequence" (a sequence of slices, a reader that *is* one, a sequence of files, a sequence of readers), and used "frame" and "slice" for the same 2D plane. Now one word per concept — **slice** (a 2D plane), **chunk** (an output-aligned block), **source** (one set of input images being converted) — and "sequence" survives nowhere, including the base class name: a follow-up rename turned `SliceSequence` into `SliceReader`, matching `BufferedSliceWriter`/`BufferedSliceReader`, which were already using "slice".

| before | after |
| --- | --- |
| `FrameSequence` / `NDFrameSequence` | `SliceReader` (one class, see 6) |
| `PimsTiffReader` | `TiffSliceReader` |
| `PimsCziReader` | `CziImageSource` (see 4) |
| `PimsDm3Reader` / `PimsDm4Reader` | `Dm3SliceReader` / `Dm4SliceReader` |
| `pims.ImageReader` | `SingleImageSliceReader` |
| `pims.ImageSequence` | `MultiImageSliceReader` |
| `pims.ReaderSequence` | `StackedFileSliceReader` |
| `PimsImages` | `SlicedImageSource` |
| `ChunkedImages` | `ChunkedImageSource` |
| `ImsChunkedImages` / `MrcChunkedImages` | `ImsImageSource` / `MrcImageSource` |
| `get_frame` / `frame_shape` | `get_slice` / `slice_shape` |

One wart worth naming: `slice` now means both our 2D plane and Python's builtin, which `_SlicedView.__getitem__(key: int | slice)` uses. Unambiguous in context, but real.

## 3. The `ImageSource` parent

`SlicedImageSource` and `ChunkedImageSource` already had an identical surface everywhere `image_conversion.py` touches them — `dtype`, `num_channels`, `channel`, `get_possible_layers()`, `expected_bbox`, and a per-chunk read that differed **in name only** (`copy_to_view` vs `read_chunk`; same signature, same return, both consumed via `named_partial`). They now share `ImageSource`, and the method is `copy_chunk_to_view` on both.

This is also what fixed `chunked_image_source.py`'s references to its sibling. Ten of the thirteen were not comparisons but a *shared contract* with nowhere to live — the return shape, `order="F"`, the flip conventions — so they pointed sideways. They now point at the parent that states it.

It does **not** collapse `_prepare_chunked_image_conversion` / `_prepare_sliced_image_conversion`. Those differ for a real reason — exact bbox vs. placeholder-and-correct — and are now `initial_layer_bounding_box` / `chunk_grid` / `final_bounding_box` on the source, so the caller has no branch.

## 4. CZI as a chunked source

CZI stores tiled subblocks and states its exact extents in metadata, which is exactly what `ChunkedImageSource` is for. `read(roi=…)` now fetches just the box a chunk needs, one z-plane at a time, instead of decoding whole planes into a slice writer. `expected_bbox` is exact, so no placeholder is written and nothing is corrected afterwards, and multi-timepoint files get a real `t` axis.

With CZI off the sliced path, `SlicedImageSource` loses `czi_channel` entirely; the option now reaches the reader through `ReadOptions.format_options` (see 5).

Missing `pylibCZIrw` now names the `webknossos[czi]` extra instead of silently dropping `.czi` from the supported suffixes, since `CziImageSource` registers as an optional chunked reader.

## 5. One factory, one options object

Three things the first pass left behind, each a concept the conversion path did not need:

- **`open_image_source(images, options)`** is now the only way a source is created. `try_open_chunked_image_source` and the four "try chunked, else fall back to sliced" sites in `image_conversion.py` are gone, along with `has_image_z_dimension`; the conversion driver no longer imports either strategy.
- **`ReadOptions`** replaces the `channel, swap_xy, flip_x, flip_y, flip_z, is_segmentation` bundle, which was spelled out twelve times across eight files. Format-specific knobs (`czi_channel`) live in `format_options`, so `class_open_kwargs` and the filtering it drove are gone: a reader reads the names it knows and ignores the rest. `with_layer_selection()` applies one combination from `get_possible_layers()` for the re-open loop.
- **`is_segmentation` is gone**, because nothing read it. Its only consumer was `_allow_channels_first` in the axis guessing, which was already unreachable: `SingleImageSliceReader` short-circuits it (a single image's leading axis is always 1) and `MultiImageSliceReader` forced it False.

## 6. Every reader names its own axes

`SlicedImageSource.__init__` had two axis-inference schemes: one for readers that name their dimensions, and a fallback that guessed an axis order from the raw shape of an unlabelled array — 2D/3D/4D/5D cases, `_assume_color_channel`, and a pinned timepoint. The fallback existed for exactly two readers, `SingleImageSliceReader` and `MultiImageSliceReader`, which decode the everyday raster formats and always produce `(y, x)` or `(y, x, c)`. They now say so.

Guessing was never reliable, because the leading axis of a flat shape is ambiguous — a sequence axis and a channel axis look identical — and only the reader knows which it has. Declaring it removes the ambiguity instead of resolving it:

- `SliceReader` unifies what were two classes, `SliceSequence` and `NDSliceSequence`, since no flat reader is left.
- `_open_slice_reader()` yields a `SliceReader` instead of `Any`. That was the one type hole left after the pims removal, and it existed only because the yielded object could also be a list wrapping a reader, or a bare ndarray. The list-wrapping of 2D images and the `images[timepoint]` pinning go with it, as do five `isinstance` branches.
- `expected_bbox` goes from ~75 lines over four shape/axis combinations to one path reading `sizes`.
- `_iter_loop_size` becomes `SliceReader.flat_index()`, the inverse of the index arithmetic `get_slice()` already did — the two were separate implementations of the same layout, in different modules.

**The axis resolver went with it.** The ported `_make_get_slice` ranked candidate reader methods across three sort passes by iteration and read cost. Every reader declares exactly one method (`TiffSliceReader` `"cyx"`/`"yx"`, DM3/DM4 `"zyx"`/`"yx"`; `StackedFileSliceReader` overrides the setter and never reaches it), so the ranking had no observable effect, and since the requested axes are always a subset of the declared ones, `_bundle` was unreachable and `_transpose` was the identity. What is left is: index away what was not asked for, then transpose. Asking for an axis the reader cannot produce now raises rather than silently looping.

## 7. The source says whether its channels are RGB

`_channels_are_one_rgb_layer` decided by re-deriving the reader from the path suffix — a third dispatch mechanism next to the registry and `open_slice_reader` — and reached into `SingleImageSliceReader.supported_file_extensions()` to do it. `ImageSource.channels_are_rgb` says it directly. `image_conversion.py` no longer maps suffixes to readers at all; the RGB formats named in the error message are their own explicit `_RGB_IMAGE_EXTENSIONS` constant, which is the one place a user-facing list of them belongs.

`MultiImageSliceReader` sets the flag from the files it actually opened rather than from its own class, because imageio also decodes formats it is not the preferred reader for: a directory of TIFFs ends up there, and a `c` axis in those is separate acquisitions.

## Breaking changes

- **Formats only reachable through pims's own bundled readers can no longer be converted**: `.jp2`/`.j2k`, `.webp`, `.ppm`/`.pgm`/`.pbm`, `.tga`, `.sgi`, `.exr`, `.hdr`, `.psd`, `.dds`, `.pcx`, `.ras`, `.xbm`/`.xpm` (pims `ImageIOReader`); `.lsm`, `.stk` (pims `TiffStack`); `.cine`, `.seq`, `.spe`. TIFF, CZI, DM3/DM4, `.png`/`.jpg`/`.jpeg`/`.gif`/`.bmp`/`.ico`, `.ims` and MRC are unaffected. The `scifio-test.jp2` fixture case is removed from `TEST_IMAGES_ARGS`.
- **`add_layer_from_images` no longer accepts a `pims.FramesSequence`** — paths, glob strings and sequences of paths only.
- **`tifffile` is no longer in the base install.** `scm-pims` required it unconditionally, so it happened to be present; TIFF conversion now needs the extra it was always declared under (`webknossos[tifffile]` / `[all]`).
- **A CZI dimension other than X/Y/Z/T/C with length > 1 is refused** (`UnsupportedImageDataError`) rather than written to an extra bbox axis: rotation, illumination, phase, view, block. Refusing beats silently pinning. **Scenes are unaffected** — they are not a dimension but regions inside `total_bounding_rectangle`, so multi-scene files convert exactly as before.
- **`Bgra32` now maps to `uint8`** instead of `"<4u1"`, a subarray dtype `add_layer` could not have used. Almost certainly a latent bug, but there is no Bgra fixture to prove it.

Video and multi-frame GIF were already unreachable before this PR (no `imageio-ffmpeg` in the lock, and `ImageReader`'s priority 12 beat `ImageIOReader`'s 6 for `gif`), so nothing changes there.

## Behaviour changes from the cleanup

Three, all in the same direction, plus one thing that was unspecified rather than changed. All four are in the changelog under Fixed.

- **`add_layer_from_images(<directory of RGB pngs>)` converts instead of raising.** It used to fail with "cannot convert the 3 channels of these images into a single layer", because a directory has no suffix to match against the RGB formats. The same images passed as a glob or as a list converted fine, and `from_images` passes lists, so only the direct call was affected.
- **`expected_bbox` reports the real x/y for RGB input.** A 10×8 RGB png reported `(3, 10, 1)` — the channel count as the x extent. Downstream this was harmless, since `initial_layer_bounding_box` overwrites x/y with a placeholder and `final_bounding_box` replaces it with what was written, but it did make the "some images are larger than expected" warning fire on every RGB image.
- **Small images are no longer reinterpreted.** A 3-pixel-wide grayscale image was read as a 1×1 RGB one, and a 1-row RGB image as channels-first, both from `_assume_color_channel` misfiring on a real extent.

The fourth was already unspecified rather than changed: iter-axis order came from `list(set(...))`, so with two or more non-z iter axes the flat slice layout depended on string hash randomization. It is sorted now.

## Interaction with #1496

The RGB-format list in `UnsupportedImageDataError`'s message follows #1496's approach — an explicit `_RGB_IMAGE_EXTENSIONS` constant, not derived from a reader class — but with its membership corrected to what this PR actually supports: `jp2`, `j2k`, `webp`, `ppm`, `pgm`, `pbm`, `tga`, `targa` and `pcx` are dropped, since this PR makes them unreadable and the message would otherwise have advised users to reach for formats that no longer work.

#1496's `_classify_open_failure` / `_classify_read_failure` carry over unchanged in behavior, retargeted at the new names. Their reasoning holds verbatim: `open_slice_reader` also raises `UnknownFormatError` both when no reader claims the suffix and when every reader that claimed it failed.

## Other changes

- `add_layer_from_images` accepts any `Sequence` of paths, not only a `list`. mypy rejected the natural `list[UPath]` argument once the union stopped containing an untyped `Any` — `list` is invariant. Two existing tests hit this.
- Glob/directory results are ordered with `natsort`, consistently with how `from_images` already sorted them.
- The two `warnings.filterwarnings` blocks in `from_images` — four filters naming reader modules by string literal, which a rename would silently break — are one `_quiet_reader_warnings()` context manager taking the names from the modules themselves.

## Testing

**164 passed, 8 skipped** across `test_add_layer_from_images.py`, `test_from_images.py` and the new `test_slice_reader.py`, unchanged across every commit here. ruff and mypy clean (`webknossos`, `tests`, `examples`, `script_collection`).

- **`test_slice_reader.py`** (24 tests) covers what used to be a third-party library's tested behavior: axis bookkeeping, the adapter's drop/transpose paths and its refusal of axes a reader cannot produce, `_SlicedView` slicing/reversal/chaining/laziness, dispatch and priority, natural ordering, and zip archives. It caught a real bug during development — `MultiImageSliceReader.__del__` raised `AttributeError` when `__init__` aborted before `_zipfile` was set.
- **CZI**: all 32 combinations of `czi_channel` × `swap_xy` × `flip_x` × `flip_y` × `flip_z` match a pylibCZIrw reference voxel for voxel on the real fixture, at 32³ shards over 512×512×30 — heavily multi-shard, where the flip bugs fixed in #1495 lived. New CZI tests build genuine `.czi` files with pylibCZIrw's writer, so they run in CI without a download; the rejected-dimension test uses a stand-in reader, because the writer silently ignores extra dimensions.
- The 8 Linux-CI-only download cases (dm4, dm3 single **and list** — the only `StackedFileSliceReader` coverage — dnasample1.dm3, czi, gif, jpg, png) were driven directly on a dev machine after each of the first four commits. **They have not been exercised since**: the last GitHub Actions run on this branch was for `dc754cc1`, and no run has been created for any commit after it, so commits 5–7 and everything after them are untested against those fixtures. The axis work in 6 touches `StackedFileSliceReader`, so the dm3-list case is the one to watch once CI runs again.
- `tests/dataset/test_dataset.py` was not run here: it needs Docker/minio and a server on `localhost:9000`. It is untouched by this PR.


webknossos/webknossos/dataset/_utils/image_source.py
```diff
@@ -0,0 +1,136 @@
...
+
+from numpy.typing import DTypeLike
+
+from ...geometry.nd_bounding_box import NDBoundingBox
+from ..layer.view import MagView
+
+
+def compute_channel_selection(
+    raw_num_channels: int, channel: int | None
+) -> tuple[int, int | None, int | None, list[int] | None]:
```
normanrz: The return type should be a namedtuple or dataclass/attr.

webknossos/webknossos/dataset/_utils/image_source.py
```diff
@@ -0,0 +1,136 @@
...
+        the sizes reported by `copy_chunk_to_view`.
+        """
+
+    @abstractmethod
+    def copy_chunk_to_view(
+        self,
+        bbox: NDBoundingBox,
+        mag_view: MagView,
+        dtype: DTypeLike | None = None,
+    ) -> tuple[tuple[int, int], int | None]:
```
normanrz: also namedtuple

webknossos/webknossos/dataset/_utils/slice_sequence.py
```diff
@@ -0,0 +1,411 @@
...
+Only the behavior the conversion path actually relies on is implemented:
+
+* random access to numbered slices, plus lazy slicing of a reader,
+* named axes with `bundle_axes` / `iter_axes` / `default_coords`, and the
+  resolver that maps a reader's registered `get_slice` onto whatever axes the
+  caller asked for.
+
+Unlike pims, slices are plain `np.ndarray`s — there is no `pims.Frame` wrapper
+and no per-slice metadata, because nothing downstream reads it.
+"""
```
normanrz: Remove the pims references

webknossos/Changelog.md
```diff
@@ -14,6 +14,8 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
...
+- Replaced the `scm-pims` dependency with an in-library implementation of the parts this library used, and dropped `scm-pims` and `slicerator`. Two consequences: formats that were only reachable through readers `pims` bundled can no longer be converted — `.jp2`/`.j2k`, `.webp`, `.ppm`/`.pgm`/`.pbm`, `.tga`, `.sgi`, `.exr`, `.hdr`, `.psd`, `.dds`, `.pcx`, `.ras`, `.xbm`/`.xpm`, `.lsm`, `.stk`, `.cine`, `.seq` and `.spe` — and `Dataset.add_layer_from_images` no longer accepts a `pims.FramesSequence` instance, only paths, glob strings and sequences of paths. TIFF, CZI, DM3/DM4, `.png`/`.jpg`/`.jpeg`/`.gif`/`.bmp`/`.ico`, `.ims` and MRC are unaffected. [#1498](https://github.com/scalableminds/webknossos-libs/pull/1498)
```
normanrz: The changelog should focus on the user-facing effects
```suggestion
- Support for a number of conversion formats has been removed: `.jp2`/`.j2k`, `.webp`, `.ppm`/`.pgm`/`.pbm`, `.tga`, `.sgi`, `.exr`, `.hdr`, `.psd`, `.dds`, `.pcx`, `.ras`, `.xbm`/`.xpm`, `.lsm`, `.stk`, `.cine`, `.seq` and `.spe`. [#1498](https://github.com/scalableminds/webknossos-libs/pull/1498)
- 
- `Dataset.add_layer_from_images` no longer accepts a `pims.FramesSequence` instance, only paths, glob strings and sequences of paths. [#1498](https://github.com/scalableminds/webknossos-libs/pull/1498)
```

webknossos/webknossos/dataset/_utils/image_reader_registry.py
normanrz, line 1: Rename to image_source_registry.py and combine with registry for chunked_image_source.py

webknossos/webknossos/dataset/_utils/image_source.py
normanrz: Now returns a `ChannelSelection` NamedTuple (`num_channels`, `selected_channel`, `first_n_channels`, `possible_channels`), each field documented on the class rather than in the function docstring.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_utils/image_source.py
normanrz: `copy_chunk_to_view` now returns a `ChunkResult` NamedTuple (`xy_size`, `max_value`). The consumer in `image_conversion.py` reads the fields by name instead of `zip(*...)`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_utils/slice_sequence.py
normanrz: Already done in `ee23daae`, which landed after you commented — all pims references are out of the source now (only `Changelog.md` keeps them). That commit also caught two that were wrong: an earlier rename had turned `pims.ReaderSequence` into `pims.StackedFileSlices`, a class that never existed.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/Changelog.md
normanrz: Applied. Split into separate user-facing bullets, dropped the stray empty one, and added a third for the `.czi` extra-dimension rejection which belonged under Breaking Changes too. The two `Changed` entries are now phrased as effects (faster/less memory, names the extra to install) rather than implementation.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_utils/image_reader_registry.py
normanrz: Done — `image_source_registry.py` now holds both registries; `chunked_image_source.py` is left with just the ABC. One import pass, one warning, one env var, and one `get_valid_suffixes()` replacing the two every caller had to union. It also closed a hole this PR opened: `_OPTIONAL_READERS` now covers slice readers, so with `tifffile` out of the base install a `.tif` names `webknossos[tifffile]` instead of silently vanishing from the supported formats.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_utils/image_source_registry.py
```diff
@@ -0,0 +1,260 @@
...
+import warnings
+from os import environ
+
+from upath import UPath
+
+from .chunked_image_source import ChunkedImageSource
+from .slice_sequence import SliceSequence
+
+
+class UnknownFormatError(Exception):
```
normanrz: Shouldn't this be one of the `ImageConversionError` from errors.py?

webknossos/webknossos/dataset/_utils/image_source_registry.py
```diff
@@ -0,0 +1,251 @@
...
+from .chunked_image_source import ChunkedImageSource
+from .image_source import ImageSource, ReadOptions
+from .slice_sequence import SliceSequence
+
+
+class UnknownFormatError(Exception):
+    """Raised when no registered reader can open a file."""
+
+
+_SLICE_READER_CLASSES: list[type[SliceSequence]] = []
```
normanrz: Why are these "readers" and not "source classes"?

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -0,0 +1,251 @@
...
+        suffix = images.suffix.lstrip(".").lower()
+        for cls in _CHUNKED_IMAGE_SOURCE_CLASSES:
+            if suffix in cls.class_exts():
+                # Remote UPaths deliberately do reach the reader, which raises
+                # its own "must be a local file path" error via is_remote_path.
+                return cls(images, options)
+    return SlicedImageSource(images, options)
+
+
+class _OptionalReader(NamedTuple):
```
normanrz, line 140: Reader?

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -0,0 +1,251 @@
...
+    dependency is missing never imports, and so cannot be asked — which is
+    exactly when this is needed, to turn "unsupported format" into "install
+    `webknossos[czi]`". test_optional_reader_suffixes_match_class_exts keeps
+    the two in sync."""
+
+
+# Every optional reader, declared once. Covers both strategies: slice readers
+# are just as optional as chunked ones now that tifffile is not in the base
+# install.
+_OPTIONAL_READERS: tuple[_OptionalReader, ...] = (
```
normanrz: Reader?

webknossos/webknossos/dataset/_utils/image_source_registry.py
normanrz: Yes — it now subclasses `UnsupportedImageFormatError`, so it is catchable as `ImageConversionError` and as `ValueError` like every other conversion failure. It stays a distinct name because `SlicedImageSource._classify_open_failure` uses it as the registry's internal "dispatch found nothing" signal and normally replaces it with a fuller error; this only changes what happens on the paths where it escapes. The three raise sites now also fill in `suffix` and `supported_suffixes` (`path` stays unset — `open_images` takes a str that may be a glob).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_utils/image_source_registry.py
normanrz: Because a `SliceSequence` subclass is not an `ImageSource` — it only produces 2D slices, and `SlicedImageSource` is the source that wraps it. A chunked one *is* a source, which is what its `…ImageSource` name says. I have made that explicit in the module docstring (reader = knows one input format, registers here; source = `ImageSource`, what `image_conversion.py` drives) and made the registry consistently say "reader": `_CHUNKED_IMAGE_SOURCE_CLASSES` → `_CHUNKED_READER_CLASSES`, `register_chunked_image_source` → `register_chunked_reader`, `get_valid_chunked_image_suffixes` → `get_valid_chunked_reader_suffixes`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
normanrz: Kept as `_OptionalReader`, now unambiguous: this table covers both kinds (`TiffSlices` alongside `CziImageSource`), so it cannot be named after either one alone — and "reader" is exactly the umbrella the docstring now defines.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
normanrz: Same as above — `_OPTIONAL_READERS` holds both slice and chunked readers, so "reader" is the correct umbrella term rather than "source".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_utils/image_source_registry.py
normanrz: Follow-up on this: rather than move it to errors.py, `UnknownFormatError` is gone in c19f4333. It turned out nothing caught it by type — `_try_open_images` catches bare `Exception`, and `_classify_open_failure` decides on the suffix — so the class existed only at its three raise sites and in one test. It was also strictly weaker than the error it would have sat beside (no `path`, since `open_images` takes a str that may be a glob; no `missing_extras`). Those sites now raise `UnsupportedImageFormatError` directly.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/tests/dataset/test_slice_sequence.py
```diff
@@ -0,0 +1,327 @@
...
+
+
+def test_open_images_dispatches_on_suffix(tmp_upath: UPath) -> None:
+    png_path = tmp_upath / "single.png"
+    _write_png(png_path, 7)
+    assert isinstance(open_images(str(png_path)), SingleImageSlices)
+
+
+def test_open_images_prefers_higher_class_priority(tmp_upath: UPath) -> None:
+    # Both TiffSlices (19) and, in principle, any lower-priority reader claim
```
fm3: 19?

webknossos/tests/utils.py
```diff
@@ -1,13 +1,23 @@
...
 import h5py
 import httpx
 import numpy as np
+import pytest
 from numpy.typing import DTypeLike
 from upath import UPath
 
+# pylibCZIrw ships no wheel past cp313, so the czi extra is uninstalled on
+# newer Pythons; tests that need it skip instead of failing.
+HAS_PYLIBCZIRW = importlib.util.find_spec("pylibCZIrw") is not None
```
fm3, line 19: should we make the python versions where it is not installed explicit so that the tests also catch the case where it should be installed but isn’t?

webknossos/webknossos/dataset/_utils/image_conversion.py
```diff
@@ -210,41 +221,18 @@ def _find_unavailable_input_formats(input_upath: UPath) -> dict[str, str]:
...
-    }
-)
-
-
 def _describe_rgb_formats() -> str:
-    return ", ".join("." + suffix for suffix in sorted(_RGB_IMAGE_SUFFIXES))
+    """The formats whose channels mean colour, for the error message naming
+    which ones can share a layer. Taken from the reader so it cannot drift into
+    naming formats that can no longer be read."""
+    return ", ".join("." + suffix for suffix in sorted(SingleImageSlices.class_exts()))
```
fm3: So it’s now just _all_ SingleImageSlices extensions?

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -479,17 +338,7 @@ def from_images(
...
-                "ignore",
-                category=UserWarning,
-                module="pims",
-            )
-            warnings.filterwarnings(
-                "once",
-                category=UserWarning,
-                module="pims_images",
-            )
+        with _quiet_reader_warnings(sliced_source_action="once"):
```
fm3, line 379: nice extraction :)

webknossos/webknossos/dataset/_utils/image_source.py
```diff
@@ -0,0 +1,212 @@
...
+
+from ...geometry.mag import Mag
+from ...geometry.nd_bounding_box import NDBoundingBox
+from ..layer.view import MagView
+
+
+@dataclass(frozen=True)
+class ReadOptions:
+    """How to read a source, as one value — every layer between
+    `add_layer_from_images` and a reader's constructor carries all of it."""
```
fm3:
```suggestion
    """How to read an `ImageSource`"""
```

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -0,0 +1,212 @@
...
+    `compute_channel_selection`'s business."""
+
+    swap_xy: bool = False
+    flip_x: bool = False
+    flip_y: bool = False
+    flip_z: bool = False
+    """See `ImageSource.copy_chunk_to_view`."""
+
+    format_options: Mapping[str, int | None] = field(default_factory=dict)
+    """Knobs only some formats understand, e.g. `czi_channel`. A source reads
```
fm3: knobs?

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -0,0 +1,212 @@
...
+
+    num_channels: int
+    """How many channels will actually be written."""
+
+    selected_channel: int | None
+    """A single pinned channel index, or None if all of them are used."""
+
+    first_n_channels: int | None
+    """Truncation to the first 3 of 3-or-more channels (e.g. RGB out of RGBA),
+    or None when none applies."""
```
fm3: Whe name says n but the comment says 3. Maybe use n in the comment too?

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -0,0 +1,212 @@
...
+        Called once per chunk, potentially in parallel across processes, so
+        implementations must not carry open file handles across the call.
+
+        Conventions every implementation shares, so that output is
+        interchangeable:
+
+        * `swap_xy` swaps which source axis feeds which output axis.
+        * `flip_x` mirrors the source's **y** axis and `flip_y` its **x** axis
+          — the axes are named for the output, not the source. Each flip
+          mirrors the whole extent, never one chunk in isolation.
```
fm3: This should reference the ReadOptions class, otherwise one wonders why these args are listed in the comment, but not in the method signature

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -0,0 +1,243 @@
...
+
+    suffixes: frozenset[str]
+    """What its class_exts() returns, restated because a reader whose
+    dependency is missing never imports and so cannot be asked — which is
+    exactly when this is needed. test_optional_reader_suffixes_match_class_exts
+    keeps the two in sync."""
+
+
+# Every optional reader, declared once. Slice readers are as optional as
+# chunked ones now that tifffile is not in the base install.
```
fm3:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/ims_image_source.py
```diff
@@ -38,39 +39,20 @@ def _read_ims_metadata_quietly(
     return shape, dtype  # type: ignore[return-value]
 
 
-@register_chunked_images
-class ImsChunkedImages(ChunkedImages):
+@register_chunked_reader
```
fm3: Should this be named `register_chunked_image_source`? What’s the difference between a source and a reader?

webknossos/webknossos/dataset/_image_conversion/ims_image_source.py
```diff
@@ -38,39 +39,20 @@ def _read_ims_metadata_quietly(
...
+class ImsImageSource(ChunkedImageSource):
     """
-    ChunkedImages implementation for Imaris .ims files. Reads shard-sized 3D
-    blocks directly from the underlying HDF5 file via h5py and writes them
-    to mag_view directly — no slice-by-slice pims reading, no BufferedSliceWriter.
-    This is the only supported way .ims files are read for conversion.
+    ChunkedImageSource for Imaris .ims files. Reads shard-sized 3D blocks
+    straight out of the underlying HDF5 file via h5py and writes them to
+    mag_view. This is the only supported way .ims files are read.
```
fm3:
```suggestion
    mag_view.
```

webknossos/webknossos/dataset/_utils/raster_slices.py
```diff
@@ -0,0 +1,257 @@
...
+def _collect_files(
+    path_spec: str | Iterable[str], sort_explicit_lists: bool
+) -> tuple[list[str], zipfile.ZipFile | None]:
+    """Resolves a directory, glob pattern, zip archive or iterable of paths
+    into the list of files to read, in the order they should be read."""
+    if not isinstance(path_spec, str):
+        filepaths = [str(i) for i in path_spec]
+        # Deliberate asymmetry: StackedFileSlices re-sorts an explicitly
+        # passed list, MultiImageSlices does not. Both orders are long-standing,
+        # and changing either would silently reorder z in existing conversions.
```
fm3: But shouldn’t we do just that if the old ordering was wrong? The argument is weak (or not well-explained)

webknossos/webknossos/dataset/_utils/raster_slices.py
```diff
@@ -0,0 +1,257 @@
...
+    if plane.ndim == 3:
+        return "yxc"
+    raise UnsupportedImageDataError(
+        f"Got a {plane.ndim}-dimensional image from {source}, but a raster "
+        + "image must decode to (y, x) or (y, x, channels)."
+    )
+
+
+@register_slice_reader
+class SingleImageSlices(SliceSequence):
```
fm3: Okay now those are slice sequences and also slice readers but not sources? I’m starting to get a little lost here.

webknossos/webknossos/dataset/_image_conversion/common_slice_readers.py
```diff
@@ -0,0 +1,257 @@
+"""Readers for common 2D raster images, and for sequences of image files.
```
fm3, line 1: not super happy with the file name, aren’t _all_ our inputs raster images? Why introduce this term now? I know this word only in contrast to e.g. vector images. So maybe just common_image_slices or something?

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -0,0 +1,508 @@
...
+)
+from .raster_slices import MultiImageSlices, StackedFileSlices
+from .slice_sequence import SliceSequence, _SlicedView
+
+# The x/y extent is only discovered while reading, so the layer starts out
+# deliberately oversized and is cut back down by final_bounding_box().
+SAFE_LARGE_XY: int = 10_000_000_000  # 10 billion
+
+
+class SlicedImageSource(ImageSource):
```
fm3, line 43: This one seems important. Maybe it should have a docstring? Many other things that seem less important do have one 😅

webknossos/webknossos/dataset/_utils/sliced_image_source.py
```diff
@@ -0,0 +1,508 @@
...
+        image_paths: UPath | list[UPath],
+        options: ReadOptions,
+    ) -> None:
+        """
+        Sorts the reader's axes into the two lists that shape every slice:
+        `_iter_axes`, what the sequence steps through (empty for a single 2D
+        image), and `_bundle_axes`, what one slice is — "y" and "x", preceded
+        by "c" when several channels are written.
+
+        Only sorting, not guessing: every reader names its own axes.
```
fm3: That’s all this `__init__` does?

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -0,0 +1,508 @@
...
+                self._iter_axes.append("z")
+            self._bundle_axes = bundle_axes
+
+        self.num_channels, self._channel, self._first_n_channels, possible_channels = (
+            compute_channel_selection(raw_num_channels, self._channel)
+        )
+        if possible_channels is not None:
+            self._possible_layers["channel"] = possible_channels
+
+    def _normalize_original_images(self) -> str | list[str]:
```
fm3: From this name I would think about intensity normalization, not paths. Maybe the name could be improved

webknossos/webknossos/dataset/_utils/sliced_image_source.py
```diff
@@ -0,0 +1,508 @@
...
+                        }
+                    )
+                    upper_bounds = lower_bounds + mag_view.bounding_box.get_shape("z")
+                    slices = images[lower_bounds:upper_bounds]
+                if self._options.flip_z:
+                    slices = slices[::-1]
+
+                with mag_view.get_buffered_slice_writer(
+                    # Previously only z_start and its end were important, now the slice writer needs to know
+                    # which axis is currently written.
```
fm3: previously as in above in this function? Or before this commit?

fm3 (changes requested):
I’m afraid this one got a bit too complex for me. Could you explain again what exactly is the difference between a (Slice)Sequence, a Reader and a Source? And why do we need all three?

Very happy we resolved the frame/slice naming doubling. Again the colour/color/rgb naming as well as ext/extension/suffix thing stood out.

The dataset/_utils package is getting cramped. Maybe extract to something more explicit like image_conversion?

My slight confusion may shine through in some of the comments below

webknossos/webknossos/dataset/_image_conversion/ims_image_source.py
normanrz: This is the way.

webknossos/webknossos/dataset/_utils/raster_slices.py
normanrz: These are all `SliceReader`s now

webknossos/webknossos/dataset/_image_conversion/common_slice_readers.py
normanrz: The module is now named `common_slice_readers`

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
normanrz: Changed to `_resolve_original_image_paths`

webknossos/webknossos/dataset/_utils/sliced_image_source.py
normanrz: Gone.

webknossos/tests/utils.py
normanrz: Fixed in https://github.com/scalableminds/webknossos-libs/pull/1498/commits/35ffe1b5393e2b7678e7ca3074b5b780e63f43f2. Hopefully, we'll soon get new builds that go all the way to Python 3.15.

fm3 (approved):
testdata directory should be removed, and assuming the changes from the other PRs land here cleanly, and pending CI, I’d say it’s good to go now too

PR #1502, opened 2026-08-13
Add Layer.export API: as_ozx (RFC-9 OME-Zarr zip), as_tiff_stack, as_ome_tiff

normanrz:
### Description:
Adds `layer.export`, a small accessor on `AbstractLayer` (so it works for local/remote and color/segmentation layers alike) with three methods to export a layer's data — optionally cropped to a `bounding_box` and/or restricted to a `mag` — to common bioimaging formats:

- **`as_ozx`** — a single zipped OME-Zarr archive compliant with [NGFF RFC-9](https://ngff.openmicroscopy.org/rfc/9/) ("Zipped OME-Zarr"). Materializes a Zarr3 copy of the requested region/mags in a temp directory, then packages it into an RFC-9 compliant zip: root `zarr.json` first, all `zarr.json` metadata entries in breadth-first order before any chunk/shard data, `ZIP_STORED` compression (Zarr already compresses chunks), forced `ZIP64`, and the RFC-9 archive comment. If `mag` is given, the archive contains that mag plus every coarser mag already present on the layer; if `mag` is `None`, the full pyramid is exported. Shard shape is picked per mag to just cover the exported region (searching upward from the default chunk shape, capped at the default shard shape — or the wider from-images default for flat/low-z data), or can be fixed via the new `shard_shape` argument. This avoids depending on the source layer's own shard shape, which would otherwise pad a cropped export's declared extent (webknossos Zarr arrays always span from voxel 0) far beyond what was actually exported.
- **`as_tiff_stack`** — a directory of per-slice TIFF files (`NNNNNN.tiff`, or `{filename_prefix}_NNNNNN.tiff`), digit-width sized to the actual slice count.
- **`as_ome_tiff`** — a single pyramidal OME-TIFF file.

Both TIFF variants are adapted from voxelytics' `export_tiff.py` reference implementation, simplified into synchronous, no-executor methods that write directly to a `UPath` (local or remote) — `output_path` accepts `str | PathLike | UPath` on all three methods.

The earlier standalone `examples/export_ome_zarr_zip.py` example (from an earlier iteration of this work) is simplified to call `layer.export.as_ozx(...)` directly instead of hand-rolling the zip packaging.

### Issues:
- N/A

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation (docstrings + `docs/mkdocs.yml` nav entry added; no additional narrative docs page)
 - [x] Added / Updated Tests
 - [x] Considered adding this to the Examples (existing `export_ome_zarr_zip.py` example updated to use the new API)


webknossos/webknossos/dataset/layer/export.py
normanrz, line 1: Will replace a lot of code in https://github.com/scalableminds/voxelytics/blob/master/voxelytics/worker/jobs/export_tiff.py

webknossos/tests/dataset/test_layer_export.py
```diff
@@ -0,0 +1,119 @@
...
+def test_as_ozx_single_mag(tmp_upath: UPath) -> None:
+    _dataset, layer, _data = make_layer(tmp_upath)
+    zip_path = tmp_upath / "color_mag2.ozx"
+
+    layer.export.as_ozx(output_path=zip_path, mag=layer.get_mag("2-2-1").mag)
+
+    with zipfile.ZipFile(str(zip_path)) as zip_file:
+        names = zip_file.namelist()
+        assert "1/zarr.json" not in names
+        assert "2-2-1/zarr.json" in names
```
Tobias314, line 94: one could assert that there is only one folder with a zarr.json in it to make it more robust (e.g. making sure there is also no 1-1-1 etc.)

webknossos/tests/dataset/test_layer_export.py
Tobias314, line 1: For the tiff tests no different mags are tested. I guess depends on how detailed we want the tests to be. But might be doable by parameterizing the existing tests to test both mag=None and mag =Mag(2,2,2)

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+# tiff-export implementation.
+OME_TIFF_DOWNSAMPLING_STEPS = 5
+OME_TIFF_DOWNSAMPLING_MIN_PIXEL = 16
+
+
+def _resolve_export_bbox(
+    layer: "AbstractLayer", bounding_box: NDBoundingBox | None
+) -> NDBoundingBox:
+    bbox = bounding_box if bounding_box is not None else layer.bounding_box
+    return bbox.intersected_with(layer.bounding_box)
```
Tobias314:
```suggestion
    return layer.bounding_box if bounding_box is None else bounding_box.intersected_with(layer.bounding_box)
```

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+    additional_axes_indices = []
+    channel_offset = 1
+    axes = mag_view.bounding_box.axes
+    if is_single_section:
+        # z is already removed from the shape of single-section data.
+        axes = tuple(axis for axis in axes if axis != "z")
+    for index, axis_name in enumerate(axes):
+        if axis_name not in ("x", "y", "z"):
+            additional_axes_indices.append(index + channel_offset)
+    # All additional axes only extend by one value, so they can be dropped.
```
Tobias314: why is this, should we assert this if this is always the case?

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+        data_slice = data_slice.transpose((2, 1, 0))
+    return data_slice
+
+
+def _squeeze_nd_data_to_3d(
+    data: np.ndarray,
+    bbox: NDBoundingBox,
+    mag_view: "MagView",
+    is_single_section: bool,
+) -> np.ndarray:
```
Tobias314: Not sure whether I get this: So we always want to silently convert nd to 3d? Also, we get the axes both from the mag_view.bbox as well as from a given bbox. Can they differ?
This seems to assume that all axes other than x,y,z have size 1 and can be squeezed. Why is this?

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+                data_format="zarr3",
+                bounding_box=bbox,
+                largest_segment_id=layer._get_largest_segment_id_maybe(),
+            )
+            for target_mag in target_mags:
+                source_mag_view = layer.get_mag(target_mag)
+                local_bbox = bbox.align_with_mag(target_mag)
+                if local_bbox.is_empty():
+                    continue
+                data = source_mag_view.read(absolute_bounding_box=local_bbox)
```
Tobias314: Doesn't this load a full mag into memory which can cause OOM issues for large datasets?

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+        """
+        output_path = UPath(output_path)
+        fixed_shard_shape = (
+            Vec3Int.from_vec_or_int(shard_shape) if shard_shape is not None else None
+        )
+        layer = self._layer
+        bbox = _resolve_export_bbox(layer, bounding_box)
+        target_mags = sorted(m for m in layer.mags if mag is None or m >= mag)
+
+        with TemporaryDirectory() as tmpdir:
```
Tobias314, line 262: Do we always need to go the extra step via the tmp dataset? Couldn't we zip the input directly if we find that the underlying data is already stored using zarr and that we do not need to change the chunk/shard shape?

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+        # z is already removed from the shape of single-section data.
+        axes = tuple(axis for axis in axes if axis != "z")
+    for index, axis_name in enumerate(axes):
+        if axis_name not in ("x", "y", "z"):
+            additional_axes_indices.append(index + channel_offset)
+    # All additional axes only extend by one value, so they can be dropped.
+    return data.squeeze(axis=tuple(additional_axes_indices))
+
+
+def _write_ome_zarr_zip(layer_dir: UPath, output_path: UPath, ome_version: str) -> None:
```
Tobias314, line 165: Might this be IO-heavy enough to profit from multi-threading?
But we always have to write sequential, I guess, so we could only parallelize reading which might only make sense in certain cases e.g. if the input is in S3 and output on fast local storage. Maybe something for a follow-up PR.

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+                data,
+                subifds=len(downsampling_factors),
+                resolution=(voxel_size_nm[0] / 1e7, voxel_size_nm[1] / 1e7),
+                metadata=metadata,
+                resolutionunit="CENTIMETER",
+                compression=compression,
+            )
+            for factor in downsampling_factors:
+                tif.write(
+                    data[:, :, ::factor, ::factor],  # cheap downsampling
```
Tobias314: I guess there is a reason why we do the "cheap downsampling" even though this might result in downsampled versions that are unrecognizable if the data is noisy?

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,356 @@
...
+        mag_view = _resolve_export_mag(layer, mag)
+        bbox = _resolve_export_bbox(layer, bounding_box).align_with_mag(mag_view.mag)
+        compression = "zlib" if layer.category == SEGMENTATION_CATEGORY else None
+
+        voxel_size_nm = (
+            layer.dataset.voxel_size_with_unit.to_nanometer().to_np()
+            * mag_view.mag.to_np()
+        )
+
+        data = mag_view.read(absolute_bounding_box=bbox)
```
Tobias314, line 442: This might also read the full mag if no bbox is given, resulting in OOM for large datasets? I guess, this is fine here, since we always need to write the full tiff in one go and tiffs are expected to fit into memory anyway?
So calling this for huge datasets makes no sense, right?

Maybe calling code (e.g. in the CLI interface) should to guard against this?

webknossos/webknossos/dataset/layer/export.py
Tobias314: Great! Is the plan to delete the old code in a subsequent PR?
Also, why do we get rid of the executor. Doesn't parallel execution make sense when writing a tiff stack?

Tobias314 (commented):
Nice, I have added some comments and questions.

webknossos/webknossos/dataset/layer/export.py
normanrz: Yeah, I think that would be for later.

webknossos/webknossos/dataset/layer/export.py
normanrz: We first normalize to 3D and later remove axes again in `_slice_to_image`. Tiff has some requirements about the axes, so normalizing first makes it easier.

webknossos/webknossos/dataset/layer/export.py
normanrz: Fwiw, the worker already has an export limit. If users run that on their own, they should make their own choice of how big they want to export.

webknossos/webknossos/dataset/layer/export.py
normanrz: Fixed with chunked copy.

webknossos/tests/dataset/test_layer_export.py
normanrz: Well, if `mag` is present, that mag and all coarser mags are exported.

webknossos/webknossos/dataset/layer/abstract_layer.py
```diff
@@ -350,6 +352,16 @@ def name(self) -> str:
 
         return self._name
 
+    @property
+    def export(self) -> LayerExport:
+        """Export this layer's data to common bioimaging file formats.
+
+        Returns:
+            LayerExport: Accessor object with `as_ozx`, `as_tiff_stack` and
+                `as_ome_tiff` export methods.
```
normanrz:
```suggestion
            LayerExport
```

webknossos/webknossos/dataset/layer/export.py
normanrz: I'd schedule that for a followup.

webknossos/webknossos/dataset/layer/export.py
normanrz: In the worker, the tiff export did not parallelize with an executor. It just spawned one job. We could parallelize the tiff writing in a follow-up. Tiled writing would also be interesting.

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -0,0 +1,493 @@
...
+                bounding_box=target_bbox,
+                largest_segment_id=layer._get_largest_segment_id_maybe(),
+            )
+            for target_mag in target_mags:
+                source_mag_view = layer.get_mag(target_mag)
+                # ceil=True: a coarser mag's own array always covers the
+                # ceil-of-mag-factor extent of the layer's bounding box (see
+                # Layer.add_mag), so floor-aligning here would silently
+                # truncate the last row/column/slice of real, already
+                # downsampled data at that mag.
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/layer/export.py
normanrz: This actually was an issue: true ND datasets weren't supported. Fixed in https://github.com/scalableminds/webknossos-libs/pull/1502/commits/3fe74eda5d1ba15674cd68ee45c54a5622d10542

webknossos/webknossos/dataset/layer/export.py
normanrz: Refactored now in https://github.com/scalableminds/webknossos-libs/pull/1502/commits/3fe74eda5d1ba15674cd68ee45c54a5622d10542

webknossos/webknossos/dataset/layer/export.py
normanrz: Nobody has complained so far :D

webknossos/tests/dataset/test_layer_export.py
```diff
@@ -24,6 +26,29 @@ def make_layer(dataset_path: UPath) -> tuple[Dataset, Layer, np.ndarray]:
...
+    layer = dataset.add_layer(
+        "color",
+        COLOR_CATEGORY,
+        dtype="uint8",
+        data_format="zarr3",
+        bounding_box=BoundingBox((0, 0, 0), (64, 64, 64)),
+    )
+    data_by_mag = {}
+    for mag_name, size in [("1", (64, 64, 64)), ("2-2-2", (32, 32, 32))]:
+        data = (np.random.rand(*size) * 255).astype(np.uint8)
```
Tobias314, line 48: nit: use a seed for randomness in tests to prevent any chance of flaky tests

webknossos/tests/dataset/test_layer_export.py
```diff
@@ -67,22 +92,63 @@ def test_as_ozx_single_mag(tmp_upath: UPath) -> None:
...
 
 
-def test_as_tiff_stack_pixel_values(tmp_upath: UPath) -> None:
+def test_as_ozx_odd_sized_layer_covers_full_coarser_mags(tmp_upath: UPath) -> None:
+    """Regression test: a coarser mag's own array always covers the
+    ceil(mag-factor)-of-the-layer-size extent (see Layer.add_mag), so a
+    layer whose size isn't an exact multiple of a mag's factor must still
+    export that mag's full, real data - not silently truncate the last
+    row/column/slice, which floor-aligning the exported region used to do.
+    """
```
Tobias314: the test asserts that the exported data is the same as the layer data, which is good.
But it does not acutally assert ceiling the size as written in the docstring. So e.g. if `.downsample()` no longer ceils, the test would still pass. I think, this is fine. But maybe change the docstring/comment.

webknossos/webknossos/dataset/layer/export.py
```diff
@@ -250,9 +341,14 @@ def as_tiff_stack(
...
         filename_prefix: str = "",
     ) -> None:
-        """Exports the layer as a directory of per-slice TIFF files, one file
-        per z-section, under `output_path`. Files are named `NNNNNN.tiff`, or
-        `{filename_prefix}_NNNNNN.tiff` if `filename_prefix` is given.
+        """Exports the layer as a directory of per-slice TIFF files, one
+        file per z-section (and per combination of any additional axes,
+        e.g. time), under `output_path`.
+
+        Files are named `NNNNNN.tiff` (or `{filename_prefix}_NNNNNN.tiff`)
```
Tobias314: is "NNNNNN" an obvious way of describing this or might it help to add something like "with NNNNNN being the 0-padded z-index"

Tobias314 (approved):
Nice, LGTM, only added some minor comments

webknossos/webknossos/dataset/layer/export.py
normanrz:
```suggestion
        Files are named `{NNNNNN}.tiff` (or `{filename_prefix}_{NNNNNN}.tiff`)
```

PR #1504, opened 2026-08-14
Add converters for Zarr, OME-Zarr, N5, and Neuroglancer precomputed

normanrz:
Stacked on #1498 ("Internalize the pims dependency") — please review/merge that first. This PR targets its branch rather than `master` so the diff here only shows what's new.

## What

Adds four new read-only input formats to `Dataset.from_images`/`add_layer_from_images`, using the `ChunkedImageSource` framework #1498 introduces (its own docstring already names `zarr, n5` as the intended next subclasses):

- **Plain Zarr arrays** — v2 (`.zarray`) and v3 (`zarr.json`).
- **OME-Zarr multiscale groups** — NGFF 0.4 (on Zarr v2) and NGFF 0.5 (on Zarr v3).
- **N5** — plain datasets and the common `s0`/`s1`/... multiscale pyramid convention (Neuroglancer/BigDataViewer-N5/n5-utils/paintera).
- **Neuroglancer precomputed** volumes, which are inherently multiscale.

For any multiscale source, **the finest resolution is picked automatically** (by comparing actual resolution/scale factors, not list position) and written as `Mag(1)`; WEBKNOSSOS's existing downsampling (`Layer.downsample()`) builds the rest of the mag pyramid afterward — the same two-step flow every other converter already uses, so no changes were needed there. A specific level can still be requested via `format_options={"scale": <index>}`, the same mechanism `czi_channel` already uses, and `allow_multiple_layers=True` converts every level as its own layer for free.

## How

All four formats are built on `tensorstore`, which is already a base dependency (used elsewhere for webknossos's own dataset storage — `Zarr2Array`/`Zarr3Array`/`N5Array`/`NeuroglancerPrecomputedArray` in `_array.py`), so **no new dependency or `pyproject.toml` extra is needed**; the new readers are registered unconditionally, like `raster_slices`/`dm_slices`.

- `_utils/tensorstore_chunked_source.py`: a shared `_TensorStoreChunkedImageSource` base. Subclasses resolve a tensorstore open spec and an axis-role mapping (`t`/`c`/`z`/`y`/`x` → physical dimension); the base reopens the array fresh per chunk (required — chunks may run in separate processes) and does the generic read + axis permutation. Axis roles come from labeled dimensions when available (Zarr v3 `dimension_names`, or an OME-Zarr group's own `axes` metadata) and fall back to OME-NGFF's canonical positional order otherwise.
- `_utils/zarr_image_source.py`, `_utils/n5_image_source.py`, `_utils/neuroglancer_precomputed_image_source.py`: the three format-specific readers — metadata parsing, multiscale-level discovery/selection, building the tensorstore open spec.
- Unlike `.ims`/MRC/`.czi`, these readers don't reject remote paths — `tensorstore`'s kvstore already handles `s3://`/`gs://`/`http(s)://` the same way it does for webknossos's own dataset storage.

Since these formats are directories rather than single files, two small extensions to the existing framework:
- `ChunkedImageSource` gains an optional `probe_directory()` hook, consulted when a directory's suffix matches no registered reader — this is what lets a bare (unsuffixed) Zarr/N5 store be recognized by its contents, and is the *only* way a neuroglancer precomputed volume is recognized at all (it has no suffix convention).
- `image_conversion.py`'s directory walk now treats a directory some chunked reader claims as one atomic leaf item — the same way it already treats a single `.ims` file — instead of globbing into its internal chunk files.

## Testing

New `tests/dataset/test_tensorstore_image_sources.py` (18 tests) constructs each reader directly against small **real** stores written with `tensorstore` itself (no monkeypatching needed) — plain arrays, multiscale groups/pyramids with levels listed out of finest-first order (to verify selection goes by resolution, not position), `probe_directory` for bare stores, and error paths. Plus end-to-end `add_layer_from_images`/`from_images` tests per format in the existing `test_add_layer_from_images.py`/`test_from_images.py`, including a directory-walk test and a downsample-afterward check.

`./precommit.sh` clean; full `test_from_images.py`/`test_add_layer_from_images.py`/`test_tensorstore_image_sources.py` suite passes (146 passed, 22 deselected — the slow network-fixture-download tests).


normanrz:
Closing in favor of #1512

PR #1505, opened 2026-08-14
Replace minio with rustfs for S3 test server

normanrz:
### Description:
MinIO is being phased out as a comfortable dependency here — its Homebrew core formula is on a deprecation track (disable date 2027-02-17), and the project has pushed users toward AGPLv3 licensing and its own separate tap. This switches the local S3-compatible test server used by the `webknossos` test suite from MinIO to [RustFS](https://github.com/rustfs/rustfs): an Apache-2.0, Rust-based, single-binary S3-compatible server explicitly designed as a MinIO drop-in.

Since the test suite only touches S3 through `s3fs`/`UPath(..., key=, secret=, endpoint_url=)`, this is a self-contained infra change — no test logic changes, only the fixture that spins the server up.

**Bonus over the old MinIO setup:** rather than requiring a manual per-OS install step (`brew install minio` / downloading `minio.exe`) or Docker, `tests/constants.py` now auto-downloads and caches the matching RustFS binary for the current OS/arch on first use (via the GitHub releases API), the same pattern `test.py` already uses for test-data zips. This collapses the old three-way `sys.platform` branching (native macOS / native Windows / Docker on Linux) into a single code path for all platforms, and drops the Docker dependency for this fixture entirely.

### Changes
- `webknossos/tests/constants.py`: `use_minio()` → `use_rustfs()`, with a new `_ensure_rustfs_binary()` that downloads/caches the right release asset into `webknossos/.rustfs_bin` (gitignored). `MINIO_*` constants renamed to `S3_*`.
- `webknossos/tests/test_cli.py`, `webknossos/tests/dataset/test_dataset.py`: updated imports/fixture names accordingly.
- `.github/workflows/ci.yml`: removed the Windows "Install minio" step (no longer needed); added `actions/cache` for `.rustfs_bin` on both Linux and Windows jobs to avoid re-downloading and to reduce the risk of hitting the unauthenticated GitHub API rate limit across the CI matrix.
- `CONTRIBUTING.md`: updated setup instructions. Noted caveat: RustFS has no prebuilt binary for Intel macOS (Apple Silicon only) — Intel Mac devs need to build from source or run RustFS via Docker manually.
- `.gitignore`: swapped the stale `minio_data` entry for `.rustfs_bin/` and `testoutput_rustfs`.

### Verification
Verified locally on macOS/arm64:
- Fresh download → server startup → S3 round-trip via `test_cli.py::test_tiff_cubing_zarr_s3` passes.
- Cached-binary fast path (no re-download) confirmed on a second run.
- Broader S3 coverage: 88/91 `output_path1`/`remote`/`with_symlink`-parametrized tests in `test_dataset.py` pass; the 3 unrelated failures hit a real WEBKNOSSOS server API (`localhost:9000`) that wasn't spun up for this targeted check — not a RustFS regression.
- `./format.sh check`, `./lint.sh`, `./typecheck.sh` all pass.

Not yet verified: the Linux and Windows CI paths (this was tested from macOS) — first CI run on this PR should confirm the cold-cache download and `actions/cache` behavior there.

### Issues:
- N/A

### Todos:
- [ ] Updated Changelog — skipped; this is test-infrastructure only, no user-facing library change
- [ ] Updated Documentation — skipped; only CONTRIBUTING.md dev-setup notes, updated above
- [x] Added / Updated Tests — existing S3 tests retargeted to RustFS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -235,12 +235,6 @@ jobs:
           enable-cache: true
           cache-dependency-glob: "webknossos/uv.lock"
 
-      - name: Cache RustFS binary
-        uses: actions/cache@v4
-        with:
-          path: webknossos/.rustfs_bin
-          key: rustfs-bin-${{ runner.os }}
-
       - name: Set up Python ${{ matrix.python-version }}
         run: uv python install ${{ matrix.python-version }}
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -154,6 +154,12 @@ jobs:
           enable-cache: true
           cache-dependency-glob: "webknossos/uv.lock"
 
+      - name: Cache RustFS binary
+        uses: actions/cache@v4
+        with:
+          path: webknossos/.rustfs_bin
+          key: rustfs-bin-${{ runner.os }}
+
       - name: Set up Python ${{ matrix.python-version }}
         run: uv python install ${{ matrix.python-version }}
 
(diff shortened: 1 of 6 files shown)


normanrz:
Closing in favor of #1506. `moto` seems a better choice, because it can be installed via uv directly and offers similar performance like minio or rustfs (with a proper wsgi server). Also, rustfs caused the Windows CI to hang.

PR #1512, opened 2026-08-18
Add Zarr/OME-Zarr/N5/Neuroglancer-precomputed converters

normanrz:
### Description:
Adds five new read-only input formats to the `ChunkedImageSource` framework introduced for `.ims` and MRC (#1495):

- plain Zarr arrays (v2 `.zarray` and v3 `zarr.json` array nodes)
- OME-Zarr multiscale groups (NGFF 0.4 on Zarr v2, NGFF 0.5 on Zarr v3)
- zipped OME-Zarr archives (`.ozx`, [NGFF RFC-9](https://ngff.openmicroscopy.org/rfc/9/)) — a single ZIP file whose root is a Zarr v3 OME-Zarr multiscale group, read without ever extracting it
- plain N5 datasets and the common N5 multiscale pyramid convention (`s0`, `s1`, …)
- Neuroglancer precomputed volumes (`info` + per-scale chunked data)

For any multiscale source, the finest resolution level is picked and written as `Mag(1)`; WEBKNOSSOS's existing downsampling (`Layer.downsample()` / `write_layer(..., downsample=True)`) builds the rest of the mag pyramid afterward, same as every other converter — no changes to that machinery. These readers go through tensorstore, which is already a base (non-optional) dependency, so **no new extra is required** for any of them, `.ozx` included. Only local input paths are supported for now.

When an OME-Zarr file (0.4, 0.5 or `.ozx`) carries `omero` channel metadata, it's now parsed and applied during conversion: each converted layer's `default_view_configuration` gets the channel's color, intensity range, min/max, and whether it starts disabled, and — when channels are split into one layer each — the layer is named from the channel's label (e.g. `dataset__DAPI`) instead of `dataset__channel4`, whenever one is available.

Implementation notes for reviewers:
- A shared `TensorStoreChunkedImageSource` base (`tensorstore_chunked_image_source.py`) reads a box generically once a subclass resolves a tensorstore open spec and an axis-role mapping — labeled axes when available (Zarr v3 `dimension_names`, OME `axes` metadata), an OME-NGFF-style positional fallback otherwise.
- `ChunkedImageSource` gained an optional `probe_directory()` hook so bare/unsuffixed store directories and neuroglancer-precomputed volumes (which have no suffix convention at all) are recognized by content rather than extension.
- `from_images()`'s directory walk now treats a recognized chunked-source directory as a single input entry instead of descending into its internal chunk files (previously those files were silently skipped, since none carry a recognized extension).
- Multiscale resolution selection reuses the existing `get_possible_layers()`/`allow_multiple_layers` mechanism (the same one `czi_channel` uses) via a new `"scale"` format option — no new public API parameters needed.
- `OzxImageSource` (`ozx_image_source.py`) reads a `.ozx` archive's root `zarr.json` via stdlib `zipfile` to resolve the OME-Zarr multiscale group, then opens the chosen dataset as a `zarr3` array through tensorstore's `zip` kvstore driver, which wraps the archive's own (`file`/`s3`/`gs`/`http(s)`) kvstore — nothing is ever extracted to disk. The OME-Zarr multiscale ranking/axis-naming logic is shared with `ZarrImageSource` via the new `ome_zarr_multiscale.py`, since both read the identical `multiscales` metadata and only differ in how the resolved dataset path becomes something openable.
- `omero.channels` parsing lives in `ome_zarr_multiscale.py` too, built into `LayerViewConfiguration` up front so both readers just look up the channel they're writing. `ImageSource` gained two hooks with no-op defaults — `suggested_view_configuration` and `layer_split_label` — so every other format (CZI, TIFF, N5, Neuroglancer-precomputed, `.ims`, MRC) is unaffected; `add_layer_from_images` merges the suggested color with the existing red/green/blue/hue-rotation fallback rather than replacing it outright, so a channel with no usable color still gets one.

### Issues:
- builds on #1495

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -117,23 +117,47 @@ def open_slice_reader(path_spec: str, **kwargs: object) -> SliceReader:
     )
 
 
+def _find_chunked_reader_class(path: UPath) -> type[ChunkedImageSource] | None:
```
normanrz:
```suggestion
def _find_chunked_image_source_class(path: UPath) -> type[ChunkedImageSource] | None:
```
Please do not use the term "chunked reader" anywhere. It is a "chunked image source".

webknossos/webknossos/dataset/_image_conversion/n5_image_source.py
```diff
@@ -0,0 +1,184 @@
...
+        self._possible_layers: dict[str, list[int]] = {}
+
+        resolved_path = self._resolve_dataset_path(path, options)
+
+        try:
+            self._ts_spec = {"driver": "n5", "kvstore": _make_kvstore(resolved_path)}
+            array = ts.open(
+                self._ts_spec, open=True, context=TS_CONTEXT, recheck_cached="open"
+            ).result()
+        except CorruptImageError:
```
normanrz: Where would that be raised from?

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
```diff
@@ -0,0 +1,153 @@
...
+
+import numpy as np
+import tensorstore as ts
+from upath import UPath
+
+from .._utils.tensorstore_helpers import TS_CONTEXT
+from ..errors import UnsupportedImageDataError
+from .chunked_image_source import ChunkedImageSource
+
+# Axis role names, in the canonical order a `_read_source_box` result uses
```
normanrz: Why are these called "axis roles". I think they should just be called "axes" like in the NDBoundingBox.

webknossos/webknossos/dataset/_image_conversion/zarr_image_source.py
```diff
@@ -0,0 +1,261 @@
...
+            )
+            return self._resolve_multiscale(path, attributes, options)
+
+        raise CorruptImageError(
+            f"{path} is not a valid Zarr store (no {ZARR_JSON_FILE_NAME}/"
+            f"{ZARRAY_FILE_NAME}/{ZGROUP_FILE_NAME} found).",
+            path=path,
+        )
+
+    def _resolve_multiscale(
```
normanrz:
```suggestion
    def _resolve_ome_multiscale(
```

webknossos/webknossos/dataset/_image_conversion/zarr_image_source.py
```diff
@@ -0,0 +1,261 @@
...
+        self, path: UPath, attributes: dict[str, Any], options: ReadOptions
+    ) -> tuple[UPath, list[str] | None]:
+        multiscales = attributes.get("multiscales")
```
normanrz: Check that the OME-Zarr "version" is either 0.4 or 0.5, otherwise raise unsupported with what the version says it is.

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
normanrz: Renamed `_find_chunked_reader_class` to `_find_chunked_image_source_class`, and fixed the "chunked reader" phrasing in the docstrings/comments this PR introduces to say "chunked image source" instead.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_image_conversion/n5_image_source.py
normanrz: Nothing in that try block actually raises `CorruptImageError` (`_make_kvstore` builds a dict, `ts.open()` raises tensorstore's own errors) — it was dead defensive code copied from a pattern that didn't apply here. Removed the `except CorruptImageError: raise` clause (and the analogous one in zarr_image_source.py).

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
normanrz: Renamed to just "axes" throughout, matching `NDBoundingBox.axes`: `guess_axis_roles`->`guess_axes`, `normalize_axis_role`->`normalize_axis`, `_axis_roles`->`_axes`, and the local variables/constants that spelled out "role".

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_image_conversion/zarr_image_source.py
normanrz: Renamed to `_resolve_ome_multiscale`.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/webknossos/dataset/_image_conversion/zarr_image_source.py
normanrz: Added a check: the version is read from wherever each NGFF revision stores it (inside the multiscale entry for 0.4, on the enclosing "ome" object for 0.5) and `UnsupportedImageFormatError` is raised naming the actual version when it isn't "0.4" or "0.5". Added a test covering both Zarr versions.

_🤖 Addressed by [Claude Code](https://claude.com/claude-code)_

webknossos/tests/utils.py
```diff
@@ -32,40 +30,3 @@ def TestTemporaryDirectoryNonLocal() -> Generator[UPath, None, None]:
     temp_dir = UPath(f"memory:///{random_prefix}")
     temp_dir.mkdir(parents=True, exist_ok=True)
     yield temp_dir
-
-
-def create_synthetic_czi(
```
normanrz, line 37: Moved to data_fixtures.py

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -48,7 +48,7 @@ def get_valid_slice_reader_extensions() -> set[str]:
     return valid_extensions
 
 
-def get_valid_chunked_reader_extensions() -> set[str]:
+def get_valid_chunked_image_source_extensions() -> set[str]:
```
normanrz, line 51: Naming consistency.

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
```diff
@@ -0,0 +1,131 @@
+"""Shared OME-Zarr multiscale group parsing: ranking a group's `datasets` by
```
normanrz, line 1: Extracted from zarr_image_source.py

webknossos/tests/data_fixtures.py
```diff
@@ -99,3 +139,226 @@ def create_synthetic_multi_timepoint_ims(
...
                 data = np.full((z, y, x), t * 100 + c, dtype=dtype)
                 ch.create_dataset("Data", data=data)
+
+
+def write_zarr_v2_array(path: UPath, data: np.ndarray) -> None:
+    """Writes `data` as a plain Zarr v2 array (a single chunk, no compression)
+    via tensorstore, for tests of readers that consume a bare `.zarray`."""
+    array = ts.open(
+        {
+            "driver": "zarr",
```
Tobias314: nit: use "zarr2" to make it more explicit?

webknossos/tests/dataset/test_n5_image_source.py
```diff
@@ -0,0 +1,100 @@
...
+from upath import UPath
+
+from tests.data_fixtures import write_n5_array
+from webknossos.dataset import CorruptImageError
+from webknossos.dataset._image_conversion.image_source import ReadOptions
+from webknossos.dataset._image_conversion.n5_image_source import N5ImageSource
+
+
+def _open(path: UPath, **options: Any) -> N5ImageSource:
+    return N5ImageSource(path, ReadOptions(**options))
```
Tobias314, line 14: why do we need this function, couldn't we just write this one line directly?

Tobias314: Ok, seems like this _open function is a patter in the tests. Still not sure whether this adds that much but also not a problem, I guess :sweat_smile:

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
```diff
@@ -49,6 +49,12 @@ def __init__(self, path: UPath, options: ReadOptions) -> None:
     def supported_file_extensions(cls) -> set[str]:
         """File extensions (without the dot, lowercase) this class can read."""
 
+    @classmethod
+    def probe_directory(cls, _path: UPath) -> bool:
+        """Whether this class can open `path` as a directory-based store, used
+        when no registered extension matches. Default: no directory sniffing."""
```
Tobias314, line 55: I understand that this can be used to check whether a specific image source class can read a given directory. But I don't get what this has to do with matching extensions. Is this relevant here?

webknossos/tests/dataset/test_ozx_image_source.py
```diff
@@ -0,0 +1,226 @@
...
+    np.testing.assert_array_equal(block[0], coarse)
+
+
+def test_single_resolution_ozx_group_reports_no_scale_choice(tmp_upath: UPath) -> None:
+    ozx_path = tmp_upath / "single_scale.ozx"
+    data = np.ones((4, 8, 8), dtype=np.uint8)
+    write_ozx_file(ozx_path, [("0", data, [1.0, 1.0, 1.0])], _AXES_CZYX[1:])
+
+    source = _open(ozx_path)
+    assert source.get_possible_layers() is None
```
Tobias314: Probably, there is a reason for this but I do not get why we have no possible layers. We have a file with data so I would expect to get back one layer?

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -193,6 +194,22 @@ def _to_callable(
             raise ValueError(f"Got unexpected ConversionLayerMapping value: {self}")
 
 
+def _iter_convertible_paths(root: UPath, valid_extensions: set[str]) -> Iterator[UPath]:
```
Tobias314, line 197: should the root dir itself get checked whether it already is a valid data source?
I think the old implementation also didn't do this so might be intended?

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -126,6 +127,20 @@ def get_possible_layers(self) -> dict[str, list[int]] | None:
...
+        the channel this source writes, when the format's own metadata
+        carries one. None for formats with no such metadata."""
+        return None
+
+    def layer_split_label(self, key: str, value: int) -> str:
+        """A layer-name suffix component for one `get_possible_layers()`
+        split entry. Defaults to `f"{key}{value}"`; a format may override
+        this with something more descriptive, e.g. an OME-Zarr channel's
+        `omero` label."""
+        return f"{key}{value}"
```
Tobias314, line 140: If key, value come from self.get_possible_layers() why is value type int and not list[int]?

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -117,23 +120,47 @@ def open_slice_reader(path_spec: str, **kwargs: object) -> SliceReader:
     )
 
 
+def _find_chunked_image_source_class(path: UPath) -> type[ChunkedImageSource] | None:
+    """The registered chunked image source that would open `path`, by
+    extension or, for a directory that no extension matches, by
+    `probe_directory`. None if none claims it."""
```
Tobias314: nit: could maybe be removed so that the docstring only describes the what, callers do not care that much about the how

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -117,23 +120,47 @@ def open_slice_reader(path_spec: str, **kwargs: object) -> SliceReader:
...
+    if path.is_dir():
+        for cls in _CHUNKED_IMAGE_SOURCE_CLASSES:
+            if cls.probe_directory(path):
+                return cls
+    return None
+
+
+def is_chunked_source_directory(path: UPath) -> bool:
+    """Whether `path` is a directory a registered chunked image source
+    recognizes as one store (by extension or by `probe_directory`) — so a
```
Tobias314: again, maybe leave this out. This is an implementation detail and one needs to also look into the implementation of _find_chunked_image_source_class to understand this instead of just seeing this function as an independent unit.

webknossos/tests/dataset/test_ozx_image_source.py
Tobias314: Ok, this returns whether the data of one source should be split into multiple layers, right? If its None the source becomes one layer. We do not want to split here, so all fine. A different name for the function might be better but I guess we cannot change this because of backwards compatibility?

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
```diff
@@ -0,0 +1,252 @@
...
+        if transform.get("type") == "scale":
+            return transform["scale"]
+    raise CorruptImageError(
+        f"OME-Zarr dataset {dataset.get('path')!r} of {path} has no 'scale' "
+        "coordinateTransformation.",
+        path=path,
+    )
+
+
+def _resolution_key(
```
Tobias314, line 38: since I have seen this twice: there might be edge cases where the fines resolution is not unique: e.g. option 1 having 3,3,1 nm per voxel and option 2 having 2,2,2 nm per voxel. Option 1 is finer in z but option 2 finer in xy.

Not that important since this should not happen but might be worth throwing an error or warning so this does not go unnoticed.

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
```diff
@@ -0,0 +1,164 @@
...
+    if ndim == 5:
+        return ("t", "c", "z", "y", "x")
+    raise UnsupportedImageDataError(
+        f"Cannot place the {ndim} axes of {path} — only 2 to 5 dimensional "
+        "arrays are supported.",
+        path=path,
+    )
+
+
+class TensorStoreChunkedImageSource(ChunkedImageSource):
```
Tobias314, line 82: Maybe put "local" into the name to make clear that we cannot deal with remote paths? This fact is relied on by subclasses and other code parts and it might not be obvious since they all take UPaths.

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
```diff
@@ -0,0 +1,164 @@
...
+    multiscale sources) and `self._axes` (the physical dimension order, from
+    `t`/`c`/`z`/`y`/`x`) in `__init__`, together with the usual
+    `ChunkedImageSource` fields via `compute_channel_selection()`.
+    """
+
+    _ts_spec: dict[str, Any]
+    _axes: tuple[str, ...]
+    # Set by subclasses via compute_channel_selection(), same as
+    # CziImageSource/ImsImageSource.
+    _first_n_channels: int | None
```
Tobias314: referencing those other classes makes it harder to read.
Also, if every subclass does this, can we maybe pull the compute_channel_selection() call into the base?

Tobias314, line 95: Does our type checker enforce those attributes getting overridden?

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
Tobias314: maybe add "helpers" or "parsing" to the name to make clear that those are pulled out methods used by the zarr image sources.

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
Tobias314, line 1: one could add unit tests for the public functions of this file

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
```diff
@@ -0,0 +1,252 @@
...
+    return channels[channel_index]
+
+
+def suggested_view_configuration(
+    channels: tuple[OmeChannelMetadata, ...] | None,
+    channel_index: int | None,
+    num_channels: int,
+) -> LayerViewConfiguration | None:
+    """The `omero`-derived view configuration for the channel this source
+    writes: `channel_index` if pinned, else channel 0 when there is only a
```
Tobias314: "this source" is not true anymore, I guess, because this is no longer part of an ImageSource class.

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
```diff
@@ -0,0 +1,252 @@
...
+    channel = _omero_channel_at(channels, channel_index, num_channels)
+    return channel.view_configuration if channel else None
+
+
+def layer_split_label(
+    channels: tuple[OmeChannelMetadata, ...] | None, key: str, value: int
+) -> str | None:
+    """A layer-name suffix component for one `get_possible_layers()` split
+    entry, from the channel's `omero` label. Only resolves anything for
+    `key == "channel"`; None when there's no usable label, so the caller
```
Tobias314, line 177: I don't really get this key=="channel". This "channel" seems to be a magic value and I do not fully understand yet what its semantic is.

normanrz:
Heads up: this PR's new \`TensorStoreChunkedImageSource\` readers (Zarr, OME-Zarr, \`.ozx\`, N5, neuroglancer-precomputed) inherit \`ChunkedImageSource.expected_bbox\` unchanged, and \`guess_axes\`'s own 5D fallback treats \`t,c,z,y,x\` as the canonical layout — so converting a multi-channel, multi-timepoint array into a single (non-split) layer is a very reachable case here, not an edge case.

As things stand on \`release-4.0\` today, that combination hits \`assert num_channels == 1\` inside \`NDBoundingBox.normalize_axes()\`, because \`ChunkedImageSource.expected_bbox\`'s t-axis branch only ever omitted \`"c"\` — it was never given a channel count above 1 to carry. \`test_zarr_image_source.py\` doesn't currently cover \`num_channels > 1\`, so this wouldn't be caught by the existing tests.

#1515 fixes this at the shared \`ChunkedImageSource\` base (unrelated to that PR's own scope — it's a typing cleanup that surfaced the bug along the way), so every reader here inherits the fix automatically once it's in. I test-merged #1515 onto this branch and confirmed end-to-end with a synthetic Zarr v3 array (\`dimension_names=["t","c","z","y","x"]\`, 3 channels × 2 timepoints): crashes without #1515, converts correctly with it. No merge conflicts between the two PRs either way round — worth landing #1515 first (or after, doesn't matter for conflicts, just for correctness of this PR's own multi-channel+multi-timepoint conversions).

webknossos/webknossos/dataset/_image_conversion/ozx_image_source.py
```diff
@@ -0,0 +1,169 @@
...
+                "requires a zipped OME-Zarr archive's root to be an "
+                "OME-Zarr multiscale group.",
+                path=path,
+            )
+        attributes = root_metadata.get("attributes", {})
+        ome = attributes.get("ome", attributes)
+        multiscale = resolve_ome_multiscale(ome, path=path)
+        self._omero_channels = multiscale.channels
+
+        rank = options.format_option("scale")
```
Tobias314, line 97: is the "scale" also a magic value? I see it in different locations (also with the possible_layers) while I don't fully understand its semantic.

Tobias314 (approved):
I have taken a look. Tests look good.
I have also added some comments (mainly questions)
Since I do not know the details of the specs, especially the zarr spec, I could not go into details there without spending time with the specs. I think most is just plain parsing code. However, I find it a bit complicated how different zarr groups and channels are turned into WK layers. The `possible_layers' and `get_possible_layers` variables/functions seem to play a role here but I do not find it straight forward how they encode things. It feels like this was meant for single images which just have one array with multiple channels and is now put onto all the possibilities that zarr arrays provide.

webknossos/tests/data_fixtures.py
normanrz:
```suggestion
            "driver": "zarr2",
```

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
normanrz: In image_source_registry.py, first the extension of a path is examined. If no matching ImageSource is found, the probe_directory mechanism is being used.

webknossos/tests/dataset/test_ozx_image_source.py
normanrz: Fixed in https://github.com/scalableminds/webknossos-libs/pull/1512/changes/be696efb1cbc99f08ddae769cebc90d75825665e

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -126,6 +127,20 @@ def get_possible_layers(self) -> dict[str, list[int]] | None:
...
+        """Format-specific default display (color, intensity range, ...) for
+        the channel this source writes, when the format's own metadata
+        carries one. None for formats with no such metadata."""
+        return None
+
+    def layer_split_label(self, key: str, value: int) -> str:
+        """A layer-name suffix component for one `get_possible_layers()`
+        split entry. Defaults to `f"{key}{value}"`; a format may override
+        this with something more descriptive, e.g. an OME-Zarr channel's
+        `omero` label."""
```
```suggestion
        split entry."""
```

webknossos/webknossos/dataset/_image_conversion/image_source.py
normanrz: This function is called for one entry of `self.get_possible_layers()`

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -117,23 +120,47 @@ def open_slice_reader(path_spec: str, **kwargs: object) -> SliceReader:
...
+    caller walking a directory tree should treat it as a single leaf entry
+    rather than descend into it."""
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
normanrz: Fixed the calling code in https://github.com/scalableminds/webknossos-libs/pull/1512/changes/47f708afe27bd6ef5d5cd567e1327d2b0bf4fe1b

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
normanrz: Fixed in https://github.com/scalableminds/webknossos-libs/pull/1512/changes/fa48f6379a9e99bba8d079128acb767002b24260

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
normanrz:
```suggestion
    """The `omero`-derived view configuration for the channel:
    `channel_index` if pinned, else channel 0 when there is only a
```

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
normanrz: Yeah, this is a bit weird. There are multiple "reasons" (that correspond to axes) why a layer can get split. Multi-channel is one of them, which is being handled here. I'll work on that abstraction.

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
```diff
@@ -0,0 +1,164 @@
...
+    tensorstore open spec, already pointing at the chosen resolution level for
+    multiscale sources) and `self._axes` (the physical dimension order, from
+    `t`/`c`/`z`/`y`/`x`) in `__init__`, together with the usual
+    `ChunkedImageSource` fields via `compute_channel_selection()`.
+    """
+
+    _ts_spec: dict[str, Any]
+    _axes: tuple[str, ...]
+    # Set by subclasses via compute_channel_selection(), same as
+    # CziImageSource/ImsImageSource.
```
normanrz:
```suggestion
    # Set by subclasses via compute_channel_selection()
```

webknossos/webknossos/dataset/_image_conversion/ozx_image_source.py
normanrz: The user can specify which scale to pick. I removed it from possible_layers (now get_layer_split_options).

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
normanrz: No — checked this empirically rather than from memory. I temporarily removed the `self._first_n_channels = ...` assignment from `N5ImageSource.__init__` (renamed the unpacked tuple element to an unused local instead) and ran `mypy` on the module; it reported **"Success: no issues found"** even though the attribute is never set on that subclass. Restored the assignment afterward — `mypy -p webknossos.dataset._image_conversion` is clean again.

`_ts_spec`, `_axes` and `_first_n_channels` here are bare PEP 526 annotations (`name: Type`, no `= value`). That only declares the attribute's *type* for the checker's benefit — it creates no class attribute, and mypy doesn't verify that every subclass's `__init__` actually assigns it. It just trusts the annotation wherever the attribute is used in the base class's own methods, with no cross-subclass initialization tracking. Our mypy config (`disallow_untyped_defs`, `no_implicit_optional`, `strict_equality`, ...) doesn't add any check for this either — mypy has no flag for "prove every declared attribute is initialized," and we don't run Pyright (which does have `reportUninitializedInstanceVariable`, off by default anyway).

So today this is only enforced by convention plus a runtime `AttributeError` the first time something touches an unset attribute — not the type checker. All four current subclasses (Zarr, N5, Ozx, NeuroglancerPrecomputed) do set them correctly. Happy to harden this (e.g. abstract properties, or a `__init_subclass__` check) if you'd like, but wanted to flag that nothing currently catches a future subclass forgetting to.

*(Written by Claude, on Norman's behalf.)*

webknossos/webknossos/dataset/_image_conversion/ome_zarr_helpers.py
normanrz: Added a dedicated test file: [`test_ome_zarr_helpers.py`](https://github.com/scalableminds/webknossos-libs/blob/98d58c14/webknossos/tests/dataset/test_ome_zarr_helpers.py) (24 tests), covering `resolve_ome_multiscale`, `suggested_view_configuration` and `layer_split_label` directly against plain metadata dicts rather than only indirectly through `ZarrImageSource`/`OzxImageSource`:

- resolution ranking (out-of-order `datasets`, ignoring the channel axis' own scale factor, the axes-less Zarr v2 fallback)
- version validation (0.4 vs 0.5, where each carries `version`, unsupported version rejected)
- error paths (missing `multiscales`, empty `datasets`, missing `scale` transform, unsupported axis name)
- `omero.channels` parsing wired through into `OmeMultiscale.channels`
- `suggested_view_configuration`'s pinned/single-channel/ambiguous/out-of-range/no-metadata cases
- `layer_split_label`'s key-gating, sanitization and empty-after-sanitization fallback

In `8599b89b..98d58c14`.

*(Written by Claude, on Norman's behalf.)*

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -117,23 +120,40 @@ def open_slice_reader(path_spec: str, **kwargs: object) -> SliceReader:
...
-        for cls in _CHUNKED_IMAGE_SOURCE_CLASSES:
-            if extension in cls.supported_file_extensions():
-                # Remote UPaths deliberately do reach the reader, which raises
-                # its own "must be a local file path" error via is_remote_path.
-                return cls(images, options)
+        cls = _find_chunked_image_source_class(images)
+        if cls is not None:
+            # Remote UPaths deliberately do reach the image source, which
+            # raises its own "must be a local file path" error via
+            # is_remote_path, for sources that need one.
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/image_source_registry.py
```diff
@@ -204,10 +224,16 @@ def _import_readers() -> str | None:
     import_exceptions = []
 
     # No optional dependency beyond imageio, which is a hard requirement — but
-    # imported here so that every reader registers in one place.
+    # imported here so that every reader registers in one place. The
+    # tensorstore-backed readers need no optional dependency either:
+    # tensorstore is a base (non-optional) dependency of this package.
```
normanrz:
```suggestion
    # imported here so that every reader registers in one place.
```

webknossos/webknossos/dataset/_image_conversion/zarr_image_source.py
```diff
@@ -0,0 +1,196 @@
...
+        super().__init__(path, options)
+        self._possible_layers: dict[str, list[int]] = {}
+
+        resolved_path, axis_name_hint, self._omero_channels = self._resolve_array_path(
+            path, options
+        )
+
+        try:
+            driver = (
+                "zarr3" if (resolved_path / ZARR_JSON_FILE_NAME).is_file() else "zarr"
```
normanrz:
```suggestion
                "zarr3" if (resolved_path / ZARR_JSON_FILE_NAME).is_file() else "zarr2"
```

webknossos/webknossos/dataset/dataset.py
```diff
@@ -1158,6 +1159,7 @@ def add_layer_from_images(
         * `dtype`: the read image data will be convertoed to this dtype using `numpy.ndarray.astype`
         * `channel`: may be used to select a single channel, if multiple are available
         * `czi_channel`: may be used to select a channel for .czi images, which differs from normal color-channels
+        * `scale`: may be used to select a resolution level for a multiscale source (Zarr/OME-Zarr, `.ozx`, N5, neuroglancer-precomputed), 0 = finest (the default). A multiscale source always contributes a single level, never one layer per level.
```
normanrz:
```suggestion
        * `scale`: may be used to select a resolution level for a multiscale source (Zarr/OME-Zarr, `.ozx`, N5, neuroglancer-precomputed), 0 = finest (the default).
```

webknossos/Changelog.md
```diff
@@ -21,13 +21,15 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
...
 
 ### Added
+- Added support for converting plain Zarr arrays (v2 and v3), OME-Zarr multiscale groups (NGFF 0.4 and 0.5), zipped OME-Zarr archives (`.ozx`, [NGFF RFC-9](https://ngff.openmicroscopy.org/rfc/9/)), plain N5 datasets and N5 multiscale pyramids, and Neuroglancer precomputed volumes via `Dataset.from_images`/`add_layer_from_images`. For a multiscale source, the finest resolution level is converted as `Mag(1)`; the rest of the mag pyramid is built by the usual `downsample=True`/`Layer.downsample()` afterwards, as with every other format. `.ozx` archives are read directly through tensorstore's `zip` kvstore driver, without extracting them to disk. Only local input paths are supported for now. Builds on the `ChunkedImageSource` abstraction introduced for `.ims` and MRC. [#1495](https://github.com/scalableminds/webknossos-libs/pull/1495)
```
```suggestion
- Added support for converting plain Zarr arrays (v2 and v3), OME-Zarr multiscale groups (NGFF 0.4 and 0.5), zipped OME-Zarr archives (`.ozx`, [NGFF RFC-9](https://ngff.openmicroscopy.org/rfc/9/)), plain N5 datasets and N5 multiscale pyramids, and Neuroglancer precomputed volumes via `Dataset.from_images`/`add_layer_from_images`. [#1495](https://github.com/scalableminds/webknossos-libs/pull/1495)
```

PR #1515, opened 2026-08-19
Use NormalizedBoundingBox internally in the image-conversion pipeline

### LINKED ISSUE #1514: Use NormalizedBoundingBox internally in the image-conversion pipeline (ImageSource/ChunkedImageSource/SlicedImageSource)
## Context

Convention (per @normanrz, in review of #1512): inside the codebase, `NormalizedBoundingBox` should be used to remove ambiguity (fixed axis order, explicit channel axis). `BoundingBox`/`NDBoundingBox` should only appear on user-facing interfaces.

That convention is already followed correctly at the storage layer:
- `BaseArray.read`/`write` (`webknossos/dataset/layer/view/_array.py`) take `NormalizedBoundingBox`.
- `View`'s internal read/write plumbing (`webknossos/dataset/layer/view/view.py`) does the same.
- The public boundary, `MagView.write(..., absolute_bounding_box: NDBoundingBox)`, accepts `NDBoundingBox` (correct — it's user-facing) and normalizes it internally before handing off.

The image-conversion pipeline under `webknossos/dataset/_image_conversion/` does not follow it. `ImageSource`/`ChunkedImageSource`/`SlicedImageSource` are fully private (underscore-prefixed package, never exposed to library users — `Dataset.add_layer_from_images`/`Dataset.from_images` are the actual public entry points, and neither takes a bounding-box argument), yet their entire shared interface is typed `NDBoundingBox`/`BoundingBox`:

- `webknossos/dataset/_image_conversion/image_source.py`
  - `ImageSource.expected_bbox -> NDBoundingBox` (line 146)
  - `ImageSource.copy_chunk_to_view(bbox: NDBoundingBox, ...)` (line 153)
  - `ImageSource.initial_layer_bounding_box(mag1_expected_bbox: NDBoundingBox) -> NDBoundingBox` (line 175-176)
  - `ImageSource.chunk_grid(layer_bounding_box: NDBoundingBox, ...) -> list[NDBoundingBox]` (line 183-188)
  - `ImageSource.final_bounding_box(layer_bounding_box: NDBoundingBox, ...) -> NDBoundingBox` (line 195-199)
- `webknossos/dataset/_image_conversion/chunked_image_source.py` — same interface implemented (lines 86-246), plus constructs a plain `BoundingBox` internally at line 103
- `webknossos/dataset/_image_conversion/sliced_image_source.py` — same interface implemented (lines 406-499), plus a `bbox: BoundingBox | NDBoundingBox` parameter at line 302 and a plain `BoundingBox` construction at line 420

This predates #1512 (introduced by #1495/#1498, the `.ims`/MRC `ChunkedImageSource` framework and its later `SlicedImageSource` split) and is shared by every reader (`.ims`, MRC, CZI, TIFF, DM3/DM4, Zarr, N5, neuroglancer precomputed, ...), so it's out of scope for any single reader's PR to fix — it needs to change once, centrally, for the whole `ImageSource` contract.

## Instructions

Retype the shared internal `ImageSource` interface (and its two implementations) to use `NormalizedBoundingBox` instead of `NDBoundingBox`/`BoundingBox`, mirroring how `BaseArray`/`View` already do this at the storage layer:

1. In `image_source.py`, change the abstract methods listed above (`expected_bbox`, `copy_chunk_to_view`, `initial_layer_bounding_box`, `chunk_grid`, `final_bounding_box`) to use `NormalizedBoundingBox` throughout.
2. Update `chunked_image_source.py` and `sliced_image_source.py` accordingly — including the internal `BoundingBox(...)` constructions (line 103 and line 420), which should become `NormalizedBoundingBox(...)` (with an explicit `"c"` axis, removing the ambiguity a plain `BoundingBox` has about whether a channel axis is implied).
3. `sliced_image_source.py`'s `bbox: BoundingBox | NDBoundingBox` parameter (line 302) should collapse to a single `NormalizedBoundingBox` type once its caller(s) are updated.
4. At the actual public boundary — where `image_conversion.py` calls `mag_view.write(..., absolute_bounding_box=...)` and where `layer.bounding_box = ...` is assigned (`image_conversion.py:730`, already correctly using `BoundingBox.from_ndbbox(...)`) — keep converting to `NDBoundingBox`/`BoundingBox`, since those *are* the user-facing/public signatures.
5. Every concrete reader (`ims_image_source.py`, `mrc_image_source.py`, `czi_image_source.py`, `tiff_slices.py`, `dm_slices.py`, `raster_slices.py`, `zarr_image_source.py`, `n5_image_source.py`, `neuroglancer_precomputed_image_source.py`, `ozx_image_source.py`, and the shared `tensorstore_chunked_image_source.py` base) only implements `_read_source_box`/similar and doesn't construct bounding boxes directly, so this should be a contained, mechanical change to the shared base classes plus their call sites in `image_conversion.py` — not a per-reader change.
6. Run the existing `_image_conversion`/`add_layer_from_images`/`from_images` test suite (`tests/dataset/test_add_layer_from_images.py`, `tests/dataset/test_from_images.py`, plus format-specific test files) to confirm behavior is unchanged; this is a typing/internal-representation cleanup, not a behavior change.

normanrz:
## Context

Inside the codebase, `NormalizedBoundingBox` should be used to remove ambiguity (fixed axis order, explicit channel axis). `BoundingBox`/`NDBoundingBox` should only appear on user-facing interfaces. This is already followed at the storage layer (`BaseArray`/`View`) and was applied to the export pipeline in a sibling refactor.

The image-conversion pipeline (`webknossos/webknossos/dataset/_image_conversion/`) didn't follow it: `ImageSource`/`ChunkedImageSource`/`SlicedImageSource` are fully private, yet typed their whole shared interface as `NDBoundingBox`/`BoundingBox`, including two internal `BoundingBox(...)` constructions that silently omitted a `"c"` axis even when the source had real channels.

Fixes #1514.

## Changes

- Using `NormalizedBoundingBox` inside the image conversion pipeline.
- Making sure that converted datasets always have `c`, `x`, `y`, `z` axes, even if the input datasets don't have such axes. Additional axes are prepended.
- Using constants for the axis names (e.g. `X_AXIS`, `CXYZ_AXES`) throughout.

## Verification

- `./format.sh`, `./lint.sh`, `./typecheck.sh` all clean throughout.
- Full `test_add_layer_from_images.py`, `test_from_images.py`, `test_czi_image_source.py`, `test_slice_reader.py`, `tests/geometry/`, plus the N5/Zarr/OZX/neuroglancer-precomputed source tests — all passing.


normanrz:
Note for reviewers: this PR's fix in \`ChunkedImageSource.expected_bbox\` (making the t-axis branch handle \`num_channels > 1\`, not just single-channel) is actually load-bearing for #1512, not just for \`.ims\`/CZI.

#1512's new \`TensorStoreChunkedImageSource\` readers (Zarr, OME-Zarr, \`.ozx\`, N5, neuroglancer-precomputed) inherit \`ChunkedImageSource.expected_bbox\` unchanged, and their own axis-guessing fallback treats \`t,c,z,y,x\` as *the* canonical 5D layout (\`guess_axes\`: \`if ndim == 5: return ("t", "c", "z", "y", "x")\`). Without this fix, converting a multi-channel, multi-timepoint OME-Zarr/Zarr array into a single (non-split) layer would hit the same \`assert num_channels == 1\` crash this PR fixes for \`.ims\`/CZI — and #1512's test suite doesn't currently cover that combination (\`test_zarr_image_source.py\` only exercises \`num_channels == 1\`).

I test-merged #1512 onto this branch and confirmed end-to-end with a synthetic Zarr v3 array (\`dimension_names=["t","c","z","y","x"]\`, 3 channels × 2 timepoints): crashes without this fix, converts correctly with it (correct \`expected_bbox\`/layer axes, correct round-tripped data). No merge conflicts between the two PRs either way round.

webknossos/tests/dataset/test_slice_reader.py
```diff
@@ -325,3 +332,106 @@ def test_image_sequence_reads_zip_archive(tmp_upath: UPath) -> None:
...
+        # not going through the real file-opening machinery.
+        images = _TsyxReader()
+        if hasattr(self, "_bundle_axes"):
+            images.default_coords.update(self._default_coords)
+            images.bundle_axes = self._bundle_axes
+            images.iter_axes = self._iter_axes
+        yield images
+
+
+def test_no_real_z_axis_gets_a_singleton_without_relabeling_a_real_axis(
```
Tobias314: do we want to call this "singleton". Sounds like the desing pattern, which is not mean here, right?

Tobias314:
> Note for reviewers: this PR's fix in `ChunkedImageSource.expected_bbox` (making the t-axis branch handle `num_channels > 1`, not just single-channel) is actually load-bearing for #1512, not just for `.ims`/CZI.

But #1512 is now already merged. If this PR is load-bearing, why did the other PR got merged?

normanrz:
> > Note for reviewers: this PR's fix in `ChunkedImageSource.expected_bbox` (making the t-axis branch handle `num_channels > 1`, not just single-channel) is actually load-bearing for #1512, not just for `.ims`/CZI.
> 
> But #1512 is now already merged. If this PR is load-bearing, why did the other PR got merged?

I think that comment (by Claude) was not helpful. a) It is not load-bearing, but a follow-up cleanup. b) #1512 got merged into the "release-4.0" branch, not "master" yet.

webknossos/tests/dataset/test_add_layer_from_images.py
```diff
@@ -621,6 +621,61 @@ def test_ims_three_uint8_channels_become_rgb_layer_without_allow_multiple_layers
...
 
+def test_ims_rgb_layer_with_multiple_timepoints(
+    tmp_upath: UPath, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    # Three uint8 channels combine into a single RGB layer (see
+    # test_ims_three_uint8_channels_become_rgb_layer_without_allow_multiple_layers
+    # above), and multiple timepoints add a "t" axis (see
+    # test_ims_from_images_multi_timepoint above) — this combination used to
+    # crash, since expected_bbox couldn't carry both "c" and "t" for a
+    # single, non-split layer.
```
Tobias314:
```suggestion
    # An input containing three uint8 channels at multiple timepoints should 
    # create an RGB layer (multiple channels, "c" axis) with an additional time ("t") axis.
```

webknossos/tests/dataset/test_slice_reader.py
```diff
@@ -325,3 +332,106 @@ def test_image_sequence_reads_zip_archive(tmp_upath: UPath) -> None:
 def test_image_sequence_reports_missing_files(tmp_upath: UPath) -> None:
     with pytest.raises(OSError, match="No files were found"):
         MultiImageSliceReader(str(tmp_upath / "nothing_*.png"))
+
+
+# t, s, y, x — two non-"c" iterated axes and no "z" axis at all.
```
Tobias314: what is an "iterated" axes?

webknossos/tests/dataset/test_slice_reader.py
```diff
@@ -325,3 +332,106 @@ def test_image_sequence_reads_zip_archive(tmp_upath: UPath) -> None:
...
+        "test",
+        category="color",
+        dtype=source.dtype,
+        num_channels=1,
+        data_format="zarr3",
+    )
+    layer.bounding_box = source.initial_layer_bounding_box(box)
+    # A small explicit shard/chunk shape avoids padding every write out to
+    # the (1024, 1024, 1024) default, which is both wasteful and pushes the
+    # unrelated read path below to its limits for this tiny test volume.
```
Tobias314:
```suggestion
    # tiny chunk/shard shape to reduce IO in tests
```

webknossos/tests/dataset/test_slice_reader.py
```diff
@@ -325,3 +332,106 @@ def test_image_sequence_reads_zip_archive(tmp_upath: UPath) -> None:
...
+    mag_view = layer.add_mag(1, compress=False, chunk_shape=8, shard_shape=8)
+
+    # Captures what actually gets written, rather than reading it back
+    # through the storage backend — a 6-axis array (s, t, c, z, y, x) is not
+    # a combination the tensorstore-backed zarr3 reader supports, which is a
+    # storage-layer limitation unrelated to what this test verifies.
```
Tobias314:
```suggestion
    # Capture the output without going through any actual storage layer or doing file IO
```

webknossos/tests/dataset/test_slice_reader.py
Tobias314: Basically, we always want to use x,y,z as spatial axis. If one of them does not exist, add it with size 1. Every other axis (c, t, s, ...) should be splitted/iterated to produce separate outputs/layers.
Does this described the expected behavior and could maybe even used as comment?

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
```diff
@@ -83,37 +88,30 @@ def _read_source_box(
...
         The exact bounding box of the data, in the source's native Mag(1)
         space — never a placeholder, since these formats know their extents.
 
-        Channels are never reported as a "c" axis here: like every other
-        layer bounding box, this describes only the spatial (and, for
-        unpinned multi-timepoint data, "t") extent — the channel count is
-        conveyed separately, via the `num_channels` passed to `add_layer()`,
-        whether the channels end up combined into one layer or split into
-        one layer each.
+        Always carries an explicit "c" axis sized `num_channels`.
```
Tobias314: If "always has c" is part of the NormalizedBoundingBox definition, this should be part of the NormalizedBoundingBox docstring but not re-stated here.

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -91,6 +93,37 @@ def compute_channel_selection(
...
+
+    Unlike `NDBoundingBox.normalize_axes()` (which keeps a missing "c" axis
+    implicit, matching xyz-only boxes elsewhere, e.g. remote datasets),
+    image-conversion's own internal boxes are always meant to carry one, per
+    the codebase's `NormalizedBoundingBox` convention. A missing "c" axis is
+    inserted right after "t" if present, otherwise first.
+    """
+    if "c" in box.axes:
+        return box.normalize_axes(num_channels)
+    insert_at = box.axes.index("t") + 1 if "t" in box.axes else 0
```
Tobias314: Having "t" here kinda breaks bbox generalization since we no longer treat all bbox axes equally. We already treat "c" special. I feel, we are also distinguishing between spatial (x,y,z and "other" axes like t, s). Treating "t" as another special axis is possible, I guess, but might add even more complexity and potential edge cases. I would avoid this as much as possible.

Also, if kept, it could become a constant instead of having "t" as magic value.

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -141,14 +174,20 @@ def layer_split_label(self, key: str, value: int) -> str:
...
-    def expected_bbox(self) -> NDBoundingBox:
+    def expected_bbox(self) -> NormalizedBoundingBox:
         """The bounding box the data is expected to occupy, in Mag(1). Exact,
-        or an oversized placeholder."""
+        or an oversized placeholder.
+
+        Carries a "c" axis whenever there is more than one channel. A
+        single-channel source may still omit it, conveying that count
+        separately via `num_channels`, the same way every other ND layer
+        bounding box does.
```
Tobias314: That is part of the NormalizedBoundingBox type/contract. So I would add the docstring there instead of here.

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -334,16 +335,24 @@ def copy_chunk_to_view(
...
-                    # The sequence is a flat run over every iter axis, so narrow
-                    # it to the stretch this chunk's non-z coordinates select
-                    # before indexing z within it.
+                    # The sequence is a flat run over every iter axis. When the
+                    # last one is genuinely "z", it is batched (multiple
+                    # values per chunk) and narrowed to below; every other
+                    # axis is exactly one value per chunk (see the assert
+                    # above) and only selects where in the flat run this
+                    # chunk starts. Without a real "z" — the box's "z" is then
+                    # a singleton placeholder, not tied to any iterated axis —
```
Tobias314, line 345: do we want to call this singleton?

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -334,16 +335,24 @@ def copy_chunk_to_view(
...
+                    # every iter axis, including the last one, is one of those
+                    # single-valued selectors instead.
+                    has_real_z = "z" in self._iter_axes
+                    outer_axes = self._iter_axes[:-1] if has_real_z else self._iter_axes
```
Tobias314: does this assume that "z" is always the last item of _iter_axes? If that's the case, this seems brittle. Maybe add at least an assertion?

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -417,12 +428,22 @@ def expected_bbox(self) -> NDBoundingBox:
...
+                    self.num_channels
+                )
 
             # Several axes are stepped through (e.g. "t" and "z"), so each one
             # has to be named in the box.
             axes_names = self._iter_axes + self._bundle_axes
             axes_sizes = [sizes[axis] for axis in axes_names]
+            if "z" not in axes_names:
+                # No axis is genuinely called "z" (e.g. only "t" and "s" are
+                # stepped through). A singleton "z" is added instead of
```
Tobias314: singleton

Also, remove everything after "instead". I think explaining what we don't do does not add much here.

Tobias314 (commented):
LGTM, see my comments.

You use "singleton" for an axis with size 1. This might get confused with the singleton pattern. It might be fine if we agree on using it this way, though. If not, rename it everywhere, I have commented its occurrences.

We often depend on special axes names (x,y,z,t,c). Currently, we always use magic value strings. It might be best to not have this at all and treat all axes equally so that the name does not matter. If that's not possible, we should have constants instead of strings. I have commented it once but I think that holds in a lot of places and might need a separate refactoring. Still, we could try to not add new occurrences of this.

normanrz:
> You use "singleton" for axis with size 1. Might get confused with the singleton pattern. Might be fine if we agree on also using it this way. If not, rename it everywhere, I have commented its occurrences.

"Singleton axis" is a commonly-used term for an axis that just has a size of 1. It is also referenced in the numpy docs: https://numpy.org/doc/stable/reference/generated/numpy.expand_dims.html I think it is a nice term and would like to keep it. Also, it is only used in comments and tests.

webknossos/tests/dataset/test_slice_reader.py
normanrz: We always want c,x,y,z. If one is missing add it as a singleton axis. I'll change the name and comment.

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/image_source.py
normanrz: Yeah. I agree we don't need to treat `t` specially here. I still added a `T_AXIS` constant, because `t` is common in tiffs and OME-Zarr.

webknossos/webknossos/dataset/_image_conversion/image_source.py
```suggestion
        or an oversized placeholder.
```

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -403,59 +413,67 @@ def channel(self) -> int | None:
...
             axes_sizes = [sizes[axis] for axis in axes_names]
-            axes_sizes[axes_names.index("x")] = x_size
-            axes_sizes[axes_names.index("y")] = y_size
-            if "c" in axes_names:
-                # sizes["c"] is the source's raw channel count, but only
+            if "z" not in axes_names:
+                # No axis is genuinely called "z" (e.g. only "t" and "s" are
+                # stepped through). A singleton "z" is added instead of
+                # relabeling a real axis, which would corrupt its meaning —
+                # a real "t" axis reported as depth, say.
```
normanrz:
```suggestion
                # stepped through). A singleton "z" is added.
```

webknossos/tests/dataset/test_slice_reader.py
normanrz: Renamed the test to test_no_real_z_axis_gets_a_size_one_axis_without_relabeling_a_real_axis and reworded the comment along the lines you suggested (explicit x/y/z(+c) contract, everything else iterated/split). Also fixed a couple of nearby comments that had been mangled by an earlier find/replace ([webknossos.dataset](webknossos.dataset) -> C_AXIS/Z_AXIS as literal text instead of code). — Claude

webknossos/webknossos/dataset/_image_conversion/image_source.py
normanrz: Expanded the docstring to explain why "t" specifically is special-cased here (and not some other iterated axis like "s"): the codebase already has a fixed t,c,x,y,z order (TCXYZ_AXES) for timepoint-carrying formats (tiff, OME-Zarr, ...), so inserting "c" right after "t" keeps that order. Kept the T_AXIS constant as you both agreed. — Claude

webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
normanrz: Good catch, yes it does assume that. Added an assertion (`assert self._iter_axes[-1] == "z" ...`) right where the assumption is used, referencing __init__ where the invariant is established. — Claude

webknossos/tests/dataset/test_slice_reader.py
normanrz: "Iterated" axes are the ones stepped through slice by slice (as opposed to "bundled" axes like y/x, which make up one slice). Reworded the comment to spell that out. — Claude

webknossos/webknossos/dataset/_image_conversion/image_source.py
normanrz: Follow-up: went further and dropped the t-special-casing entirely — a missing "c" axis is now always inserted first, unconditionally, so it no longer depends on where (or whether) "t" is present. Updated the one place that relied on the old t,c,x,y,z ordering to build its array (ChunkedImageSource.copy_chunk_to_view) and the tests asserting that order (now c,t,x,y,z / CTXYZ_AXES). — Claude

normanrz: Correction to my last update: I reverted the c-always-first change. Turned out t,c,x,y,z (TCXYZ_AXES) is load-bearing, not incidental -- test_compare_nd_tifffile, which compares against a real downloaded OME-TIFF read independently via the tifffile library, only passes with "c" inserted right after "t". So this is back to inserting "c" right after "t" when present, first otherwise (using the T_AXIS constant). While debugging this I also found and fixed three unrelated pre-existing bugs from the axis-constants refactor that were breaking most of the image-conversion test suite (bad C_AXIS/X_AXIS substitution in sliced_image_source.py, a lowercase/uppercase mismatch in czi_image_source.py's dimension check, and a wrong x/y/z vs z/y/x canonical order in tensorstore_chunked_image_source.py). Full test_add_layer_from_images.py + test_slice_reader.py + test_czi_image_source.py suite is now 162 passed / 8 skipped / 0 failed. — Claude

webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -91,6 +94,42 @@ def compute_channel_selection(
...
+    Unlike `NDBoundingBox.normalize_axes()` (which keeps a missing "c" axis
+    implicit, matching xyz-only boxes elsewhere, e.g. remote datasets),
+    image-conversion's own internal boxes are always meant to carry one, per
+    the codebase's `NormalizedBoundingBox` convention.
+
+    "t" is the only axis treated specially here, not some other iterated
+    axis (e.g. "s"): the codebase already has a fixed t,c,x,y,z axis order
+    (`TCXYZ_AXES`) for the timepoint-carrying formats this matters for
+    (tiffs, OME-Zarr, ...), so a missing "c" is inserted right after "t"
+    when "t" is present, to match that order, and first otherwise.
```
```suggestion
    If there is a "t" axis a fixed t,c,x,y,z axis order will be used, because that is a
    common convention. Other axes can appear at arbitrary locations.
```

Tobias314 (approved):

PR #1516, opened 2026-08-20
Add support for new bigint segmentId format and HTTP API v15

fm3:
### Description:
- WK now supports full uint64 range for segment ids, and introduced an envelope format for large segment ids in JSON (both API responses, now API v15 and datasource-properties.json files on disk)
- This PR adapts the libs to those changes.

### Corresponding WEBKNOSSOS PR:
- https://github.com/scalableminds/webknossos/pull/9765

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/client/context.py
+++ b/webknossos/webknossos/client/context.py
@@ -68,9 +68,11 @@ from ._defaults import DEFAULT_HTTP_TIMEOUT, DEFAULT_WEBKNOSSOS_URL
 from .api_client import (
     DatastoreApiClient,
     DatastoreApiClientV13,
+    DatastoreApiClientV14,
     TracingStoreApiClient,
     WkApiClient,
     WkApiClientV13,
+    WkApiClientV14,
 )
 
 load_dotenv()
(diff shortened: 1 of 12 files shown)


webknossos/webknossos/client/context.py
```diff
@@ -138,19 +138,33 @@ def organization_id(self) -> str:
 
     @cached_property
     def api_client(self) -> WkApiClient:
-        cls = WkApiClientV13 if self.api_version == 13 else WkApiClient
-        return cls(
+        if self.api_version == 13:
```
normanrz: Wouldn't it be more consistent if we had V14 (and the WkApiClient with v15) clients?

normanrz (approved):

webknossos/webknossos/client/context.py
fm3: Yes, this is taking a bit of a shortcut because the v14 client can also handle v15 servers now that the new structuring hook is in place everywhere (and can still handle the number format). It would be more explicit with a V14 client class rather than just passing the number in, but it would essentially be identical. What do you prefer?

normanrz: Let's make it explicit.

normanrz: Good point — added explicit `WkApiClientV14`/`DatastoreApiClientV14` classes so the version picking mirrors V13/V14/(default=v15) consistently, instead of passing `webknossos_api_version` at runtime. Pushed in 719eb0d9.

normanrz: Done in https://github.com/scalableminds/webknossos-libs/commit/719eb0d9b57f51341d7320a09d5fb7e0c7a15f0f

PR #1517, opened 2026-08-21
Fix UnexpectedStatusError/CannotHandleResponseError breaking when unpickled

normanrz:
## Summary

`UnexpectedStatusError` and `CannotHandleResponseError` (`webknossos/webknossos/client/api_client/errors.py`) crashed with `AttributeError: 'str' object has no attribute 'request'` whenever they were unpickled — for example when raised inside a `ProcessPoolExecutor`/`multiprocessing` worker and shipped back to the parent process.

## Root cause

Both exceptions require a live `httpx.Response` in `__init__` to build their message. `BaseException`'s default `__reduce__` pickles only `self.args` (the already-formatted message string) and reconstructs the exception on unpickling via `cls(*self.args)`. So unpickling called e.g. `UnexpectedStatusError(msg_string)`, which set `response` to that string. `request_label` then accessed `response.request`, raising the `AttributeError` seen in the reported traceback:

```
File ".../errors.py", line 19, in request_label
    if response.request is None:
       ^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'request'
```

## Fix

Added a `__reduce__` override on `ApiClientError` that reconstructs the exception via plain `Exception.__init__`, using the already-formatted message and bypassing the subclasses' custom `__init__` (which needs a live response object) during unpickling. This covers both `UnexpectedStatusError` and `CannotHandleResponseError` since they both inherit from `ApiClientError`.

## Testing

- Verified a direct `pickle.dumps`/`pickle.loads` round-trip on both exception types preserves subclass identity and message.
- Reproduced the original failure mode with a `ProcessPoolExecutor` worker raising `UnexpectedStatusError`; confirmed it is now caught correctly in the parent process with the full message intact.
- Ran `./precommit.sh` (ruff format, lint, mypy across `webknossos`, `tests`, `examples`, `script_collection`) — all pass.
- Added a `Changelog.md` entry under `Unreleased > Fixed`.

### DIFF
--- a/webknossos/Changelog.md
+++ b/webknossos/Changelog.md
@@ -20,6 +20,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 
 ### Fixed
 - Fixed that local dataset/layer/mag/attachment path resolution on Windows converted mapped/substituted network drives (e.g. `Z:\...`) to their UNC form (`\\server\share\...`), which TensorStore's local file driver rejected. [#1513](https://github.com/scalableminds/webknossos-libs/issues/1513)
+- Fixed that `UnexpectedStatusError` and `CannotHandleResponseError` raised an `AttributeError` when unpickled (e.g. when raised inside a `ProcessPoolExecutor` worker), instead of reproducing the original error.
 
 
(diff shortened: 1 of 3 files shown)


fm3 (approved):
LGTM, do you know what’s with the CI?

PR #1519, opened 2026-08-24
Fix View.num_channels defaulting to 0 without a channel axis

normanrz:
## Summary
`View.num_channels` fell back to `0` when the view's normalized bounding box has no `"c"` axis, which is incorrect — a missing channel axis means an implicit single channel, not zero channels.

## Change
Changed the fallback in `View.num_channels` from `0` to `1`, matching the default used elsewhere in the codebase (e.g. `pims_images.py`, `dataset.py`, `normalized_bounding_box.py`).


fm3 (approved):
may need changelog entry?

PR #1522, opened 2026-08-26
Recognize RGB TIFFs' samples axis as the channel axis

### LINKED ISSUE #1499: RGB TIFFs are converted as a 3-slice z-stack instead of an RGB layer
## Context

Found while working on multi-channel handling for TIFFs (`channels_are_rgb`, see `webknossos/dataset/_image_conversion/image_conversion.py`). RGB TIFFs turn out not to reach that logic at all: `TiffSliceReader` never reports a channel axis for them, so they're read as a z-stack before any RGB decision gets a chance to apply.

`tifffile` names the samples-per-pixel axis `S`, and only names it `C` when the file carries explicit axis metadata. [tiff_slice_reader.py:34](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/_image_conversion/tiff_slice_reader.py#L34) copies the series' axes verbatim, and [sliced_image_source.py:92](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py#L92) derives the channel count with `images.sizes.get("c", 1)`. For an ordinary RGB TIFF (`YXS`) there is no `"c"` key at all, so `raw_num_channels` is silently `1` and the three samples fall through to `_other_axes`/`_page_axes` instead, where they're iterated as pages — i.e. read as z-slices.

## Expected Behavior

An RGB TIFF should convert like the same pixels in any other colour format: **one layer with `num_channels=3`**, `z=1`.

The equivalent RGB PNG already does exactly that, via `channels_are_rgb=True` on `SingleImageSliceReader`.

## Current Behavior

The three colour samples become three z-slices of a single-channel layer — an 8×16 RGB image converts into a 16×8×3 grayscale stack whose "slices" are the red, green and blue planes.

Reproduced directly against `open_image_source`:

```python
from upath import UPath
from webknossos.dataset._image_conversion.image_source_registry import open_image_source
from webknossos.dataset._image_conversion.image_source import ReadOptions

# RGB TIFF (testdata/rgb_tiff/test_rgb.tif, 32x32 pixels, axes=YXS shape=(32, 32, 3))
tiff_src = open_image_source(UPath("testdata/rgb_tiff/test_rgb.tif"), ReadOptions())
print(tiff_src.num_channels, tiff_src.expected_bbox, tiff_src.channels_are_rgb)
# 1  BoundingBox(topleft=(0, 0, 0), size=(32, 32, 3))  False   <- 3 samples read as z

# same pixels as an RGB PNG
png_src = open_image_source(UPath("rgb.png"), ReadOptions())
print(png_src.num_channels, png_src.expected_bbox, png_src.channels_are_rgb)
# 3  BoundingBox(topleft=(0, 0, 0), size=(16, 8, 1))  True
```

The repo's own fixture shows it too: `testdata/rgb_tiff/test_rgb.tif` is `axes=YXS shape=(32, 32, 3)`, and the expectation recorded for it in `REPO_IMAGES_ARGS` ([test_add_layer_from_images.py:1022](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/tests/dataset/test_add_layer_from_images.py#L1022)) is `c=1` with `z` derived from the sample count — i.e. the current behaviour is pinned by a test, so fixing this means updating that expectation. There is a newer test covering the RGB decision, [test_rgb_image_stays_a_single_layer](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/tests/dataset/test_add_layer_from_images.py#L626), but it only exercises PNG, not TIFF, so it doesn't catch this.

## Steps to Reproduce the bug

```python
import numpy as np, tifffile
from PIL import Image
from upath import UPath
from webknossos.dataset._image_conversion.image_source_registry import open_image_source
from webknossos.dataset._image_conversion.image_source import ReadOptions

data = np.zeros((8, 16, 3), dtype="uint8")   # 8 high, 16 wide, RGB
tifffile.imwrite("rgb.tif", data, photometric="rgb")
Image.fromarray(data, mode="RGB").save("rgb.png")

def describe(path):
    src = open_image_source(UPath(path), ReadOptions())
    return src.num_channels, src.expected_bbox

print(describe("rgb.tif"))   # (1, size=(16, 8, 3))  <- three z-slices
print(describe("rgb.png"))   # (3, size=(16, 8, 1))   <- one RGB layer
```

Both `Dataset.from_images` and `Dataset.add_layer_from_images` inherit this, since it happens in the reader that both go through.

## Notes towards a fix

Treating `S` as a channel axis in `TiffSliceReader`/`SlicedImageSource` is the obvious direction, but it needs care:

- `S` and `C` can both be present (e.g. axes `ZCYXS`), so `S` cannot simply be renamed to `C` — a synthetic `ZCYXS` file currently reports `num_channels=3` from the real `C` axis while `S` survives as its own extra dimension in the bounding box (`possible_layers={"channel": [0, 1, 2]}`, bbox axis `s` of size 3), and that behaviour should be preserved.
- `TiffSliceReader.get_slice_2d` already indexes away page axes that aren't in `bundle_axes`, which is how `S` is dropped today for the plain-RGB case; the frame-getter registration (`self._set_get_slice(self.get_slice_2d, "cyx")` vs `"yx"` at [tiff_slice_reader.py:53-56](https://github.com/scalableminds/webknossos-libs/blob/master/webknossos/webknossos/dataset/_image_conversion/tiff_slice_reader.py#L53-L56)) would have to account for a samples axis as well.
- A samples axis of 3 with `photometric="rgb"` is unambiguous colour; other sample counts (2, 4, …) are not, and should presumably keep whatever they do now rather than being guessed at.
- Once TIFF reports real channels for the `S`-only case, the existing RGB decision in `image_conversion.py` (`_channels_are_one_rgb_layer` / `_RGB_IMAGE_EXTENSIONS`) should already do the right thing without further changes: `.tif`/`.tiff` are deliberately left out of `_RGB_IMAGE_EXTENSIONS` (TIFF usually holds separate acquisition channels, not colour), but the fallback a few lines below — "not one of the RGB image formats, but three uint8 channels still display as RGB in WEBKNOSSOS, so write them into one layer instead of erroring" (`image_conversion.py` around line 547) — combines exactly 3 uint8 channels into one layer by default (`allow_multiple_layers=False`) anyway. So no second allowlist decision should be needed here, just getting `TiffSliceReader` to report the channel in the first place.

## Your Environment for bug

- Operating System and version: macOS (Darwin 25.5.0)
- Version of webKnossos-libs (Release or Commit): confirmed present on `release-4.0`, also present on `master`

normanrz:
### Description:

- A samples axis of exactly 3, tagged `photometric=rgb`, with `uint8` data is now recognized as the channel axis (matching WEBKNOSSOS's existing rule that only three uint8 channels display as RGB). `S` and a real `C` axis can coexist (e.g. `ZCYXS`); in that case `S` is left alone.

- Verifying this against the issue's own `mag=2` test fixture surfaced a second, pre-existing bug: converting a single 2D multi-channel image at `mag` greater than 1 crashed with a mag-alignment `AssertionError`.

### Issues:
- fixes #1499

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/_image_conversion/sliced_image_source.py
```diff
@@ -475,12 +475,17 @@ def chunk_grid(
     ) -> list[NormalizedBoundingBox]:
         """Batches of z-slices spanning the full x/y extent, which cannot be
         split while it is still the placeholder."""
-        del mag
-        batch_size = self._resolve_batch_size(mag_view, batch_size)
+        # _resolve_batch_size returns a count of mag-native voxels (it
+        # defaults to and validates against the view's own chunk/shard z
+        # size), but layer_bounding_box is in mag1 space, so it has to be
+        # scaled up by the mag factor before it can chunk that box.
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/tiff_slice_reader.py
```diff
@@ -32,20 +32,42 @@ def __init__(self, path: UPath) -> None:
...
+
+            # tifffile names the samples-per-pixel axis "S", only naming it "C"
+            # when the file carries explicit multi-channel axis metadata. A
+            # samples axis of 3 tagged as RGB is unambiguous colour, so it is
+            # treated as the channel axis rather than as extra z-slices,
+            # matching WEBKNOSSOS's own rule that only three uint8 channels
+            # display as RGB (see channels_fit_one_layer). "S" and "C" can
+            # both be present (e.g. ZCYXS); in that case "S" is left alone
+            # and stays a plain extra axis.
+            treat_s_as_c = (
```
normanrz: maybe this comment and variable name should be made after the fact that the tiff is RGB because of the photometric tag instead of focussing on c over s?

normanrz: Done in 0e26fc7 — renamed `treat_s_as_c` to `is_rgb_tiff` and reworded the comment to lead with the tiff being unambiguous RGB (photometric tag + 3 uint8 samples), with the S→C rename now framed as a consequence of that rather than the main point.

— Claude

webknossos/webknossos/dataset/_image_conversion/tiff_slice_reader.py
```diff
@@ -32,20 +32,42 @@ def __init__(self, path: UPath) -> None:
...
                 _tmp = _tiff.pages[0]
             else:
                 _tmp = _tiff["pages"][0]  # type: ignore
             assert _tmp is not None, "No pages found in tiff file."
             self._dtype = _tmp.dtype or np.dtype("uint8")
             self._shape = _tmp.shape
+
+            # A tiff tagged photometric=rgb with 3 uint8 samples is
+            # unambiguous colour, matching WEBKNOSSOS's own rule that only
+            # three uint8 channels display as RGB (see channels_fit_one_layer).
```
fm3:
```suggestion
            # A tiff tagged photometric=rgb with 3 uint8 samples should propagate as rgb/uint24 to the output.
```

webknossos/Changelog.md
```diff
@@ -56,6 +56,8 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
...
 - Fixed that renaming a layer of a zarr-streamed `RemoteDataset` (where layer metadata cannot be persisted) raised an opaque `StopIteration` instead of the expected `RuntimeError` explaining that the layer is read-only. [#1518](https://github.com/scalableminds/webknossos-libs/pull/1518)
 - Fixed that `UnexpectedStatusError` and `CannotHandleResponseError` raised an `AttributeError` when unpickled (e.g. when raised inside a `ProcessPoolExecutor` worker), instead of reproducing the original error. [#1517](https://github.com/scalableminds/webknossos-libs/pull/1517)
+- Fixed RGB TIFFs converting into a single-channel z-stack (the 3 colour samples read as z-slices) instead of one RGB layer, since `tifffile` names the samples-per-pixel axis `S`, not `C`. A samples axis of exactly 3 tagged `photometric=rgb` with `uint8` data is now recognized as the channel axis, matching how the same pixels convert from a `.png`. [#1499](https://github.com/scalableminds/webknossos-libs/issues/1499) [#1522](https://github.com/scalableminds/webknossos-libs/pull/1522)
```
fm3:
```suggestion
- Fixed RGB TIFFs converting into a single-channel z-stack (the 3 channels read as z-slices) instead of one RGB layer, since `tifffile` names the samples-per-pixel axis `S`, not `C`. A samples axis of exactly 3 tagged `photometric=rgb` with `uint8` data is now recognized as the channel axis, matching how the same pixels convert from a `.png`. [#1499](https://github.com/scalableminds/webknossos-libs/issues/1499) [#1522](https://github.com/scalableminds/webknossos-libs/pull/1522)
```

fm3 (approved):

PR #1525, opened 2026-08-28
Store the observed data min/max in the layer view configuration

### LINKED ISSUE #1510: Store min/max in view config
## Detailed Description
During image conversion, all image data is read. The conversion pipeline should store min and max values and put it into the view configuration.

normanrz:
### Description:
- Image conversion already reads every voxel, so the value range it observes is now kept and written to the color layer's default view configuration as `min`/`max`. WEBKNOSSOS no longer has to fall back to the full dtype range.
- `intensityRange` is left alone — WEBKNOSSOS derives it from a real histogram with outlier clipping (`clip_histogram_saga.ts`), which a plain min/max would only approximate badly.
- NaNs and infinities are ignored: one of them should not wipe out the range of an otherwise ordinary image, and neither is any use as a display bound (an infinite one would also be written to `datasource-properties.json` as invalid JSON).
- Values supplied by the source format itself (e.g. an OME-Zarr `omero` window) take precedence; only gaps are filled.
- Segmentation layers are unaffected: they keep recording their maximum in `largestSegmentId`.

Each chunk reports a `ValueRange` next to the `max_value` it already reported, and the ranges are combined once the layer is written. `max_value` stays a separate exact integer, since uint64 segment ids do not survive a round trip through `float`.

`testdata/tiff/datasource-properties.zarr-fixture.json` gains the `defaultViewConfiguration` a conversion now produces, which is what `test_tiff_cubing_zarr_s3` compares against.

### Issues:
- fixes #1510

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -138,6 +139,12 @@ class ChunkResult(NamedTuple):
     """The largest value seen, for `largest_segment_id` on segmentation layers.
     None when the source cannot report one."""
 
+    statistics: ValueStatistics | None = None
+    """The range and distribution of the written values, for the layer's
+    default view configuration. Separate from `max_value`, which stays an exact
+    integer because uint64 segment ids do not survive a round trip through
+    `float`. None when the chunk held no usable value."""
```
```suggestion
    """The range and distribution of the written values, for the layer's
    default view configuration. Separate from `max_value`, which stays an exact
    integer for segmentations. None when the chunk held no usable value."""
```

webknossos/Changelog.md
```diff
@@ -32,6 +32,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
...
 - Added support for WEBKNOSSOS API version 15, which allows segment ids to use the full uint64 range. Such ids may now be serialized as `{"customJsonEncoding": "bigint", "value": "<decimal string>"}` instead of a plain JSON number, both in API responses and in `datasource-properties.json`.
 - Added `neuroglancerPrecomputed` as a valid `AttachmentDataFormat` for `MeshAttachment`s. [#1518](https://github.com/scalableminds/webknossos-libs/pull/1518)
+- `Dataset.from_images`/`add_layer_from_images` now derive a color layer's default view configuration from the data they read: `min`/`max` from the observed value range, and `intensityRange` from an approximate histogram clipped the way the WEBKNOSSOS "clip histogram" button clips it, so sparse outliers do not wash out the display. Values provided by the source format itself (e.g. an OME-Zarr `omero` window) take precedence. [#1525](https://github.com/scalableminds/webknossos-libs/pull/1525)
```
```suggestion
- `Dataset.from_images`/`add_layer_from_images` now derive a color layer's default view configuration from the data they read: `min`/`max` from the observed value range, and `intensityRange` from an approximate histogram. Values provided by the source format itself (e.g. an OME-Zarr `omero` window) take precedence. [#1525](https://github.com/scalableminds/webknossos-libs/pull/1525)
```

webknossos/webknossos/dataset/_image_conversion/image_conversion.py
```diff
@@ -825,6 +829,35 @@ def add_layer_from_images(
             # Could not guess the category; keep the one already set.
             pass
 
+        # The category guess may have replaced the layer object.
+        final_layer = dataset._layers[layer.name]
+        if final_layer.category == COLOR_CATEGORY and statistics is not None:
+            # The observed value range becomes the layer's min/max, and the
+            # histogram gathered along with it an intensity range clipped the
+            # way WEBKNOSSOS clips it.
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/value_statistics.py
```diff
@@ -0,0 +1,212 @@
...
+
+DEFAULT_THRESHOLD_RATIO = 0.0001
+"""Share of the values clipped off, split between both ends. Same default as
+WEBKNOSSOS's own histogram clipping."""
+
+HISTOGRAM_SAMPLE_SIZE = 65_536
+"""How many values are binned per call at least, when there are that many.
+Everything above is sampled, which keeps the cost down while leaving far more
+values than the clipping threshold needs. WEBKNOSSOS samples too, from three
+viewport planes."""
```
normanrz:
```suggestion
values than the clipping threshold needs."""
```

fm3:
Claude review says this, what do you think of it? Also, do you know what’s with the tests? Will do human review tomorrow.

```
  1. image_conversion.py:830 — low, high = statistics.value_range is written straight into LayerViewConfiguration.min/max, but value_statistics.py documents that value_range deliberately
     keeps ±inf when source data has infinities (only the histogram range is narrowed to finite values). A float image with inf/-inf could end up with infinite min/max in the view config,
     which _save_dataset_properties() then writes to datasource-properties.json right after — likely producing invalid/non-serializable JSON.
  2. value_statistics.py ValueStatistics.combined() — the single-entry fast path returns counts as uint32, while the multi-entry merge path builds it as int64. Same documented field,
     inconsistent dtype depending on how many chunks got merged — a latent trap for any caller assuming a fixed dtype.
```

normanrz:
Both findings were real — fixed in 497ef6cb, along with the CI failure. (Claude wrote this reply and the fix.)

**1. Infinities in `min`/`max` — correct, and it would have produced invalid JSON.** `value_range` kept `±inf` while only the histogram range was narrowed to finite values, so a float image containing an infinity would have written `Infinity` into `datasource-properties.json`. Infinities are now left out of the value range too, not just the histogram: an infinite display bound is useless anyway. Data whose values are *only* NaN/infinite now yields no statistics at all, same as all-NaN did before. `test_view_configuration_min_max_ignores_non_finite` converts a float32 TIFF containing NaN, `inf` and `-inf` and asserts the view configuration comes out finite.

**2. `combined()` dtype.** `combined()` now always returns `int64`, including on the single-entry path. The narrow `uint32` is deliberate — every chunk hands its counts back to the parent process and it halves that payload — so the dtype is now a property of which method produced the value rather than of how many chunks there were, and the field documents it. `test_combined_counts_are_wide_whatever_it_combined` pins it.

**The tests:** `test_tiff_cubing_zarr_s3` compares the produced `datasource-properties.json` against `testdata/tiff/datasource-properties.zarr-fixture.json`, which has no `defaultViewConfiguration` — the conversion now adds one, so the fixture needed updating. Nothing else was failing. The values are `intensityRange [9.0, 254.0]`, `min 9.0`, `max 255.0`, identical across Linux 3.10–3.14, Windows, and locally with both `--jobs 2` and a sequential executor, so the fixture is safe to pin. For reference, an exact port of `clip_histogram_saga.ts` run over that same data gives `(10, 254)` — one grey level apart, which is the sampling.

webknossos/Changelog.md
```diff
@@ -35,6 +35,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
...
 - Added support for WEBKNOSSOS API version 15, which allows segment ids to use the full uint64 range. Such ids may now be serialized as `{"customJsonEncoding": "bigint", "value": "<decimal string>"}` instead of a plain JSON number, both in API responses and in `datasource-properties.json`. [#1516](https://github.com/scalableminds/webknossos-libs/pull/1516)
 - Added `neuroglancerPrecomputed` as a valid `AttachmentDataFormat` for `MeshAttachment`s. [#1518](https://github.com/scalableminds/webknossos-libs/pull/1518)
+- `Dataset.from_images`/`add_layer_from_images` now store the value range observed while reading the images as the color layer's default view configuration `min`/`max`, so WEBKNOSSOS no longer has to assume the full dtype range. Values provided by the source format itself (e.g. an OME-Zarr `omero` window) take precedence. [#1525](https://github.com/scalableminds/webknossos-libs/pull/1525)
```
```suggestion
- `Dataset.from_images`/`add_layer_from_images` now store the value range observed while reading the images as the color layer's default view configuration `min`/`max`. Values provided by the source format itself (e.g. an OME-Zarr `omero` window) take precedence. [#1525](https://github.com/scalableminds/webknossos-libs/pull/1525)
```

fm3 (approved):
LGTM thanks :) I also tested locally for a small test dataset and the stored view config is handled by webknossos as expected.

PR #1527, opened 2026-08-30
Allow passing a dataset id to RemoteDataset.open

normanrz:
### Description:
- `RemoteDataset.open("59e9cfbdba632ac2ab8b23b5")` now works: if the first positional argument is shaped like a dataset id (24 hex chars) and such a dataset exists, it is used as the dataset id. Otherwise it is resolved as a dataset name as before, so existing calls are unaffected.
- Implemented in the shared `_parse_remote`, so `Dataset.open_remote`, `Dataset.download` and `trigger_reload_in_datastore` benefit too.

### Issues:
- n/a

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests

### DIFF
--- a/webknossos/webknossos/dataset/remote_dataset.py
+++ b/webknossos/webknossos/dataset/remote_dataset.py
@@ -1349,6 +1349,7 @@ class RemoteDataset(AbstractDataset[RemoteLayer, RemoteSegmentationLayer]):
         else:
             annotation_id = None
 
+        dataset_name: str | None = None
         if dataset_id is None:
             assert dataset_name_or_url is not None, (
                 f"Please supply either a dataset_id or a dataset name or url to Dataset.{caller}()."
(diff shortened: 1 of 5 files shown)


webknossos/webknossos/dataset/remote_dataset.py
```diff
@@ -1377,7 +1394,14 @@ def _parse_remote(
                 dataset_name = dataset_name_or_url
                 organization_id = organization_id or current_context.organization_id
 
-                dataset_id = cls._disambiguate_remote(dataset_name, organization_id)
+                # An id-shaped argument is treated as a dataset id, unless no such
+                # dataset exists. Then it is resolved as a dataset name.
+                if _DATASET_ID_REGEX.fullmatch(dataset_name) and cls._dataset_id_exists(
```
fm3: should the exists check also use the context constructed below? or is this irrelevant for this case?

fm3 (approved):
long-needed addition :)

webknossos/webknossos/dataset/remote_dataset.py
normanrz: Good catch — it did matter: with an explicit `webknossos_url=` (or a deprecated-style URL, which already had this issue for the name lookup), the lookup ran against the current context instead of the requested instance.

Restructured `_parse_remote` in 90557e45 so the URL branches only extract fields, the context manager is built right after, and both the id-exists check and `_disambiguate_remote` run inside `with context_manager:`. (Comment written by Claude.)

PR #1528, opened 2026-08-30
Leave chunk memory layout to the storage backends in image conversion

normanrz:
### Description:

`ChunkedImageSource.copy_chunk_to_view` materialized every chunk with `astype(dtype, order="F")` before writing it. That was the most expensive step in the conversion pipeline and most-often unneccessary. 

This surfaced on a real 25024×33502×8355 uint64 segmentation with 1024³ shards, where a hand-written tensorstore→`MagView` copy of all 9 scales finished in 6 h while `from_images` was under 50% of mag1 after 48 h.

Also, don't write all-background shards at all.

### Todos:
 - [x] Updated Changelog
 - [ ] Updated Documentation — n/a, no API change
 - [ ] Added / Updated Tests — behaviour is unchanged, covered by the existing conversion tests:
    - `test_add_layer_from_images.py`, `test_from_images.py`, `test_ozx_image_source.py`, `test_czi_image_source.py`, `test_neuroglancer_precomputed_image_source.py`, `test_zarr_image_source.py`, `test_n5_image_source.py`: 201 passed, 8 skipped
    - additionally verified 512³ uint64 precomputed → Zarr3 **and** → WKW conversions byte-for-byte against the source through the changed path
 - [ ] Considered adding this to the Examples — n/a


webknossos/webknossos/dataset/_image_conversion/image_source.py
```diff
@@ -202,7 +202,9 @@ def copy_chunk_to_view(
         * `ReadOptions.flip_x` mirrors the source's **y** axis and `ReadOptions.flip_y` its **x** axis
           — the axes are named for the output, not the source. Each flip
           mirrors the whole extent, never one chunk in isolation.
-        * When `dtype` is given, data is converted with `order="F"`.
+        * When `dtype` is given, data is converted to it. The memory layout
+          is left alone — each `BaseArray` backend arranges the chunk the way
+          its format needs.
```
```suggestion
        * When `dtype` is given, data is converted to it, but the memory layout
          is left alone.
```

webknossos/webknossos/dataset/_image_conversion/tensorstore_chunked_image_source.py
```diff
@@ -161,4 +161,8 @@ def _read_source_box(
             if not has_z:
                 block = block[np.newaxis]  # add a size-1 z axis
             slabs.append(block)
+        if len(slabs) == 1:
+            # Adding the channel axis is a view; np.stack would copy the whole
+            # block just to prepend an axis of size 1.
```
normanrz:
```suggestion
```

webknossos/Changelog.md
```diff
@@ -45,6 +45,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - `Dataset.add_layer_from_images`/`from_images` now verify the dataset properties are internally consistent after converting each layer and raise an error if not, instead of silently continuing. [#1477](https://github.com/scalableminds/webknossos-libs/pull/1477)
 
 ### Fixed
+- Sped up `Dataset.from_images`/`add_layer_from_images` for the chunk-based formats (Zarr, OME-Zarr, `.ozx`, N5, Neuroglancer precomputed, `.ims`, MRC, `.czi`). Every chunk was materialized into Fortran order before being written, which for a large shard is a full cache-hostile transposing copy — measurably the most expensive step of the conversion. The memory layout is now left to the storage backends, which arrange each chunk the way their format needs anyway. [#1528](https://github.com/scalableminds/webknossos-libs/pull/1528)
```
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
```diff
@@ -194,8 +194,12 @@ def copy_chunk_to_view(
             else block.transpose(0, 2, 3, 1)
         )
 
-        if dtype is not None:
-            block = block.astype(dtype, order="F")
+        # Only the dtype is forced, never the memory layout: each BaseArray
+        # backend arranges the chunk the way its format needs (tinywkw gathers
+        # its chunks with `tobytes(order="F")`, tensorstore consumes arbitrary
+        # strides), and materializing a large shard here just to reorder it is
+        # the most expensive step of the whole conversion.
```
normanrz:
```suggestion
        # Only the dtype is forced, never the memory layout.
        # Materializing a large shard here just to reorder would be
        # the most expensive step of the whole conversion.
```

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
```diff
@@ -194,18 +194,24 @@ def copy_chunk_to_view(
...
             # Add back the size-1 "t" axis this chunk corresponds to.
             block = block[np.newaxis]
 
-        # allow_unaligned=True: border chunks are smaller than shard_shape,
-        # since extents rarely divide evenly. Safe because parallel jobs write
-        # disjoint regions.
-        mag_view.write(block, absolute_bounding_box=bbox, allow_unaligned=True)
+        # An all-background chunk is left unwritten. Both backends discard an
+        # empty shard anyway, but only after compressing it. Safe because the
+        # mag was just created and the jobs write disjoint regions.
```
normanrz:
```suggestion
        # Safe because the mag was just created and the jobs write disjoint regions.
```

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
```diff
@@ -194,18 +194,22 @@ def copy_chunk_to_view(
             else block.transpose(0, 2, 3, 1)
         )
 
-        if dtype is not None:
-            block = block.astype(dtype, order="F")
+        # Only the dtype is forced, never the memory layout.
+        # Materializing a large shard here just to reorder would be
+        # the most expensive step of the whole conversion.
```
fm3:
```suggestion
```

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
```diff
@@ -194,18 +194,22 @@ def copy_chunk_to_view(
...
         max_value = int(block.max())
         if "t" in relative_bbox.axes:
             # Add back the size-1 "t" axis this chunk corresponds to.
             block = block[np.newaxis]
 
-        # allow_unaligned=True: border chunks are smaller than shard_shape,
-        # since extents rarely divide evenly. Safe because parallel jobs write
-        # disjoint regions.
-        mag_view.write(block, absolute_bounding_box=bbox, allow_unaligned=True)
+        # Safe because the mag was just created and the jobs write disjoint regions.
```
fm3:
```suggestion
```

fm3 (approved):
I don’t fully understand why this still works. If the underlying data is not F, the output should be different now, no? Or is a different step already doing this reordering?

The tests are still green, so obviously it still works, but could you explain for me why it does?

normanrz:
> I don’t fully understand why this still works. If the underlying data is not F, the output should be different now, no? Or is a different step already doing this reordering?
> 
> The tests are still green, so obviously it still works, but could you explain for me why it does?

Haha. I actually told Claude to remove the comment about how the array classes in _array.py already take care of F-ordering. So, yes, tensorstore, wkw, tinywkw already take care of transforming to F-order. So it doesn't make sense to do that twice.

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
normanrz: I'd leave that. If you had an existing mag, it would not be safe, because we'd want to deliberately overwrite the shards.

fm3: sure, but the same line appears two lines further down. Or is this specifically about the `any`?

fm3:
ok cool thanks!

webknossos/webknossos/dataset/_image_conversion/chunked_image_source.py
normanrz: Ok. That is indeed dumb.

normanrz:
```suggestion
        # Skipping empty shards is safe, because the mag was just created and, therefore, is empty.
```

PR #1529, opened 2026-08-31
Speed up median and mode downsampling filters

normanrz:
### Description:
This speeds up downsampling 3-6x and reduces peak memory.

Reasons:
- `non_linear_filter_3d` caused 3 full copies (also 3x peak memory, yikes!)
- median spent most time in `np.median`'s `ndarray.partition`

Solution: numba kernels that read the source buffer directly

Also:
- Replaces `np.all(cube_buffer == 0)` with `cube_buffer.any()`
- Moved numba into an extra dependency, with (slow) numpy fallbacks

### Issues:
- n/a

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


webknossos/Changelog.md
```diff
@@ -19,6 +19,8 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Added `neuroglancerPrecomputed` as a valid `AttachmentDataFormat` for `MeshAttachment`s. [#1518](https://github.com/scalableminds/webknossos-libs/pull/1518)
 
 ### Changed
+- Sped up the `median` and `mode` downsampling filters. Both now read the source buffer directly instead of reshaping it into an (elements-per-block, num-blocks) array first, which cost three full copies of the buffer. Downsampling a uint8 color layer by a factor of 2 in each dimension is about 6x faster, a uint32 segmentation layer about 3x, and the peak memory per worker drops accordingly. The results are unchanged. [#1529](https://github.com/scalableminds/webknossos-libs/pull/1529)
```
```suggestion
- Sped up the layer downsampling (`median` and `mode` filters), by 3x-6x with less peak memory consumed. [#1529](https://github.com/scalableminds/webknossos-libs/pull/1529)
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -216,32 +222,125 @@ def _median(x: np.ndarray) -> np.ndarray:
...
+        target[:, :, z_start:z_stop] = best
+    return target
+
+
+def _supports_fast_path(data: np.ndarray, factors: list[int]) -> bool:
+    return data.ndim == 3 and len(factors) == 3
+
+
+def _warn_once_about_missing_numba() -> None:
+    global _warned_about_missing_numba
```
normanrz: Maybe use an env var like in https://github.com/scalableminds/webknossos-libs/blob/38a18cda7a70b222277cbee5f902d2ec9cfb94e9/webknossos/webknossos/dataset/_image_conversion/image_source_registry.py#L252

webknossos/webknossos/dataset/layer/_downsampling_numba.py
```diff
@@ -0,0 +1,267 @@
...
+
+    return output_array
+
+
+# These kernels read the source buffer directly, which avoids the three full copies
+# that reshaping it into an (elements-per-block, num-blocks) array first costs.
+# Within a block the elements are visited in the order `dy + fy * dx + fx * fy * dz`,
+# which determines how the mode filter breaks ties.
+
+
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -216,32 +222,72 @@ def _median(x: np.ndarray) -> np.ndarray:
...
+
+
+def _warn_once_about_missing_numba() -> None:
+    # The environment is inherited by worker processes, so they stay quiet as well.
+    if environ.get("WEBKNOSSOS_SHOWED_MISSING_NUMBA_WARNING", "False") == "False":
+        environ["WEBKNOSSOS_SHOWED_MISSING_NUMBA_WARNING"] = "True"
+        warnings.warn(
+            "[INFO] Downsampling uses the numpy implementation of the non-linear "
+            "filters. Install the optional numba dependency with "
+            "'pip install webknossos[numba]' for a faster implementation.",
```
normanrz:
```suggestion
            "[INFO] Downsampling uses the numpy implementation, which is slow. "
            "Install the optional numba dependency with "
            "'pip install webknossos[numba]' for a faster implementation.",
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -216,32 +222,72 @@ def _median(x: np.ndarray) -> np.ndarray:
...
+    return best
+
+
+def _supports_fast_path(data: np.ndarray, factors: list[int]) -> bool:
+    # Numba rejects arrays with a non-native byte order, so those keep using numpy.
+    return data.ndim == 3 and len(factors) == 3 and data.dtype.isnative
+
+
+def _warn_once_about_missing_numba() -> None:
+    # The environment is inherited by worker processes, so they stay quiet as well.
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -216,32 +222,72 @@ def _median(x: np.ndarray) -> np.ndarray:
...
+            "filters. Install the optional numba dependency with "
+            "'pip install webknossos[numba]' for a faster implementation.",
+            category=UserWarning,
+            stacklevel=2,
+        )
+
+
+def warn_if_numba_is_missing(interpolation_mode: InterpolationModes) -> None:
+    """Emits the warning in this process, so that worker processes started afterwards
+    inherit the environment flag and stay quiet."""
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -18,6 +18,12 @@
 
 from .view import ArrayInfo, View
 
+try:
+    from . import _downsampling_numba
+except ImportError:
+    # numba is an optional dependency, the numpy implementations below are used instead.
```
normanrz:
```suggestion
```

webknossos/webknossos/dataset/layer/_downsampling_numba.py
```diff
@@ -0,0 +1,226 @@
+"""Numba implementations of the non-linear downsampling filters.
+
+This module is only imported when the optional `numba` dependency is installed;
+`_downsampling_utils` falls back to numpy implementations otherwise.
+"""
+
+from collections.abc import Callable
```
normanrz:
```suggestion
from collections.abc import Callable
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -216,32 +221,69 @@ def _median(x: np.ndarray) -> np.ndarray:
...
+    return best
+
+
+def _supports_fast_path(data: np.ndarray, factors: list[int]) -> bool:
+    # Numba rejects arrays with a non-native byte order, so those keep using numpy.
+    return data.ndim == 3 and len(factors) == 3 and data.dtype.isnative
+
+
+def _warn_once_about_missing_numba() -> None:
+    if environ.get("WEBKNOSSOS_SHOWED_MISSING_NUMBA_WARNING", "False") == "False":
```
fm3, line 252: this means once per multiprocessing process, right? is probably ok.

webknossos/webknossos/dataset/layer/layer.py
```diff
@@ -1164,6 +1165,7 @@ def downsample_mag(
         parsed_interpolation_mode = parse_interpolation_mode(
             interpolation_mode, self.category
         )
+        warn_if_numba_is_missing(parsed_interpolation_mode)
```
fm3: skip if median on float is selected? There numba wouldn’t help, right?

Also, maybe move this below the mag checks, so that users won’t get a warning and then a value error?

fm3 (approved):
Sweet!

Did you check if the new tests are ok fast?

webknossos/webknossos/dataset/layer/_downsampling_utils.py
normanrz: Done in d7c441c7 — switched to `WEBKNOSSOS_SHOWED_MISSING_NUMBA_WARNING`, following the image reader pattern, and to a `[INFO]` `UserWarning` to match.

One thing the env var alone didn't fix: it only suppresses duplicates if the *parent* sets it before the workers are created, and the parent never runs the filters — all downsampling happens inside jobs. Measured with the default multiprocessing executor:

| | warnings emitted |
|---|---|
| parent never touches the fallback (6 workers, 12 jobs) | 3 |
| parent hits it first, then spawns workers | 0 |

So the warning is now also emitted from `downsample_mag` before `wrap_executor` starts the pool. Verified end to end with numba uninstalled: exactly one warning.

(Written by Claude.)

webknossos/webknossos/dataset/layer/_downsampling_utils.py
normanrz: It was once per process when you wrote this, yes. I measured it with the default multiprocessing executor and got 3 warnings from 6 workers over 12 jobs.

Fixed in d7c441c7: `downsample_mag` now emits the warning in the parent, before `wrap_executor` creates the pool, so the workers inherit `WEBKNOSSOS_SHOWED_MISSING_NUMBA_WARNING=True` from the environment and stay quiet. Same setup now emits exactly one.

It is still once per process for anyone calling `downsample_cube` directly in their own pool, since nothing sets the flag ahead of that — but that path has no parent-side hook to hang it on.

(Written by Claude.)

webknossos/webknossos/dataset/layer/layer.py
normanrz: Both applied in 83b2e609, thanks.

**Float median:** correct, numba wouldn't help there. `_downsample_median` routes floating point data to `non_linear_filter_3d` either way, because `np.median` accumulates in the input dtype for float32 and has its own NaN semantics, so the numba kernel would not be bit-identical. `warn_if_numba_is_missing` now takes the layer dtype and stays quiet for that combination:

```python
if interpolation_mode == InterpolationModes.MODE or (
    interpolation_mode == InterpolationModes.MEDIAN and dtype.kind in "iu"
):
```

MODE on float still warns — the mode kernel handles floats fine, so numba does help there.

**Placement:** moved below the mag checks and below the `only_setup_mag` return, so an invalid call raises without warning first and setup-only calls stay quiet. It has to stay in front of `wrap_executor` so the worker processes still inherit the flag, so it now sits just above `bb_mag1 = ...`.

Covered by a parametrized test over the four mode/dtype combinations. One thing the tests caught: `self.dtype_per_channel` is deprecated, so it uses `self.dtype`.

(Written by Claude.)

Issue #1503, opened 2026-08-14
Speed up the webknossos test suite


normanrz:
# Speed up the webknossos test suite

## Context

Running the tests during a Claude session (or any edit–test loop) is slow enough
to block iteration. I measured rather than guessed. Baseline, serial, everything
except the WEBKNOSSOS-server-dependent files:

```
749 passed, 12 failed, 16 skipped in 1261.61s (0:21:01)
```

`call` 766.0 s / `setup` 357.8 s / `teardown` 4.4 s. Work is very lopsided — of
538 timed calls, the top 10 are 27 % of call time, the top 50 are 61 %.

Capability markers are out of scope, as decided.

---

## Finding 1 — the slow tests are slow because of a 1024³ default shard shape

This is the big one, and it is not about how much data the tests move.

`DEFAULT_SHARD_SHAPE` is `32 × 32 = 1024³` voxels
([defaults.py:23](https://github.com/scalableminds/webknossos-libs/blob/4b8048fc1dbdeab1998ba3c1d13dc3a969108684/webknossos/webknossos/dataset/defaults.py#L23)). The slowest
tests write *kilobytes* into an array whose single shard covers 1024³ voxels —
for a 3-channel uint8 layer, 3 GB of logical volume and a 32 768-entry chunk
index. Compressing or rewriting touches the whole shard.

`test_compression[zarr3]` writes **18 KB** at offset (60, 80, 100) and takes
16.7 s. Timing its steps directly:

```
Dataset(...)                          0.002s
add_layer + add_mag(compress=False)   0.010s
write 18KB unaligned                  0.006s
_is_compressed()                      0.000s
compress(sequential)                  4.362s   <—
read back                             0.002s
write again (compressed, unaligned)   3.729s   <—
```

Same test body, varying only `shard_shape`:

| shard_shape | compress | rewrite | total |
| --- | --- | --- | --- |
| default 1024³ | 3.726 s | 3.913 s | **7.639 s** |
| 256³ | 0.051 s | 0.023 s | **0.074 s** |
| 128³ | 0.045 s | 0.027 s | 0.072 s |
| 64³ | 0.150 s | 0.018 s | 0.168 s |

**100× on the same assertions.** And the default is already an env-var knob
[...]

PR #1530, opened 2026-09-01
Speed up the webknossos test suite

normanrz:
Speeds up the tests. The main changes are:
- `WK_DEFAULT_CHUNKS_PER_SHARD=8,8,8` otherwise the tests are mainly compressing empty shards
- Splitting test_dataset.py into smaller bits so that they can be run independently (not important for CI, but for Claude)
- Parallelization with xdist (mainly for workstealing and less for actual parallelism)
- Make a few big tests more efficient

Definitely needs to be reviewed commit-by-commit.

Claude's PR description with all the details:

### Description:
- Sets `WK_DEFAULT_CHUNKS_PER_SHARD=8,8,8` for the test session (root `conftest.py`, before `webknossos` is imported, plus `test.py`). The shipped 1024³ shard shape dominated the dataset tests: they write kilobytes but compressing or rewriting touches the whole shard. Five tests that depended on the shipped default now assert against `DEFAULT_SHARD_SHAPE` / `DEFAULT_CHUNKS_PER_SHARD`, and a new `test_shipped_default_shard_shape` still covers the real 1024³ default end to end.
- Shrinks `test_for_zipped_chunks` from 256³/64³ to 128³/32³ — same 64 chunk pairs and assertions, ⅛ the bytes.
- Replaces all 107 `(np.random.rand(...) * 255).astype(np.uint8)` sites with a module-level `rng.integers(...)` (4× faster, 16× less peak memory) and drops the then-redundant `np.random.seed(1234)` calls.
- Makes `ensure_gc` opt-in, requested only by the two image-conversion modules; `clear_testoutput` skips the recursive delete when nothing was written.
- Starts the in-process moto server once per session instead of once per module.
- Reuses one session-scoped 2-worker `MultiprocessingExecutor` instead of building a `cpu_count()` pool per API call. Measured, a short-lived executor costs ~1.4s at `cpu_count()` on a 10-core machine — construction is free, but the first submit spawns the workers — against well under a millisecond for a warm submit. `wrap_executor` is now called as `utils.wrap_executor` so it can be substituted; a `own_executor` marker opts out. Jobs still run out-of-process.
- Enables `pytest-xdist`: `TESTOUTPUT_DIR`, the S3 bucket and the moto port are derived from `PYTEST_XDIST_WORKER`, and `test.py` defaults to `-n <cores/2, capped at 4> --dist worksteal`. Pass `-n0` to stay in-process.
- Splits `tests/dataset/test_dataset.py` (4284 lines, 426 tests) into seven `test_dataset_*.py` modules plus `_dataset_helpers.py`; the three downsampling tests move into the existing `test_downsampling.py`. Pure file movement — no test bodies changed apart from the items above.
- Remaps `.test_durations` onto the new module paths (168 of its entries pointed at files deleted long ago) and has `nightly.yml` upload a fresh one.

### Measurements

Local, on the same machine, excluding the WEBKNOSSOS-server-dependent `test_remote_dataset.py`:

| Selection | Before | After |
| --- | --- | --- |
| `tests/dataset` | **890.9s** | **281.7s** serial · **191.6s** with `-n 4` |
| fast-module slice (195 tests) | **13.3s** | **2.4s** |
| `tests/` minus `tests/dataset` | **213.0s** | **109.0s** |

The pass/fail set is identical to `release-4.0` everywhere it could be compared, including a name-by-name diff of the 51 `test_remote_dataset.py` failures (all of which need a running WEBKNOSSOS server).

### Notes

- The `cluster_tools` lazy-`Manager()` idea from the issue is **not** included. It was implemented and benchmarked against base and came out identical — the existing `increment_usage` path already covers overlapping executors, and the only way to help sequential construction is to keep a manager process alive after the last executor closes, which leaks a process for library users. The session-scoped pool removes the cost in tests anyway.
- Worker count scales with core count rather than being hardcoded: `-n 8` on a 10-core machine was catastrophic (1201s vs 281s serial, 2662s of system time), so CI's 4-vCPU runners get 2.
- `.test_durations` cannot be genuinely refreshed without a WEBKNOSSOS server — a partial `--store-durations` run would drop the server tests entirely. The new nightly artifact is the way to refresh it.

### Issues:
- part of #1503

### Todos:
 - [x] Updated Changelog
 - [x] Updated Documentation
 - [x] Added / Updated Tests


normanrz:
Restructured into eight commits for review (written by Claude). Same tree — the final commit is byte-identical to the version that went green above, checked with `git diff --quiet`, so the CI result carries over.

| | commit |
| --- | --- |
| 1 | Shrink the default shard shape under test |
| 2 | Fix the two heavy test bodies |
| 3 | Narrow the autouse fixtures in tests/conftest.py |
| 4 | Start the in-process moto server once per session |
| 5 | Reuse one process pool across the test session |
| 6 | Make the shared test state xdist-safe and turn on parallelism |
| 7 | **Split tests/dataset/test_dataset.py into seven modules** |
| 8 | Refresh .test_durations and keep it from going stale |

Commit 7 is isolated and is **pure movement**: all 153 top-level definitions are byte-identical before and after. I verified this by parsing both sides and comparing each definition's source rather than eyeballing the diff — nothing added, removed or changed. Every behavioural change happens in commits 1–6, on the monolithic file, so the split diff carries no logic to review.

Reviewing 1–6 individually should be much easier than the previous single commit; each one is independently runnable and has its measurement in the commit message.

One correction to my earlier description: I had written that the executor change was verified by asserting on `os.getpid()` in a job. When re-checking during the restructure I found my probe used `View.for_each_chunk`, which runs inline when no executor is passed and never called `wrap_executor` at all — so it proved nothing. The paths that *do* go through `wrap_executor` are genuinely out-of-process (worker pids distinct from the main process, via both `map` and `map_to_futures`), and the pool is reused across tests, which is the actual optimisation. The conclusion stands; the evidence I originally cited for it did not.

webknossos/test.py
```diff
@@ -174,6 +174,11 @@ def main(args: list[str]) -> None:
         "MULTIPROCESSING_DEFAULT_START_METHOD", "forkserver"
     )
 
+    # Shrink the default shard shape from 1024**3 to 256**3. See conftest.py.
+    os.environ["WK_DEFAULT_CHUNKS_PER_SHARD"] = os.environ.get(
+        "WK_DEFAULT_CHUNKS_PER_SHARD", "8,8,8"
+    )
```
fm3: This seems redundant to the conftest.py – since there only setdefault is used, repeating the default 8,8,8 here should not be necessary, or am I missing something? I think it should be either-or

webknossos/tests/conftest.py
```diff
@@ -81,11 +89,59 @@ def ensure_gc() -> None:
...
 @pytest.fixture(autouse=True, scope="function")
 def clear_testoutput() -> Generator:
     TESTOUTPUT_DIR.mkdir(parents=True, exist_ok=True)
     yield
-    rmtree(TESTOUTPUT_DIR)
+    # Most tests never write here, so skip the recursive walk when empty.
+    if next(TESTOUTPUT_DIR.iterdir(), None) is None:
+        TESTOUTPUT_DIR.rmdir()
+    else:
+        rmtree(TESTOUTPUT_DIR)
```
fm3: strange that this is necessary, one would assume rmtree to do this by itself :shrug:

webknossos/tests/conftest.py
```diff
@@ -81,11 +89,59 @@ def ensure_gc() -> None:
 ### PYTEST SETUP & TEARDOWN
 
 
+@pytest.fixture(scope="session")
+def shared_executor() -> Iterator[Executor]:
+    """One process pool for the whole session.
+
+    Constructing a `MultiprocessingExecutor` costs ~0.3s, which the tests
+    otherwise pay on every API call that does not get an executor passed in.
```
fm3: what is meant by API call? HTTP? Or just public functions of the libs?

webknossos/tests/conftest.py
```diff
@@ -81,11 +89,59 @@ def ensure_gc() -> None:
...
+    """Hands the shared pool to every `utils.wrap_executor` call.
+
+    Opt out with `@pytest.mark.own_executor` where the executor lifecycle
+    itself is under test.
+    """
+    if request.node.get_closest_marker("own_executor"):
+        return
+
+    def wrap_executor(executor: Executor | None = None) -> AbstractContextManager:
+        # Resolved lazily so tests that never spawn a job don't pay for the pool.
```
fm3: what is meant by job? A worker job? or something more generic?

fm3 (approved):
This is great, down to <9 minutes in the CI too! 😎 Thanks for looking into this!

Should we also go for fewer chunks per shard in production if we detect small inputs?

webknossos/test.py
normanrz: You are right, it was redundant — removed from test.py in 45e861c4 (reply written by Claude).

test.py just spawns pytest with rootdir `webknossos/`, so the root conftest.py loads there too and its `setdefault` covers that run. Verified by unsetting the variable in the environment and asserting inside a test that `DEFAULT_SHARD_SHAPE` is still 256³. I had followed the issue text, which suggested both places, without checking that one implies the other.

webknossos/tests/conftest.py
normanrz: Fair — reverted to the plain `rmtree` in 45e861c4 (reply written by Claude).

To be clear about what the branch did: `rmtree` does remove the directory itself, so it was never about correctness. It was avoiding the recursive listing `fs.delete(..., recursive=True)` does for the ~400 tests that never write there. I benchmarked it rather than leave it to intuition: 0.13ms vs 0.10ms per iteration, so about 40ms across the whole suite. That does not justify the extra branch, so it is gone.

webknossos/tests/conftest.py
normanrz: Public functions of the library, not HTTP — reworded in 45e861c4 (reply written by Claude).

Specifically the ones taking an optional `executor`: `Layer.downsample`, `MagView.rechunk`, `View.map_chunk` and friends. When the caller passes none, `wrap_executor` builds a fresh `MultiprocessingExecutor` with `cpu_count()` workers, which is the ~0.3s. The docstring now names them.

webknossos/tests/conftest.py
normanrz: A unit of work submitted to the executor, but the wording was vague — reworded in 45e861c4 to say what is actually deferred (reply written by Claude).

The point of the laziness is that the fixture is autouse, so it runs for all ~850 tests, but most never call one of those functions. Resolving `shared_executor` eagerly would start the worker processes for every test; deferring it into the call means only the tests that really need a pool pay for it.

normanrz:
Thanks! On the production question — worth doing, and less work than it sounds, but it should be its own PR (reply written by Claude).

The mechanism already exists. `_get_shard_and_chunk_shapes` ([layer.py:114](https://github.com/scalableminds/webknossos-libs/blob/release-4.0/webknossos/webknossos/dataset/layer/layer.py#L114)) already shrinks based on the layer bounding box — but only in z, for flat data, dropping to `(4096, 4096, 32)`. There is no equivalent for data that is small in all three dimensions.

The thing actually blocking it is one line up the stack. `add_mag` deliberately throws the bounding box away for the first mag:

```python
# If this is the first mag, the bounding box is probably not yet set,
# so we don't use it for computing chunk and shard shapes
layer_bounding_box=(None if len(self.mags) == 0 else self.bounding_box),
```

"Probably not yet set" is true for `add_layer` + `add_mag` + `write`, but not for the paths where we do know the size upfront. `Dataset.write_layer` derives the bbox from `data.shape` and passes it to `add_layer` *before* calling `add_mag`, so the size is known and then discarded. Same for `add_layer_from_images`, `copy_dataset` and any `add_layer(..., bounding_box=...)`. So the fix is roughly: use the bounding box when it is non-empty rather than gating on mag count, and add an xy branch next to the existing z one.

The reason I would not fold it in here: shard shape is baked into the array metadata at creation and cannot be changed afterwards, so the heuristic has to key on the *final* size, not the current one. A layer created small and then grown via `allow_resize=True` would be stuck with many small shards — lots of objects, slow remote listing, which is the problem sharding exists to solve. That is a real behaviour change for real datasets and deserves its own review and its own benchmark, whereas everything in this PR is test-only apart from the `utils.wrap_executor` indirection.

Happy to open an issue for it with the above if you want.

webknossos/tests/conftest.py
normanrz: Correction to my previous reply, which repeated the issue's ~0.3s figure — that number is wrong, and I have fixed the docstring in 05ef4f6b (reply written by Claude).

Measured on a 10-core machine, breaking the lifecycle apart:

| | first executor | every later one |
| --- | --- | --- |
| `__init__` | 0.99s | **0.0002s** |
| first submit (spawns workers) | 1.0s | **0.96s** (2 workers) / **1.4s** (cpu_count) |
| warm submit | 0.0004s | 0.0004s |
| shutdown | 0.01s | 0.01–0.14s |

So construction is effectively free — `ProcessPoolExecutor` spawns workers lazily — and the ~0.3s in the issue looks like the one-time forkserver/Manager startup being mistaken for a per-instance cost. The real cost is ~1.4s per short-lived executor at `cpu_count()`, which is what `wrap_executor` used.

That makes the saving larger than the issue estimated: ~45 such calls in a 215-test slice is ~60s, not the 14s it projected.

webknossos/tests/conftest.py
```diff
@@ -81,6 +89,57 @@ def ensure_gc() -> None:
...
+    """One process pool for the whole session.
+
+    Library functions that take an optional `executor` (`Layer.downsample`,
+    `MagView.rechunk`, `View.map_chunk`, ...) build a fresh one whenever the
+    caller passes none. Constructing it is free, but the first submit spawns
+    the workers, and under both `spawn` and `forkserver` each of those has to
+    import webknossos again, which alone takes ~0.9s. That is ~1s for two
+    workers and ~1.4s for `cpu_count()` on a 10-core machine, every time.
+    Reusing one pool pays it once per session (per xdist worker) instead of
+    once per call; a warm submit is well under a millisecond.
```
normanrz:
```suggestion
```

normanrz:
Measured what xdist specifically contributes, since it is the highest-risk part of this PR (reply written by Claude).

Full suite, same tree, 10-core machine, two rounds:

| workers | wall clock | speedup |
| --- | --- | --- |
| `-n0` | 414s / 354s | — |
| `-n2` | 257s | 1.61x |
| `-n4` | 225s / 238s | ~1.6-1.8x |

Nowhere near the 4x the worker count suggests: the suite is I/O- and thread-bound and already uses ~2.6 cores serially, so xdist is harvesting idle cores rather than multiplying throughput, and scaling is visibly flat past two workers. Extrapolated to the 8-core CI runners, the slowest job would be ~660s without it instead of 514s — so roughly half of the 856s -> 514s improvement is xdist and half is the shard shape, the gc fixture, the executor and the test bodies. It also lowers billed minutes, since runners bill by wall time.

**Results are identical across configurations**: serial and `-n 4` both give 78 failed / 1164 passed with the same test names (the failures are all the WEBKNOSSOS-server-dependent ones, which pass on CI).

The honest cost is that xdist caused the one real bug here — the download-cache race fixed in 58f05c2a. That was latent in `download_wklibs_sample_archive` on release-4.0, but only parallelism makes it fire, and I cannot prove the download cache was the only piece of shared state left. Mitigations: a bare `pytest` stays in-process, `-n0` disables it, and `-n2` still gets 1.61x if flakiness ever shows up.

Separately: on CI xdist now stacks on top of pytest-split's 3 groups, and the slowest group is about 1.8x the fastest. Once a nightly run produces a fresh `.test_durations`, rebalancing those groups is probably the cheaper next win.

webknossos/Changelog.md
```diff
@@ -42,6 +42,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 - Added a `mag` keyword to `Layer.add_mag_as_copy` and `RemoteLayer.add_mag_as_copy` to register the copied data under a different resolution level than it has in the source dataset, matching the existing `mag` keyword of `add_mag_as_ref`. [#1526](https://github.com/scalableminds/webknossos-libs/pull/1526)
 
 ### Changed
+- Sped up the test suite. `wrap_executor` is now called as `utils.wrap_executor`, so it can be substituted in tests. [#1530](https://github.com/scalableminds/webknossos-libs/pull/1530)
```
```suggestion
- Sped up the test suite. [#1530](https://github.com/scalableminds/webknossos-libs/pull/1530)
```

PR #1531, opened 2026-09-02
[experiment, do not merge] CI runner sizing for the test suite

normanrz:
Draft experiment, **not for merge** — written by Claude, to answer whether one bigger machine plus xdist beats three 8-vCPU machines plus pytest-split.

Branched off #1530 so the test code is identical. The `webknossos_linux` matrix is replaced with a benchmark, all on Python 3.13, steps otherwise unchanged so the numbers are comparable:

| label | runner | workers | split |
| --- | --- | --- | --- |
| `ctl-8vcpu-split{1,2,3}-n4` | ubicloud-standard-8 | 4 | 3 groups (control) |
| `one-8vcpu-n4` | ubicloud-standard-8 | 4 | none |
| `one-16vcpu-n8` | ubicloud-standard-16 | 8 | none |
| `one-16vcpu-n12` | ubicloud-standard-16 | 12 | none |
| `one-30vcpu-n16` | ubicloud-standard-30 | 16 | none |
| `one-8vcpu-arm-n4` | ubicloud-standard-8-arm | 4 | none |
| `one-16vcpu-arm-n8` | ubicloud-standard-16-arm | 8 | none |

The control runs in the same batch rather than being compared against yesterday's numbers, so queue conditions and cache state match.

Two things this is expected to expose beyond wall clock: whether the suite keeps scaling past 4 workers at all (locally it does not — 2 to 4 workers only bought 1.14x), and whether arm64 is even viable, since it needs aarch64 wheels for tensorstore/imagecodecs and an arm64 webknossos docker image for `test.py`. The arm jobs are as likely to fail as to be fast; that is the point of running them.

Windows is disabled here to save runner minutes. Delete this branch once the numbers are recorded.


normanrz:
Results (written by Claude). One Python version (3.13), all nine jobs in the same batch so queue and cache conditions match. Cost columns use Ubicloud list prices: standard-8 $0.00500/min, standard-16 $0.01000/min, standard-30 $0.01875/min ([docs](https://www.ubicloud.com/docs/about/pricing)) — linear at $0.000625 per vCPU-minute, so cost tracks vCPU·s exactly.

| config | result | wall | $/run (this 1 Python version) | $/run (all 5, extrapolated) | vs control |
| --- | --- | ---: | ---: | ---: | ---: |
| **control**: 3 × 8 vCPU, `-n4`, split | ✅ | **300s** (slowest of 245/266/300) | $0.0676 | **$0.338** | — |
| one job, 8 vCPU, `-n4` | ✅ | 417s | $0.0348 | $0.174 | **−49%** but 39% slower |
| one job, 16 vCPU, `-n8` | ✅ | **304s** | $0.0507 | **$0.253** | **−25%**, same wall |
| one job, 16 vCPU, `-n12` | ❌ | 287s | — | — | — |
| one job, 30 vCPU, `-n16` | ✅ | **236s** | $0.0737 | $0.369 | +9%, 21% faster |
| one job, 8 vCPU arm64, `-n4` | ❌ | 96s | — | — | — |
| one job, 16 vCPU arm64, `-n8` | ❌ | 112s | — | — | — |

**Worth being blunt about the absolute numbers**: the 16 vCPU option saves about **8.5 cents per CI run**. At 100 runs/month that is ~$8.50. The money is not the reason to do this — same wall clock, 5 jobs instead of 15, and dropping the `.test_durations` dependency entirely (no more stale-durations maintenance, no more slowest-group-is-1.8x-the-fastest) is. If Ubicloud rounds each job up to a whole minute, consolidation looks slightly better still, since there are a third as many jobs to round.

Consolidating onto a *small* box is a clear loss on wall clock: 8 vCPU with the whole suite is 417s against the control's 300s. Splitting is doing real work there; you only win by also getting more cores. It is the cheapest option by far, if CI latency does not matter.

**arm64 is a non-starter**, and not because of Python wheels: `docker compose run prepare-test-db` dies with `exec /usr/bin/node: exec format error`. The WEBKNOSSOS image is amd64-only, so `test.py` cannot boot a server. Revisit only if multi-arch images get published. (Pricing is identical to x64 per vCPU, so there is no cost incentive either.)

**The `-n12` failure is the interesting one** and argues against just cranking workers up. It failed `tests/test_api_client.py::test_dataset_list` with `assert False` — the server-dependent tests all share one WEBKNOSSOS instance, so a global "list datasets" assertion races once enough workers upload and delete concurrently. That is state xdist cannot isolate the way it isolates `testoutput` directories and moto ports. `-n16` passed, so it is probabilistic rather than a hard threshold, which makes it worse, not better. I would not exceed `-n8` without first making those tests scope their assertions to datasets they own.

Corroborating locally on a 10-core machine: `-n2` 296s, `-n4` 248s, `-n6` 323s, `-n8` 280s, `-n10` **1283s**. Past four workers it stops helping and at ten it falls off a cliff, because each xdist worker also holds a 2-worker pool. Results were identical at every setting, so that is thrashing rather than breakage. Scaling is core-bound and workers ≈ cores/2 is a real ceiling, not a guess — note `test.py` currently caps at 4, which would have to be raised for any of this to take effect.

Not for merge — delete the branch once these numbers are noted.

PR #1532, opened 2026-09-03
Fix CI warnings

normanrz:
### Description:
Clears the warnings that show up in every CI run.

- **setup-uv cache races** — all shards of a matrix resolve the same cache key (it keys on runner platform and the runner's Python, not `matrix.python-version`), so 15 linux jobs raced to write one entry and 14 warned. Now one job per key writes it via `save-cache`; the rest restore only. `cluster_tools_slurm` shares its key with `cluster_tools_other`, which is the writer.
- **Node 20 deprecation** — `orgoro/coverage@v3.2` targets Node 20 and was being force-run on Node 24. v3.3.1 declares `using: node24`.
- **`PytestUnknownMarkWarning`** — `cluster_tools/pyproject.toml` had no `[tool.pytest.ini_options]`, so `requires_modified_slurm_config` was unregistered.
- **`PytestUnraisableExceptionWarning`** — `ims_reader.__init__` sets `self.write` before calling `open()`, which is where `self.hf` is assigned. A corrupt `.ims` makes `h5py.File` raise, and the half-built object's `__del__` → `close()` then hits `AttributeError: no attribute 'hf'`. A class-level `hf = None` default makes `close()` a no-op in that case.

Verified locally by promoting the unraisable warning to an error on the corrupt-file tests: it errors before the change and passes after. Marker registration checked with `-W error::pytest.PytestUnknownMarkWarning`. `./precommit.sh` passes.

Untouched: the upstream NumPy 2.5 recarray deprecation, the `mrcfile` RuntimeWarnings on deliberately-odd fixtures, the missing-CZI-reader notice, and the docs linkcheck warning.

### Todos:
 - [x] Updated Changelog


.github/workflows/ci.yml
```diff
@@ -48,6 +48,9 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "cluster_tools/uv.lock"
+          # Every matrix job resolves the same cache key, so let a single job
+          # write it instead of having the rest race and warn.
+          save-cache: false
```
robert-oleynik: For the jobs we set something like `save-cache: ${{ matrix.python-version == '3.11' }}`. Why not here?

normanrz: (Claude wrote this reply.)

Because those two jobs resolve the *same* cache key, so the version condition wouldn't help here.

The key is `setup-uv-2-<arch>-<runner os>-<runner python>-pruned-<hash of cache-dependency-glob>`. `cluster_tools_slurm` and `cluster_tools_other` are both `runs-on: ubuntu-latest` with the same `cache-dependency-glob: "cluster_tools/uv.lock"` and no `cache-suffix`, so their keys are byte-identical. Setting `save-cache: ${{ matrix.python-version == '3.11' }}` in both would leave two jobs racing for one key — the same warning this PR removes, just 2 racers instead of 5.

Worth noting the key contains the *runner's* uv-managed Python, not `matrix.python-version`. That's why all five shards collide in the first place, and why the `== '3.11'` condition is really just "pick one arbitrary shard" rather than anything version-specific.

`cluster_tools_slurm` is also the wrong job to be the writer: it never runs `uv sync` on the runner, only inside the containers via `docker exec`. Its runner-side `~/.cache/uv` is essentially empty, so if it won the race it would save a cache that does the other job no good.

The alternative would be a distinct `cache-suffix` per job, and then both could use the version condition. That costs a second cache entry against the 10 GB budget for a cache that's nearly empty, so I skipped it.

I've pushed a clearer inline comment saying this, since the old one clearly didn't.

.github/workflows/ci.yml
```diff
@@ -48,6 +48,11 @@ jobs:
         with:
           enable-cache: true
           cache-dependency-glob: "cluster_tools/uv.lock"
+          # Restore only: this job's key is identical to cluster_tools_other's
+          # (same runner and dependency glob), and that job is the writer.
+          # Nothing is installed on the runner here anyway — uv sync runs
+          # inside the containers.
```
robert-oleynik: > Note: While `uv` is executed inside the container, the cluster tools are mounted inside the container and the cache is *not* used, because it is *not* mounted.

We may want to add a note to say that the cache from `cluster_tools_other` is used.

robert-oleynik (approved):

.github/workflows/ci.yml
normanrz:
```suggestion
          # Restore only: The cache from `cluster_tools_other` is used.
```

PR #1280, opened 2025-03-26
Add warning for outdated libs versions

hotzenklotz:
### Description:
- PR adds a version check when importing the wk-libs. It will provide a warning to users when the installed wk libs package is outdated compared to the latest available version on PyPi.
- PR adds a `--version` CLI option to print the current version.

### Todos:
Make sure to delete unnecessary points or to check all before merging:
 - [x] Updated Changelog
 - [ ] Updated Documentation
 - [ ] Added / Updated Tests
 - [ ] Considered adding this to the Examples

PR #1536, opened 2026-09-07
Add missing packaging runtime dependency

normanrz:
The nightly [Verify Published Package](https://github.com/scalableminds/webknossos-libs/actions/runs/34071036240) run fails on every Python version:

```
File ".../webknossos/utils.py", line 32, in <module>
    from packaging.version import InvalidVersion, Version
ModuleNotFoundError: No module named 'packaging'
```

`webknossos/utils.py` has imported `packaging.version` at module level since #1280, but `packaging` was never declared as a dependency — it happened to be pulled in transitively. 4.0.0 dropped `tifffile`/`scm-pims` from the base package, so a clean `pip install webknossos` now fails on `import webknossos`.

Reproduced locally against the published 4.0.0 wheel; fixed by declaring `packaging >=21.0` and relocking.


webknossos/Changelog.md
```diff
@@ -20,6 +20,7 @@ For upgrade instructions, please check the respective _Breaking Changes_ section
 
 ### Fixed
 - Fixed that the `webknossos --version` command failed if the tifffile extra dependency was not installed. [#1534](https://github.com/scalableminds/webknossos-libs/pull/1534)
+- Added the missing `packaging` runtime dependency, which made `import webknossos` fail with `ModuleNotFoundError: No module named 'packaging'` in a clean install. [#1536](https://github.com/scalableminds/webknossos-libs/pull/1536)
```
```suggestion
- Added the missing `packaging` runtime dependency. [#1536](https://github.com/scalableminds/webknossos-libs/pull/1536)
```

fm3 (approved):
LGTM, but will need a refreshed uv.lock probably

PR #1541, opened 2026-09-14
Resolve Layer.export bounding boxes in the axes of the mag's array

normanrz:
### Description:
- `Layer.export.as_ozx` / `as_tiff_stack` / `as_ome_tiff` resolved the export region with `bounding_box.normalize_axes(num_channels).intersected_with(layer.normalized_bounding_box)`. For remote layers the layer's bounding box can lack the channel axis while the streamed array is 4D `cxyz` (seen in production for a zarr dataset whose mags have `axisOrder: {x:1, y:2, z:3}`, see scalableminds/voxelytics#4790 / scalableminds/voxelytics#4792). The intersection then raised `Operation with two bboxes is only possible if they have the same axes and axes order. ('c', 'x', 'y', 'z') != ('x', 'y', 'z')`, and even with matching axes the subsequent `mag_view.read()` would fail on tensorstore's rank check.
- `_resolve_export_bbox` now takes the `MagView` and builds the region in `mag_view.info.bounding_box.axes`, i.e. the axes of the array that is actually read: per axis name the request is clipped to the layer's extent, axes the array doesn't have are ignored, and `c` always spans all channels (as before). `as_ozx` resolves against the finest exported mag.
- Regression test with a layer whose properties bounding box is `xyz` while its zarr3 array is `cxyz`, exercising all three exporters. It fails on `master` with the error above.

### Steps to test:
- `uv run pytest tests/dataset/test_layer_export.py -n0`

### Issues:
- Blocks scalableminds/voxelytics#4795, which ports `export_tiff` to `Layer.export` and would otherwise reintroduce the production failure.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


normanrz:
@MichaelBuessemeyer these are the same fixes as https://github.com/scalableminds/voxelytics/pull/4790 and https://github.com/scalableminds/voxelytics/pull/4792

MichaelBuessemeyer (approved):
ok 👍

PR #1542, opened 2026-09-15
cluster_tools: add output_writer for persisting job results

normanrz:
### Description:
Alongside `output_pickle_path`, jobs can now hand their pickled result to a callable, so that users of `cluster_tools` can checkpoint results to non-filesystem storage (needed for voxelytics' database-backed checkpointer, scalableminds/voxelytics#4803).

- `Executor.map_to_futures(..., output_writer_getter=...)` and `submit(..., __cfut_options={"output_writer": ...})`. `CFutDict` keys are now both optional (at least one must be given); `output_pickle_path` behaves exactly as before.
- The writer receives the pickled `(True, result)` bytes — the same format `output_pickle_path` files have — and is only called in the success case, in the job process, before the result is reported back. A raising writer fails the job (the result would not be checkpointed).
- `MultiprocessingExecutor`, `SequentialExecutor`, `DaskExecutor` and the `*_with_pickling` variants go through the shared `_execute_and_persist_function`; `ClusterExecutor` serializes the writer into the job's input pickle (now a 5-tuple; `remote.py` still accepts 4-tuples) and keeps using its `cfut` outfile for transport when no `output_pickle_path` is given. `BatchingExecutor` rejects it like it rejects `output_pickle_path_getter`.

I could only run the multiprocessing/sequential executors locally (no docker); the new tests are parametrized over all executors, so the Slurm path is covered by CI.

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


cluster_tools/cluster_tools/remote.py
```diff
@@ -78,6 +79,9 @@ def worker(
         result = True, fun(*args, **kwargs)
         logging.info("Job computation completed.")
         out = pickling.dumps(result)
+        if output_writer is not None:
+            # A failing writer fails the job, since the result would not be checkpointed.
+            output_writer(out)
```
philippotto, line 84: I think this doesnt handle the case where the actual `fun` function fails. in that case the except will activate and the remaining code will only write the pickle to disk (old mechanism). I think, it would make more sense to call `output_writer` below the try-except block (similar to where the pickle file is written). the case where output_writer itself fails might need to be handled additionally.

philippotto:
@coderabbitai review

cluster_tools/cluster_tools/remote.py
normanrz: 🤖 *Reply written by Claude:*

The writer was only reached on success (a failing `fun` jumps to the `except`, which only writes the failure pickle — that is intended, the writer must never checkpoint a failure), and a writer exception was caught by the same `except` and turned into a job failure. But that was easy to misread, so 5267ee9c restructures it as you suggest: the writer is now called after the try/except, only if `result[0]` is true, in its own try/except that converts a writer error into a job failure (`(False, exc)` in the preliminary pickle). Also initialized `output_writer = None` before the `try`, so an unpickling failure cannot leave it unbound.

cluster_tools/cluster_tools/executors/multiprocessing_.py
```diff
@@ -14,31 +14,83 @@
...
+    output_pickle_path: Path | None,
+    output_writer: OutputWriter | None,
+    output: Any,
+) -> None:
+    """Persists the pickled `output` to the file and/or the writer."""
+    data = pickling.dumps(output)
+    if output_pickle_path is not None:
+        with output_pickle_path.open("wb") as file:
+            file.write(data)
+    if output_writer is not None:
```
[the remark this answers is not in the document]
normanrz: 🤖 *Reply written by Claude:*

Valid — fixed in 5267ee9c: `persist_output` now calls the writer before writing the file, and `test_failing_output_writer_leaves_no_checkpoint_file` covers it.

cluster_tools/cluster_tools/remote.py
philippotto: > the writer must never checkpoint a failure

why? this diverges from the existing pickling-serialization-logic. can't we keep this unified? on a meta level: the time you saved by having claude reply, I had to spend extra deciphering it ;)

normanrz: Yeah, sorry! I wanted to check first, but then got carried away. I'll work on this.

normanrz: IIUC the current code only persist success to the `output_pickle_path` (i.e. a checkpoint) whereas failures remain in the `preliminary_output_pickle_path` and are re-raised in the `ClusterExecutor._completion`. Arguably, Claude's code is the unified behavior.

philippotto: ok, this part makes sense. however, how is the exception wired back to the main program? or is this reponsibility left to the (preliminary) pickle file? I thought the high-level plan was to remove the pickle files.

normanrz: Yes, in the current code the preliminary pickle file is used to send back the exception. 
I could live with keeping the preliminary pickle files, because they are cleaned up at the end of the executor. Using the output_writer for transporting exceptions is acutally quite difficult, because the cluster_tools don't know what is in the function and don't have access to a potential db that holds the contents. So, we would need a more sophisticated interface (see below). What do you think?

```python
class OutputStore(Protocol):
    # job side
    def write(self, key: str, data: bytes, *, success: bool) -> None: ...
    # client side
    def poll(self, keys: Iterable[str]) -> set[str]: ...   # which keys have a written result
    def read(self, key: str) -> bytes: ...
    def delete(self, key: str) -> None: ...                # stale-on-submit and cleanup
```

Btw, do you remember why the preliminary output pickles are not removed immediately, but collected in `files_to_clean_up` and deleted on Executor close?

normanrz: Built in https://github.com/scalableminds/webknossos-libs/pull/1547

philippotto: sounds good :+1: 

> Btw, do you remember why the preliminary output pickles are not removed immediately, but collected in files_to_clean_up and deleted on Executor close?

no, I dont really remember. we forked the code from another repo. maybe this behavior dates back to that point.

normanrz:
Closing in favor of #1547

cluster_tools/cluster_tools/remote.py
normanrz: Followed up in #1547: transient outputs (jobs without a custom output key) are now deleted right after the result **or the error** has been read, so nothing is deferred to `files_to_clean_up` anymore. Failed outputs at a custom key are still kept for inspection.

Worth noting the old behavior was also buggy for failures: `_completion` appended the *final* outfile name to `files_to_clean_up`, but on failure only `<name>.preliminary` exists — so the unlink logged a "Could not delete file during clean up" warning and the preliminary file leaked for the lifetime of the cfut dir. (Written by Claude)

PR #1544, opened 2026-09-21
Fix converting tiffs with several unknown axes

normanrz:
### Description:
- tifffile names every axis it cannot identify `Q`, so a plain 4D tiff without axis metadata decodes as `QQYX`. `TiffSliceReader` registered both as `q` and failed with `axis 'q' already exists`, which then surfaced as a misleading `CorruptImageError`.
- Duplicate unknown axes are now renamed: the innermost one becomes `z` (unless the file has one), the others are numbered (`q`, `q1`, …) and end up as additional axes.

### Issues:
- none

### Todos:
 - [x] Updated Changelog
 - [x] Added / Updated Tests


normanrz:
Good catch, `with_bounds` takes a size, not an end coordinate. Fixed in cea6d70b and added a shape assertion so the test actually catches this.

_(Comment written by Claude)_

fm3 (approved):
Sweet!

PR #1546, opened 2026-09-21
Account for mag factors in the downsampling buffer shape

### LINKED ISSUE #1545: determine_downsample_buffer_shape can request more memory than expected for anisotropic mags
`determine_downsample_buffer_shape` returns a shape in the **target** mag, and its comment promises that "the data that is read is up to 1024³ vx in the source magnification". That only holds when every mag factor is 2. An anisotropic mag ladder can produce larger factors, and the read then scales with them.

Concretely, for a layer whose only mag is `8-8-1` at `voxel_size = (8, 8, 8)`, `calculate_mags_to_downsample` skips `(8,8,2)`, `(8,8,4)` and `(8,8,8)` — each has the same `max_dim` as the current mag — so the first step goes straight from `8-8-1` to `Mag(16)` with `mag_factors = (2, 2, 16)`. With the target's default shard shape of `1024³`, `buffer_shape` is `512³`, and `_downsampling_utils.py:364` turns that into a source read of `1024 × 1024 × 8192` voxels = **16 GiB in a single `read_cxyz` allocation** (`_downsampling_utils.py:370`), plus the ~2 GiB `file_buffer`. The intended budget is 1024³ source voxels, i.e. 2 GiB at uint16 which is 8× less.

## Minimal reproduction

```python
import tempfile

from cluster_tools import SequentialExecutor
from webknossos import BoundingBox, Dataset, Mag
from webknossos.dataset.layer.view import View

ds = Dataset(tempfile.mkdtemp(), voxel_size=(8, 8, 8))
layer = ds.add_layer("color", "color", dtype_per_channel="uint16", data_format="zarr3")
# Large bounding box, but no data is written: zarr3 fill_value keeps this empty on disk.
layer.bounding_box = BoundingBox((0, 0, 0), (18520, 18712, 36764))
layer.add_mag(Mag("8-8-1"), chunk_shape=(32, 32, 32), shard_shape=(4096, 4096, 32))

# Report the requested read instead of allocating it.
def spy(self, *args, **kwargs):
    size = kwargs["absolute_bounding_box"].in_mag(self.mag).size_xyz
    print(f"read from mag {self.mag}: {size} = {size.prod() * 2 / 2**30:.1f} GiB (uint16)")
    raise SystemExit

View.read_cxyz = spy

layer.downsample(from_mag=Mag("8-8-1"), coarsest_mag=Mag(1024), executor=SequentialExecutor())
```

Output on webknossos 4.1.0:

```
read from mag 8-8-1: Vec3Int(1024,1024,8192) = 16.0 GiB (uint16)
```

## Suggested fix

Let the buffer shape account for the mag factors, so the *source* read stays within the documented budget:

```python
def determine_downsample_buffer_shape(
    array_info: ArrayInfo, mag_factors: Vec3Int
) -> Vec3Int:
    # Keep the data read per tile at up to 1024³ vx in the source magnification.
    return (
        (Vec3Int.full(1024) // mag_factors)
        .pairmax(Vec3Int.ones())
        .pairmin(Vec3Int.full(512))
        .pairmin(array_info.shard_shape)
    )
```

For `(2, 2, 16)` this gives `(512, 512, 64)` and a 2 GiB read, and leaves the common 2× case unchanged. `determine_upsample_buffer_shape` has the same structure and may deserve a look.

Versions: webknossos 4.1.0, zarr3/tensorstore, numba extra installed.

normanrz:
Fixes #1545

`determine_downsample_buffer_shape` returned a fixed `512³` shape in the target mag, so the source read per tile scaled with the mag factors. For anisotropic mag ladders a single step can have factors > 2, e.g. `8-8-1 -> 16` at an isotropic voxel size has factors `(2, 2, 16)`, which read `1024 × 1024 × 8192` source voxels (16 GiB for uint16) instead of the intended `1024³`.

The buffer shape is now `min(1024 // mag_factors, 512, shard_shape)` (clamped to ≥ 1), so the source read stays within `1024³` voxels. The common 2× case is unchanged.

`determine_upsample_buffer_shape` was left as is: its buffer is in the *target* (finer) mag and the source read is `buffer_shape / factors`, so larger factors shrink rather than grow the read.

### Tests
- Unit test for `determine_downsample_buffer_shape` with isotropic, anisotropic and extreme factors, plus the shard-shape cap.
- Regression test mirroring the issue's reproduction (spying on `read_cxyz`), asserting the source read is `1024³` for the `8-8-1 -> 16` step.


webknossos/tests/dataset/test_downsampling.py
```diff
@@ -742,6 +743,70 @@ def test_downsample_custom_chunk_and_shard_shapes(tmp_upath: UPath) -> None:
     assert layer.get_mag(4).info.shard_shape.xyz == Vec3Int.full(512)
 
 
+def test_determine_downsample_buffer_shape(tmp_upath: UPath) -> None:
+    """The buffer shape (in the target mag) is chosen so that the source read
+    stays within 1024³ vx, also for mag factors larger than 2 (#1545)."""
```
Tobias314: Do we want to mention issues in test descriptions or not?

webknossos/tests/dataset/test_downsampling.py
```diff
@@ -742,6 +743,70 @@ def test_downsample_custom_chunk_and_shard_shapes(tmp_upath: UPath) -> None:
...
+    ds = Dataset(tmp_upath / "buffer_shape", voxel_size=(1, 1, 1))
+    layer = ds.add_layer("color", COLOR_CATEGORY, data_format=DataFormat.Zarr3)
+    info = layer.add_mag(1, chunk_shape=32, shard_shape=1024).info
+
+    # common isotropic case is unchanged
```
Tobias314: would not write "unchanged" in the comment because that gets confusing later on.
```suggestion
    # common isotropic case, results in isotropic buffer
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
```diff
@@ -36,12 +36,21 @@ class InterpolationModes(Enum):
...
-    # shapes uses a lot of RAM, especially for segmentation layers which use the mode filter.
+def determine_downsample_buffer_shape(
+    array_info: ArrayInfo, mag_factors: Vec3Int
+) -> Vec3Int:
+    # This is the shape of the data in the downsampling target magnification. It is
+    # chosen so that the data that is read is up to 1024³ vx in the source magnification,
+    # regardless of the mag factors (which can exceed 2 per axis for anisotropic mag
+    # ladders). Using larger shapes uses a lot of RAM, especially for segmentation
+    # layers which use the mode filter.
     # See https://scm.slack.com/archives/CMBMU5684/p1749771929954699 for more context.
```
Tobias314, line 47: Can we add a short explanation of the mag_factors param? Or is this a common term in the libs?
Maybe just turn this into a docstring with "mag_factors: how many times each axis is downsampled, e.g. 2 when downsampling from mag 4 to mag 8"

Tobias314 (approved):
Cool, LGTM.
Just some notes on comments.

webknossos/tests/dataset/test_downsampling.py
normanrz:
```suggestion
    stays within 1024³ vx, also for mag factors larger than 2."""
```

webknossos/webknossos/dataset/layer/_downsampling_utils.py
normanrz: That is common in the downsampling code. I also think it is quite self-explanatory.

